PR #29, opened 2020-02-21
Allowing investigation doc to be updated on incident data change. 

kevgliss:
Allows the investigation document to be updated whenever various incident attributes change.

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -493,6 +493,18 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
     # we load the incident instance
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
 
+    # we get the incident document
+    incident_document = get_document(
+        db_session=db_session,
+        incident_id=incident_id,
+        resource_type=INCIDENT_RESOURCE_INVESTIGATION_DOCUMENT,
+    )
+
+    # update the incident document
+    document_updates = {f"Status: {incident.status}": f"Status: Active"}
+    document_plugin = plugins.get(INCIDENT_PLUGIN_DOCUMENT_SLUG)
+    document_plugin.update(incident_document.id, document_updates, raw=True)
+
     if incident.status == IncidentStatus.active:
         if command:
             convo_plugin = plugins.get(INCIDENT_PLUGIN_CONVERSATION_SLUG)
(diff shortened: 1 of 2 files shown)


src/dispatch/incident/flows.py
```diff
@@ -493,6 +493,18 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
...
+    incident_document = get_document(
+        db_session=db_session,
+        incident_id=incident_id,
+        resource_type=INCIDENT_RESOURCE_INVESTIGATION_DOCUMENT,
+    )
+
+    # update the incident document
+    document_updates = {f"Status: {incident.status}": f"Status: Active"}
+    document_plugin = plugins.get(INCIDENT_PLUGIN_DOCUMENT_SLUG)
+    document_plugin.update(incident_document.id, document_updates, raw=True)
```
mvilanova: @kevgliss I'd move it after the status check and before updating the incident status, so we only update the doc when the status of the incident is not active.

src/dispatch/incident/flows.py
```diff
@@ -914,6 +966,20 @@ def incident_assign_role_flow(assigner_email: str, incident_id: int, action: dic
         )
 
     if assignee_role == ParticipantRoleType.incident_commander:
+        # we get the incident document
+        incident_document = get_document(
```
mvilanova: @kevgliss if we get the investigation document here, let's not do it again in line 986.

src/dispatch/incident/flows.py
```diff
@@ -914,6 +966,20 @@ def incident_assign_role_flow(assigner_email: str, incident_id: int, action: dic
...
+
+        # update the incident document
+        document_updates = {
+            f"Incident Commander: {old_commander_name}": f"Incident Commander: {incident.commander.name}"
+        }
+        document_plugin = plugins.get(INCIDENT_PLUGIN_DOCUMENT_SLUG)
+        document_plugin.update(incident_document.id, document_updates, raw=True)
+
         # we update the conversation topic
         set_conversation_topic(incident)
```
mvilanova: @kevgliss I'd suggest we do this before updating the doc and ticket to provide feedback to the user right away.

src/dispatch/incident/flows.py
mvilanova: Same for the other flows.

kevgliss:
I'm going to close this for now as it's stale. Going to see if there are better approaches for keeping dispatch/google-docs in sync.

forestmonster:
If this feature gets interesting again, just wanted to add a shout-out for using Google Spreadsheets (if we stick with GSuite). Adds the ability to address and modify specific fields, and we could make it look more or less like the investigation doc does today. Seems like Google Docs isn't gonna add this.

PR #37, opened 2020-02-25
Fix Docker build directory visibility

TheDahv:
The project Dockerfile lives in a `./docker` directory, but references
files hosted in the root directory, namely:

- `.nvmrc`
- `setup.py`

Dockerfiles can only reference files in the same folder as the build
context. This results in build errors reported in
https://github.com/Netflix/dispatch-docker/issues/2 and initially
addressed in https://github.com/Netflix/dispatch-docker/pull/1

```
ERROR: Service 'web' failed to build: COPY failed: stat
/var/lib/docker/tmp/docker-builder448268992/.nvmrc: no such file or
directory
```

This is a simple fix to move the Dockerfile into a reasonable scope via
symlink. However, open source contributors cannot build the image
locally since the build steps appear rely on access to resources on
Netflix internal network.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+docker/Dockerfile
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -10,6 +10,21 @@ Dispatch relies on multiple services to work, which are all orchestrated by `Doc
 - At least 2400MB memory
 - 2 CPU Cores
 
+## Building a Docker Image
+
+The Netflix team builds Docker images to run Dispatch. To run locally, run the
+following from the project root:
+
+```
+docker build -t docker-local .
+```
+
+This is only useful for local development purposes, as you should use the
+published Docker image described in the next section.
+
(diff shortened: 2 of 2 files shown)


kevgliss:
Was literally just looking at this error. You're correct I need to break an internal dependency (it's not needed).

TheDahv:
@kevgliss the changes you made mean I can build the image locally now.

One last problem: the docker build expects `./docker.env` to exist. I imagine you don't want local developer changes checked in to version control.

A pattern I see in projects is a `.env.example` or a `.env.schema` file with empty fields that a developer can copy over and fill out themselves.

I'd recommend the following:

- `echo docker/.env >> .gitignore`
- `echo '' > env.example`
- Update the README to instruct developers to copy and update the file before building

Do you agree?

kevgliss:
That pattern sounds reasonable to me, I believe we follow it in dispatch-docker: https://github.com/Netflix/dispatch-docker/blob/master/.env.example

PR #38, opened 2020-02-25
Add Python 3.8 classifier

abdullahselek:
Thanks for creating useful library. As it already runs on Python 3.8 in CI I think it is good to add Python 3.8 classifier.


kevgliss:
Thanks for the PR!

Issue #50, opened 2020-02-26
chmod u+x docker-entrypoint.sh && dependency


drPytho:
```
root@917329a03ed4:/etc/dispatch# ./docker-entrypoint.sh server start
bash: ./docker-entrypoint.sh: Permission denied
root@917329a03ed4:/etc/dispatch# chmod u+x docker-entrypoint.sh
root@917329a03ed4:/etc/dispatch# ./docker-entrypoint.sh
root@917329a03ed4:/etc/dispatch# ./docker-entrypoint.sh server start
./docker-entrypoint.sh: line 17: exec: server: not found
root@917329a03ed4:/etc/dispatch# ./docker-entrypoint.sh "server start"
Traceback (most recent call last):
  File "/usr/local/bin/dispatch", line 5, in <module>
    from dispatch.cli import entrypoint
  File "/usr/local/lib/python3.8/site-packages/dispatch/cli.py", line 13, in <module>
    from dispatch.application.models import *  # noqa
  File "/usr/local/lib/python3.8/site-packages/dispatch/application/models.py", line 9, in <module>
    from dispatch.models import DispatchBase, applications_incidents, TimeStampMixin
  File "/usr/local/lib/python3.8/site-packages/dispatch/models.py", line 181, in <module>
    from dispatch.status_report.models import *  # noqa
ModuleNotFoundError: No module named 'dispatch.status_report'
root@917329a03ed4:/etc/dispatch#
```

After having built the image with `docker build . -t dispatch-local`

And running it with `docker run -ti --entrypoint=bash dispatch-local`

drPytho:
Also just running it `docker run dispatch-local`
results in
`server: 1: exec: /etc/dispatch/docker-entrypoint.sh: Permission denied`

drPytho:
I propose for the first issue to add under 
`COPY ./docker/docker-entrypoint.sh ./docker/.env $DISPATCH_CONF/`
`RUN chmod u+x $DISPATCH_CONF/docker-entrypoint.sh`

drPytho:
The `ModuleNotFoundError: No module named 'dispatch.status_report'` I have no idea of how to fix.

kevgliss:
`ModuleNotFoundError: No module named 'dispatch.status_report` should be fixed by #52

drPytho:
Thanks, both issues are solved. Saw that @TheDahv made an improved solution too. :+1:

PR #51, opened 2020-02-26
Update Dockerfile

drPytho:
`chmod docker-entrypoint.sh` to make it executable in the container.


drPytho:
Should solve #50

kevgliss:
Thanks!

TheDahv:
FWIW git carries with it the file permissions. Running `chmod +x` on the source file would have carried through in the Docker build and you wouldn't need to set the permission at build time.

Only noticed it because I solved the same problem in a different branch I'm about to submit and saw the conflict.

PR #53, opened 2020-02-26
Fix docker image startup

TheDahv:
Summary of changes:

- improve parsing in startup script
 -- fix how 2-word commands are parsed by bash
 -- add the "scheduler" command as a recognized command to support the scheduler container in [docker-compose](https://github.com/Netflix/dispatch-docker/blob/f65df7ca90e51ae2c72e87175efb5c275e6b21e0/docker-compose.yml#L20-L21)

- exclude the `.env` file in the build step
 -- adjustments to the environment file shouldn't trigger a rebuild since it configures run-time settings
 -- it's not used for the build
 -- the file is more appropriately supplied at run-time via `docker run` or via `docker-compose`
- set the default command
 -- reading through the [documentation](https://hawkins.gitbook.io/dispatch/cli#start) it appears this image is trying to boot up the server as the default invocation
 -- even so, this will _still_ fail because the command expects an app
 -- the more appropriate action should be to better document how to control startup behavior in a docker environment

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -121,8 +121,7 @@ RUN set -x \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* 
 
-COPY ./docker/docker-entrypoint.sh ./docker/.env $DISPATCH_CONF/
-RUN chmod u+x $DISPATCH_CONF/docker-entrypoint.sh 
+COPY ./docker/docker-entrypoint.sh $DISPATCH_CONF/
 
 EXPOSE 8000
 VOLUME /var/lib/dispatch/files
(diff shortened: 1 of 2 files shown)


kevgliss:
Thanks for all your help @TheDahv filling in my admittedly limited knowledge of what a good docker setup looks like.

Yes, we do expect an application. I believe that should be fixed easily enough by supplying `dispatch.main:app` to the `server start` command.

TheDahv:
> Thanks for all your help @TheDahv filling in my admittedly limited knowledge of what a good docker setup looks like.

No worries. I use Docker and Compose quite a bit, and I want to use this!

> Yes, we do expect an application. I believe that should be fixed easily enough by supplying dispatch.main:app to the server start command.

Yep, PR coming soon based on https://github.com/TheDahv/dispatch-docker/tree/fix-service-configurations

Cleaning up now...

PR #58, opened 2020-02-27
Remove need for entrypoint script

daniel-gallagher:
By making a few changes to the `ENTRYPOINT`, you should be able to eliminate the need for a startup script. From what I can tell, there are three different commands that potentially will be called. All of them begin with `dispatch`, so the value for entrypoint can be set to that. Then the Dockerfile will move on to the `CMD` values which will append `server start dispatch.main:app` if no parameters are given when the container is launched. This allows for the install script to perform the initial database upgrade as expected, since it provides the parameters `database upgrade` when the container is launched and those will override the values provided in the Dockerfile.


daniel-gallagher:
In addition to the changes to the Dockerfile, I had added a couple `__init__.py` files that were missing. Those can be disregarded if necessary and are not needed for the Dockerfile changes. Though I believe they should probably exist anyways.

kevgliss:
I think this makes sense (cc @TheDahv). I've merged some additional changes in, would you mind resolving the conflicts?

daniel-gallagher:
So I just noticed the following in a fork from @TheDahv 

https://github.com/TheDahv/dispatch-docker/blob/494304bd510f7a31d6b578d1e268007013b7e7c9/docker-compose.yml#L23

https://github.com/TheDahv/dispatch-docker/blob/494304bd510f7a31d6b578d1e268007013b7e7c9/docker-compose.yml#L26

Those help me understand better the operation of the two containers and should still operate correctly with the above changes. It might even eliminate the need for explicitly defining `CMD` in the Dockerfile.

TheDahv:
You know, honestly I'm inclined to agree. I don't think the startup script added as much value as the complexity it introduced.

It _does_ do something interesting things regarding positional flags and making sure everything lands in the right place, but it also appears to be solving a problem that doesn't exist yet.

> All of them begin with dispatch, so the value for entrypoint can be set to that. 

Agree with this.

> Then the Dockerfile will move on to the CMD values which will append server start dispatch.main:app if no parameters are given when the container is launched

The bigger question for the Netflix team: is there any value to having a default command? Does the Dispatch web binary do anything useful without companion services (namely `postgres` and `dispatch-scheduler`?)

If the default behavior results in a non-functioning app, then having defaults seems misleading and the API should instead guide the user to the appropriate setup -- that is, specify no `CMD` and only have `ENTRYPOINT`

@daniel-gallagher what are the conflicts that are left over? You include changes here that have been resolved in other PRs, so perhaps this is a matter of merging up and limiting the scope of changes to the entry point?

TheDahv:
@kevgliss does Netflix have any opinions or standards on keeping neat histories? Should we be looking at squash merges or anything like that?

daniel-gallagher:
Yeah I apologize for my squirrel commits... Was rushing and distracted yesterday while testing and did not practice proper etiquette in my repo.

daniel-gallagher:
There are two other issues that still exist for me that I am trying to sort through. Both of them have to do with files and folders that are missing in the built docker image. I am going to submit an issue for each of them in a few minutes, as I feel I don't have a good enough understanding on the code yet to suggest how to fix them. Some of the back and forth commits above are evidence of that.

TheDahv:
@daniel-gallagher 

> Yeah I apologize for my squirrel commits... 

No worries! That's what squash merges are for :D
But honestly if Netflix doesn't care, it's probably fine.

daniel-gallagher:
I just submitted two issues that reference the other problems I am still trying to solve.

TheDahv:
@daniel-gallagher if @kevgliss likes your changes and approves the direction, I still think it's valuable to remove the complexity in the startup script. After I catch my breath a bit, I might try jumping back in to the startup issues.

kevgliss:
We will generally squash the commits on merge. I'm okay removing the startup script. I was using another project as a template when I initially added that behavior so it was a remnant of that.

PR #74, opened 2020-03-04
Fix Environment Variable Naming

njoerd114:
See Issue: https://github.com/Netflix/dispatch-docker/issues/18


mvilanova:
Nice catch, @njoerd114! At least we were consistent in the code 😝

nibhatta:
Does this need to be changed in the .env.example file in the dispatch-docker repo as well? 

https://github.com/Netflix/dispatch-docker/blob/master/.env.example

`GOOGLE_SERVICE_ACCOUNT_**DELAGATED**_ACCOUNT=account`

I ran the install.sh script and ran 'docker-compose up -d' and see an error on the console:
`KeyError: "Config 'GOOGLE_SERVICE_ACCOUNT_DELEGATED_ACCOUNT' is missing, and has no default."`

mvilanova:
@nibhatta, yes. I'll take care of it. Thanks

PR #83, opened 2020-03-09
update .env.example to reflect environment variables in docs

timwheelercom:
Update .env.example to better reflect the environment variables specified in the [documentation](https://hawkins.gitbook.io/dispatch/configuration/app)

Add `VUE_APP_DISPATCH_OPEN_ID_CONNECT_URL` and `VUE_APP_DISPATCH_CLIENT_ID`


docker/.env.example
```diff
@@ -1,14 +1,19 @@
...
+# Persistence
 DATABASE_HOSTNAME=""
 DATABASE_CREDENTIALS=""
 
+# Authentication
 JWKS_URL=""
-DISPATCH_HELP_EMAIL = ""
-DISPATCH_HELP_SLACK_CHANNEL = ""
+VUE_APP_DISPATCH_OPEN_ID_CONNECT_URL=""
+VUE_APP_DISPATCH_CLIENT_ID=""
```
mvilanova: These variables live in the `.env` file located in https://github.com/Netflix/dispatch/tree/develop/src/dispatch/static/dispatch. Would you mind creating a new `.env.example_vue_app` file and add them there? We may also want to add a note to let people know where the file lives.

PR #88, opened 2020-03-10
Allow for header based auth.

kevgliss:
This should allow a Dispatch to be able to respect a given header as it's the source of identity.

If you are bringing your own authentication layer you can set `DISPATCH_AUTH_HEADER_KEY` to the email address of the user making the request. e.g. (`x-amzn-oidc-identity` for ALBs, https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html)


kevgliss:
cc @willbengtson

Issue #55, opened 2020-02-27
REQUEST: Configure enabled/disabled plugins


TheDahv:
Reading through [the docs](https://hawkins.gitbook.io/dispatch/configuration) and the code, I'm not able to discern if there's a way to boot the core Dispatch experience _without_ the Slack, Google, Jira, and PagerDuty plugins?

The docs and the way the app behaves appear to imply that Dispatch will not work _without_ them configured.

However, suppose I just want to get started to evaluate without configuring everything, I'd like a way to disable those plugins just to reduce the distance from" 0-to-UI" for evaluation purposes.

Or boot Dispatch by default and enable plugins one at a time via an Admin UI.

If this isn't already supported I can update the issue title to make this more of a feature request.

mvilanova:
@TheDahv that's correct. It's not possible to start the app without configuring them first at the moment, as Dispatch relies on them to provide the core functionality. We will consider adding support for it in the future. Thanks for the feedback.

kevgliss:
We can definitely do this, the only downside is the `dispatch` isn't particularly useful without any plugins installed (essentially a crud app with some tables). From a new user perspective; does that matter? Is it enough to see something run?

TheDahv:
Here's my thinking on it:

- a simple CRUD app is enough to show value non-technical stakeholders at a company
- it enables someone like me to say "hey look, we could be standardizing how we think about incidents and keeping track of it somewhere...here Netflix has already given us a structured way to think about it"
- removing barriers to adoption and focusing on progressive enhancement means internal champions can start culture change around incident management in the small
- it enables a team to adopt Dispatch incrementally. I don't think we really use PagerDuty anymore, and I wouldn't want to have a Dispatch conversation tightly-coupled with budget line items for other services

kevgliss:
Thanks for the insight @TheDahv. We are pretty tightly coupled to those services at the moment. But  I agree with what you're saying. 

Internally, we were also discussing the value in setting up a demo instance of dispatch for folks to play with. But that would also be a large amount of work. Trying to weigh what would be most beneficial to most people. 

Unfortunately, Dispatch has a rather high startup cost associated with it today (setting up GSuite, slack etc.).

TheDahv:
Understandable for where we are. I'm happy leaving this as a feature request to be prioritized accordingly. 

I'm eager to champion this at work.. I'll keep an eye on the space and try to push things forward where I can.

lastlee:
[...]

PR #92, opened 2020-03-12
Feature/mock oncall plugin

xdecock:
Well, simply kills the oncall function to allow us to test gitlab plugin, yet, i saw mock where talked about in #55,

this Should solve the problem and let us start without a victor ops plugin. helping testing the other PR faster.


mvilanova:
@xdecock we already have an oncall plugin we use for testing in this [directory](https://github.com/Netflix/dispatch/tree/develop/tests/plugins). In order to avoid adding more mocked plugins, we should move all test plugins under a new directory called `dispatch_default` (or something like that) in the Dispatch plugins [directory](https://github.com/Netflix/dispatch/tree/develop/src/dispatch/plugins) and then import them where needed.

mvilanova:
@xdecock see #100

PR #96, opened 2020-03-12
Fixing issue where stable time would always be overwritten with utcnow()

kevgliss:
When updating the stable_at field, we were always setting this field to be utcnow(), this first checks for an existing stable_at before setting the field.

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -769,15 +769,15 @@ def incident_update_flow(
     # we load the incident instance
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
 
-    if incident_in.type != incident.incident_type.name:
+    if incident_in.incident_type.name != incident.incident_type.name:
         conversation_topic_change = True
 
-    if incident_in.incident_priority != incident.incident_priority.name:
+    if incident_in.incident_priority.name != incident.incident_priority.name:
         conversation_topic_change = True
 
     if notify == "Yes":
         send_incident_change_notifications(
-            incident, incident.title, incident.incident_type, incident.incident_priority
+            incident, incident.title, incident.incident_type.name, incident.incident_priority.name
         )
 
     if conversation_topic_change:
@@ -796,8 +796,8 @@ def incident_update_flow(
         incident.ticket.resource_id,
         title=incident.title,
         description=incident.description,
-        incident_type=incident.incident_type,
-        priority=incident.incident_priority,
+        incident_type=incident.incident_type.name,
+        priority=incident.incident_priority.name,
         commander_email=incident.commander.email,
         conversation_weblink=incident.conversation.weblink,
         document_weblink=incident_document.weblink,
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -41,10 +41,8 @@ from dispatch.document.service import get_by_incident_id_and_resource_type as ge
 from dispatch.group import service as group_service
 from dispatch.group.models import GroupCreate
 from dispatch.incident import service as incident_service
-from dispatch.incident.models import IncidentUpdate
-from dispatch.incident_priority import service as incident_priority_service
+from dispatch.incident.models import IncidentRead
 from dispatch.incident_priority.models import IncidentPriorityRead
-from dispatch.incident_type import service as incident_type_service
 from dispatch.incident_type.models import IncidentTypeRead
 from dispatch.participant import flows as participant_flows
 from dispatch.participant import service as participant_service
@@ -499,30 +497,6 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
     # we load the incident instance
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
 
-    if incident.status == IncidentStatus.active:
-        if command:
-            convo_plugin = plugins.get(INCIDENT_PLUGIN_CONVERSATION_SLUG)
-            convo_plugin.send_ephemeral(
-                command["channel_id"],
-                command["user_id"],
-                "Incident Already Active Notification",
-                blocks=[
-                    {
-                        "type": "section",
-                        "text": {
-                            "type": "plain_text",
-                            "text": "The incident is already active. Aborting command...",
-                        },
-                    }
-                ],
-            )
-        return
-
-    # we update the status of the incident
-    update_incident_status(db_session=db_session, incident=incident, status=IncidentStatus.active)
-
-    log.debug(f"We have updated the status of the incident to {IncidentStatus.active}.")
-
     # we update the status of the external ticket
     update_incident_ticket(
         incident.ticket.resource_id,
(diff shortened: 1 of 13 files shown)


src/dispatch/incident/models.py
```diff
@@ -104,7 +104,8 @@ def last_status_report(self):
     @staticmethod
     def _status_time(mapper, connection, target):
         if target.status == IncidentStatus.stable:
-            target.stable_at = datetime.utcnow()
+            if not target.stable_at:
```
kevgliss: @mvilanova this fixes the immediate issue, but I'm inclined to move this logic to the flows itself (this was a tricky one to track down). Thoughts?

mvilanova: Shouldn't `update` [here](https://github.com/Netflix/dispatch/blob/develop/src/dispatch/incident/views.py#L94) take care of updating the time?

src/dispatch/static/dispatch/src/incident/Table.vue
```diff
@@ -46,9 +46,7 @@
                 <v-icon small class="mr-2" @click="createEditShow(item)">edit</v-icon>
               </template>
               <template v-slot:item.created_at="{ item }">
```
mvilanova: I think this should be `item.reported_at`

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -509,12 +511,39 @@ def command_functions(command: str):
...
+    )
+    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+    existing_incident = IncidentRead.from_orm(incident)
+    incident_service.update(db_session=db_session, incident=incident, incident_in=incident_in)
+    incident_flows.incident_update_flow(user_email, incident_id, existing_incident, notify)
+
+
+@background_task
+def handle_assign_role_action(user_email, incident_id, action, db_session=None):
+    """Messages slack dialog daa into some thing that Dispatch can use."""
```
mvilanova: `data`

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -509,12 +511,39 @@ def command_functions(command: str):
...
+    assignee_role = action["submission"]["role"]
+    assignee_email = get_user_email(client=slack_client, user_id=assignee_user_id)
+    incident_flows.incident_assign_role_flow(user_email, incident_id, assignee_email, assignee_role)
+
+
 def action_functions(action: str):
     """Interprets the action and routes it the appropriate function."""
     action_mappings = {
         SLACK_COMMAND_STATUS_REPORT_SLUG: [status_report_flows.new_status_report_flow],
         SLACK_COMMAND_ASSIGN_ROLE_SLUG: [incident_flows.incident_assign_role_flow],
```
mvilanova: Shouldn't this command be mapped to `handle_assign_role_action`?

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -701,6 +732,7 @@ def verify_timestamp(timestamp: int):
 
     # We resolve the user's email
     user_id = action["user"]["id"]
+    print(user_id)
```
mvilanova: We should remove this `print()`.

src/dispatch/incident/flows.py
```diff
@@ -497,77 +497,21 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
...
-                    {
-                        "type": "section",
-                        "text": {
-                            "type": "plain_text",
-                            "text": "The incident is already active. Aborting command...",
-                        },
-                    }
-                ],
-            )
-        return
```
mvilanova: Are you planning to add this logic to a new function called, for example, `handle_mark_incident_acttive_command` in the views module of the Slack plugin?

src/dispatch/incident/flows.py
```diff
@@ -497,77 +497,21 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
...
-    log.debug(f"We have updated the status of the incident to {IncidentStatus.active}.")
-
     # we update the status of the external ticket
     update_incident_ticket(
         incident.ticket.resource_id,
         incident_type=incident.incident_type.name,
         status=IncidentStatus.active.lower(),
     )
 
-    log.debug(f"We have updated the status of the external ticket to {IncidentStatus.active}.")
```
mvilanova: Any reason to not keep this one?

src/dispatch/incident/flows.py
```diff
@@ -497,77 +497,21 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
...
-                    {
-                        "type": "section",
-                        "text": {
-                            "type": "plain_text",
-                            "text": "The incident is already stable. Aborting command...",
-                        },
-                    }
-                ],
-            )
-            return
```
mvilanova: Same comment as the one for the active flow.

src/dispatch/plugins/dispatch_slack/config.py
```diff
@@ -29,8 +29,8 @@
 SLACK_COMMAND_ASSIGN_ROLE_SLUG = config(
     "SLACK_COMMAND_ASSIGN_ROLE_SLUG", default="/dispatch-assign-role"
 )
-SLACK_COMMAND_EDIT_INCIDENT_SLUG = config(
-    "SLACK_COMMAND_EDIT_INCIDENT_SLUG", default="/dispatch-edit-incident"
+SLACK_COMMAND_UPDATE_INCIDENT_SLUG = config(
```
mvilanova: Note to update command name in the Slack UI, and Dispatch documentation.

src/dispatch/incident/flows.py
```diff
@@ -497,77 +497,21 @@ def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_se
...
-    log.debug(f"We have updated the status of the external ticket to {IncidentStatus.active}.")
-
-    # we update the conversation topic
-    set_conversation_topic(incident)
-
-    # we send the active notifications
-    send_incident_status_notifications(incident, db_session)
-
-    log.debug("We have sent the incident active notifications.")
-
```
mvilanova: When are we calling these functions when we call this flow directly from the Slack code when we run `/dispatch-mark-active`?

src/dispatch/incident/flows.py
```diff
@@ -584,9 +528,6 @@ def incident_stable_flow(incident_id: int, command: Optional[dict] = None, db_se
 
     log.debug(f"We have updated the status of the external ticket to {IncidentStatus.stable}.")
 
-    # we update the conversation topic
-    set_conversation_topic(incident)
-
```
mvilanova: Where are we updating the topic now? We also call this flow directly from the Slack code when run `/dispatch-mark-stable`.

src/dispatch/incident/flows.py
```diff
@@ -676,27 +614,7 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
...
-                command["user_id"],
-                "Mark Incident Stable Notification",
-                blocks=[
-                    {
-                        "type": "section",
-                        "text": {"type": "plain_text", "text": "Marking the incident as stable..."},
-                    }
-                ],
-            )
-        incident_stable_flow(incident_id, command=command, db_session=db_session)
```
mvilanova: Same as active and stable flows, plus we need to run the stable flow if incident status is active.

src/dispatch/incident/flows.py
kevgliss: These are now handled by the `update` command.

src/dispatch/incident/flows.py
kevgliss: Same as above.

src/dispatch/incident/flows.py
kevgliss: Yes, we could add it to the handle_mark_incident_active, or we could centralize it based on a signal/source.

src/dispatch/incident/flows.py
kevgliss: No, I will put it back in.

src/dispatch/incident/flows.py
mvilanova: We are not calling the `update_flow` function when we run the command. Same for the stable and closed ones. See https://github.com/Netflix/dispatch/blob/develop/src/dispatch/plugins/dispatch_slack/views.py#L504

mvilanova: We could add status to the update dialog and get rid of the mark [active|stable|closed] commands like we've discussed in the past.

src/dispatch/incident/flows.py
kevgliss: Same as others stable flow should run if `update_flow` thinks it needs to.

PR #102, opened 2020-03-13
change accessToken to idToken, add audience param for jwt_decode, add…

timwheelercom:
- utilize `idToken` (JWT) rather than `accessToken` as auth provider was failing to decode `accessToken` when calling `jwt_decode()`
 - add `DISPATCH_CLIENT_ID` which is needed for the `audience` attribute in jwt.decode according to docs: https://python-jose.readthedocs.io/en/latest/jwt/api.html
> audience (str) – The intended audience of the token. If the “aud” claim is included in the claim set, then the audience must be included and must equal the provided claim.


src/dispatch/auth/service.py
```diff
@@ -28,7 +28,7 @@ def from_bearer_token(request: Request):
         jwk_key_cache[JWKS_URL] = key
 
     try:
-        data = jwt.decode(token, key)
+        data = jwt.decode(token, key, audience=DISPATCH_CLIENT_ID)
```
timwheelercom: particularly when using Auth0 I needed to provide the `audience` arg

src/dispatch/static/dispatch/src/auth/store.js
```diff
@@ -38,20 +38,20 @@ const actions = {
 }
 
 const mutations = {
-  SET_USER_LOGIN(state, accessToken) {
-    state.accessToken = accessToken
+  SET_USER_LOGIN(state, idToken) {
+    state.idToken = idToken
     state.status = { loggedIn: true }
-    state.userInfo = jwt_decode(accessToken)
+    state.userInfo = jwt_decode(idToken)
```
timwheelercom: `idToken` is the actual JWT that needs to be decoded, not the `accessToken`

kevgliss:
Thanks for the PR! Are you using Auth0? I need to look at this closely because it seems the `PCKE` "standard" seems to leave a lot up to the implementation. My hope is that largely just naming that differs.

timwheelercom:
@kevgliss I went with Auth0 to get the project up and running. According to [RFC-7519 spec](https://tools.ietf.org/html/rfc7519#section-4.1.3):

> The "aud" (audience) claim identifies the recipients that the JWT is
   intended for.  Each principal intended to process the JWT MUST
   identify itself with a value in the audience claim.  If the principal
   processing the claim does not identify itself with a value in the
   "aud" claim when this claim is present, then the JWT MUST be
   rejected.

This is the error I was encountering as the Auth0 JWKS returned a JWT that contained an `aud` claim (which is ultimately the client id), but since it was not provided when attempting to decode said JWT token it was rejected.

kevgliss:
Took a look at this, according to https://auth0.com/docs/tokens we do not want to use `idToken` and instead of us `accessTokens`.

> ID Tokens should not be used to gain access to an API. Each token contains information for the intended audience (which is usually the recipient).

> Access Tokens (which aren't always JWTs) are used to inform an API that the bearer of the token has been authorized to access the API and perform a predetermined set of actions (specified by the scopes granted).

From there server's perspective, we only want to deal with access tokens, letting the client and the provider do the appropriate handshakes in order to provide a valid access token.

PR #103, opened 2020-03-13
Adds messaging for ChatOps commands

mvilanova:
Adds message block to welcome and incident resource messages with description and link to reference document with the goal of making ChatOps commands easier to discover by incident participants.


src/dispatch/incident/flows.py
```diff
@@ -402,7 +404,19 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
         "resource_type": INCIDENT_RESOURCE_FAQ_DOCUMENT,
     }
 
-    for d in [incident_document, incident_sheet, faq_document]:
+    conversation_commands_reference_document = {
+        "name": "Incident FAQ",
```
kevgliss: Shouldn't this be `Command Reference`?

mvilanova: Yes, will fix.

PR #112, opened 2020-03-17
Deprecates dispatch-mark-* commands in favor of dispatch-update-incident

kevgliss:


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -501,9 +517,9 @@ def command_functions(command: str):
         SLACK_COMMAND_LIST_PARTICIPANTS_SLUG: [list_participants],
         SLACK_COMMAND_LIST_RESOURCES_SLUG: [incident_flows.incident_list_resources_flow],
         SLACK_COMMAND_LIST_TASKS_SLUG: [list_tasks],
-        SLACK_COMMAND_MARK_ACTIVE_SLUG: [incident_flows.incident_active_flow],
-        SLACK_COMMAND_MARK_CLOSED_SLUG: [incident_flows.incident_closed_flow],
-        SLACK_COMMAND_MARK_STABLE_SLUG: [incident_flows.incident_stable_flow],
+        SLACK_COMMAND_MARK_ACTIVE_SLUG: [],
```
forestmonster: We should remove these later if we don't end up using them.

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -515,12 +531,14 @@ def command_functions(command: str):
 def handle_update_incident_action(user_email, incident_id, action, db_session=None):
     """Messages slack dialog data into something that Dispatch can use."""
     submission = action["submission"]
+    print(submission)
```
mvilanova: Remove

PR #115, opened 2020-03-17
Adds visibility column to the incident type model

mvilanova:
Note: these changes require running `dispatch database upgrade` to add the new visibility column to the database.
Screen Shot 2020-03-17 at 4 27 56 PM


src/dispatch/static/dispatch/src/incident_type/NewEditSheet.vue
```diff
@@ -1,107 +1,123 @@
...
+                    required
+                  />
+                </ValidationProvider>
+              </v-flex>
+              <v-flex xs12>
+                <v-select
+                  v-model="visibility"
+                  label="Visibility"
+                  :items="visibilities"
+                  hint="A visibility for your incident type"
```
kevgliss: Maybe: `A default visibility for your incident type.`

PR #118, opened 2020-03-18
Update costs no matter the status.

kevgliss:
The actual query should be pretty cheap (on a five minute interval), but we make sure that we don't update external resources if the cost hasn't changed.

### DIFF
--- a/src/dispatch/incident/scheduled.py
+++ b/src/dispatch/incident/scheduled.py
@@ -27,7 +27,12 @@ from dispatch.scheduler import scheduler
 from dispatch.service import service as service_service
 
 from .enums import IncidentStatus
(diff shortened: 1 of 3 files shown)


src/dispatch/incident/scheduled.py
```diff
@@ -218,15 +223,21 @@ def daily_summary(db_session=None):
 
 @scheduler.add(every(5).minutes, name="calculate-incident-cost")
 @background_task
-def active_incidents_cost(db_session=None):
-    """Calculates the cost of all active incidents."""
-    active_incidents = get_all_by_status(db_session=db_session, status=IncidentStatus.active)
+def incidents_cost(db_session=None):
```
mvilanova: nit: rename to `calculate_incidents_cost` and update the scheduler job name.

PR #119, opened 2020-03-18
Ensuring that the IC gets a status report reminder on status change.

kevgliss:


src/dispatch/incident/flows.py
```diff
@@ -621,6 +624,7 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
     """Runs the incident closed flow."""
     # we load the incident instance
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+    send_incident_status_report_reminder(incident)
```
mvilanova: They won't be able to. The channel will be closed. Should we remove it?

PR #128, opened 2020-03-20
Running incident status change flows based on new status, not previous

mvilanova:


mvilanova:
Flake8 error is fixed in #127

src/dispatch/incident/flows.py
```diff
@@ -751,13 +755,13 @@ def incident_update_flow(
...
-        if previous_incident.status == IncidentStatus.active:
+    if previous_incident.status.value != incident.status:
+        if incident.status == IncidentStatus.active:
             incident_active_flow(incident_id=incident.id, db_session=db_session)
-        elif previous_incident.status == IncidentStatus.closed:
-            incident_closed_flow(incident_id=incident.id, db_session=db_session)
-        elif previous_incident.status == IncidentStatus.stable:
+        elif incident.status == IncidentStatus.stable:
             incident_stable_flow(incident_id=incident.id, db_session=db_session)
+        elif incident.status == IncidentStatus.closed:
```
kevgliss: nit: Do you think it's more clear to call the stable flow here if necessary (current status is active) or to have them nested (within closed). Personally, I'm not a huge fan of overly nested flows.

mvilanova: Yeah, better in this flow now that you're suggesting it. I made the change when I resolve the conflict.

PR #132, opened 2020-03-20
Adds conference and google meet support for all new incidents.

kevgliss:
Adds support for automatically creating a video conference on incident creation (default is google meet).


willbengtson:
+1

src/dispatch/conference/service.py
```diff
@@ -0,0 +1,36 @@
...
+    return db_session.query(Conference).filter(Conference.resource_id == resource_id).one_or_none()
+
+
+def get_by_resource_type(*, db_session, resource_type: str) -> Optional[Conference]:
+    return (
+        db_session.query(Conference).filter(Conference.resource_type == resource_type).one_or_none()
+    )
+
+
+def get_by_channel_id(db_session, conference_id: str) -> Optional[Conference]:
```
mvilanova: `get_by_conference_id()`?

src/dispatch/conference/service.py
```diff
@@ -0,0 +1,36 @@
...
+    return (
+        db_session.query(Conference).filter(Conference.conference_id == conference_id).one_or_none()
+    )
+
+
+def get_all(*, db_session):
```
mvilanova: Add return value `List[Optional[Conference]]`

src/dispatch/conference/service.py
```diff
@@ -0,0 +1,36 @@
...
+    return db_session.query(Conference)
+
+
+def create(*, db_session, **kwargs) -> Conference:
```
mvilanova: We should try to use the `ConferenceCreate` model here.

src/dispatch/incident/flows.py
```diff
@@ -413,6 +431,20 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
         "resource_type": INCIDENT_RESOURCE_CONVERSATION_COMMANDS_REFERENCE_DOCUMENT,
     }
 
+    conference = create_conference(incident, [tactical_group["email"]])
+
+    log.debug("Conference created. Tacticle group added.")
```
mvilanova: Tactical

src/dispatch/incident/flows.py
```diff
@@ -413,6 +431,20 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
+
+    incident.conference = conference_service.create(
```
mvilanova: Same comment about using the pydantic model.

src/dispatch/conference/service.py
kevgliss: It returns a query object not a list of objects.

PR #138, opened 2020-03-20
Fix dump database to use DATABASE_NAME from config

diogomatsubara:
This small patch fixes the database dump command to use the configured DATABASE_NAME instead of the hard coded `dispatch` name.


kevgliss:
Much better. Thanks!

Issue #95, opened 2020-03-12
Support Basic Authentication


kevgliss:
For new folks either evaluating dispatch or simply not in the position to set up oauth flows, we should support basic username/password authentication. 

Roughly speaking this will require:

- [x] Create User table
- [x] Create /login endpoint
- [x] Add login UI

abinmn:
Can I take up this issue? I mean the backend part.

kevgliss:
That would be great, I just put together #139 that introduces the concept of authentication plugins. It would be great if you could take that system for a spin and see if I missed anything. 

Fast api also has some great starting points if you're looking for inspiration: 
https://fastapi.tiangolo.com/tutorial/security/get-current-user/

sarahcthekey:
@kevgliss Sorry if I'm asking a silly question, I can see from various PR's and discussions it's possible to set up without authentication for a POC. Is that set up out of the box now or do I need to make changes, if so is this just an enable/disable env variable?

I was actually working with the dispatch-docker repo but came across here when struggling to add things to the admin UI, I'm getting the error '. Reason: Could not validate credentials'.

The quicker I can get up and running with this the better so any help much appreciated :-)

kevgliss:
@sarahcthekey the default is with PKCE authentication... but it can easily be disabled by setting `DISPATCH_AUTHENTICATION_PROVIDER_SLUG=""`. Apologies, the documentation for this needs to be updated.

sarahcthekey:
@kevgliss OK thanks, I have added this, unfortunately now receiving a different error when I try to add a new service (or other inputs):

`IncidentType not created. Reason: undefined
`

stefanm8:
Hello @kevgliss,

Is anybody still working on this? if not I would like to take this one.

if we would implement basic authentication would this make the app work without setting up google accounts or any other ENV variables besides the necessary ones such as PSQL connection?

kevgliss:
@stefanm8 you will still require google credentials if you wish to use the Google related plugins.

stefanm8:
Thank you for fast reply.

I intend to develop plugins to replace google suite without having to configure google account.

stefanm8:
Hello been working on this and currently I have setup:
- Default JWT auth
- User table
- Login/register endpoints 

I've seen that you also implement a plugin base for providing auth plugin.

https://github.com/stefanm8/dispatch

you may see the code at src/dispatch/auth/*

PR #139, opened 2020-03-21
Refactors how authentication providers are managed.

kevgliss:
This PR accomplished two main goals:

1) Make it easier to add new ways of adding and manging authentication providers
2) Decouples our frontend code from the PKCE auth flow

Together, both of these changes should make it much easier to enable and support different auth flow, including disabling auth altogether. See #95 

We also introduce an authentication plugin system via the `AuthenticationProviderPlugin` base class. This is a simple interface that additional backend authentication flows can leverage. From a user's perspective swapping out authentication providers should be as easy as modifying the new config variable `DISPATCH_AUTHENTICATION_PROVIDER_SLUG` (we keep pkce as the default, setting this to "" disables all auth).

cc @willbengtson


diogomatsubara:
I have a branch based on this deployed at https://master-7rqtwti-2hounbfgfsnu6.eu-3.platformsh.site/dashboard with authentication disabled if people want to try it out and give feedback.

kevgliss:
@diogomatsubara that's awesome! We were thinking about hosting it somewhere for folks to play with but hadn't got a chance. I'm also planning to generate some fake data so the demo is a bit more interesting.

diogomatsubara:
@kevgliss if you land some fake data I'd be happy to keep that instance updated for others to play with. I been trying to put some data in manually but that's get tedious really fast. :-)

src/dispatch/config.py
```diff
@@ -44,19 +44,24 @@ def __str__(self) -> str:
...
 # authentication
-DISPATCH_PKCE_AUTH_ENABLED = config("DISPATCH_PKCE_AUTH", cast=bool, default=True)
-DISPATCH_HEADER_AUTH_ENABLED = config("DISPATCH_HEADER_AUTH", cast=bool, default=False)
-
-DISPATCH_AUTH_HEADER_KEY = config("DISPATCH_AUTH_HEADER", default=None)
-JWKS_URL = config("JWKS_URL", default=None)
+DISPATCH_AUTHENTICATION_PROVIDER_SLUG = config(
+    "DISPATCH_AUTHENTICATION_PROVIDER_SLUG", default="dispatch-auth-provider-pkce"
+)
+VUE_APP_DISPATCH_AUTHENTICATION_PROVIDER_SLUG = DISPATCH_AUTHENTICATION_PROVIDER_SLUG
```
mvilanova: Should we get rid of `VUE_APP_DISPATCH_AUTHENTICATION_PROVIDER_SLUG` if we're always setting it to `DISPATCH_AUTHENTICATION_PROVIDER_SLUG`, and use `DISPATCH_AUTHENTICATION_PROVIDER_SLUG` instead?

kevgliss: Unfortunately, Vue will only pick up envvars prefixed with `VUE_APP` to be injected into the vue app. I think this is the best of both worlds (only having to set one var).

PR #142, opened 2020-03-22
skip db creation in init if already exist

diogomatsubara:
I'm experimenting with dispatch for internal use at [work](https://www.platform.sh). We want to be able to deploy it in our own infrastructure and I ran into an issue during db initialization. The `dispatch database initi` code assumes the user has permissions to create/delete databases. In a PaaS like Platform.sh, you already start with an existing database and the user doesn't have permission to create new ones. Only to create/delete tables. This patch makes database init work during deployment in such case. Do you think this is the right thing to do? If you would prefer another direction, please let me know. I can further describe the use case as well.

Any feedback much appreciated! Thanks for the work on this.

### DIFF
--- a/src/dispatch/cli.py
+++ b/src/dispatch/cli.py
@@ -489,9 +489,10 @@ def database_trigger_sync():
 @dispatch_database.command("init")
 def init_database():
     """Initializes a new database."""
-    from sqlalchemy_utils import create_database
+    from sqlalchemy_utils import create_database, database_exists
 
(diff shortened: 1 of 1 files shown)


diogomatsubara:
FWIW I have a branch deployed at https://master-7rqtwti-2hounbfgfsnu6.eu-3.platformsh.site/dashboard, based on `feature/current-user-providers` + my modifications to adapt the build/deployment on our PaaS

kevgliss:
Sounds reasonable to me, thanks for the PR!

PR #145, opened 2020-03-23
Fixed 'pkce_auth' always set to 'true', ignoring env var 'VUE_APP_DISPATCH_PKCE_AUTH'

### LINKED ISSUE #144: 'pkce_auth' always set to 'true'
https://github.com/Netflix/dispatch/blob/4057b8e9860c28cd224b0bb461ff0ffc44063488/src/dispatch/static/dispatch/src/router/index.js#L25

setting default value not working like that for booleans. https://stackoverflow.com/questions/15464169/how-to-set-default-boolean-values-in-javascript

mogindi:
Closes #144.


diogomatsubara:
This code has been changed in #139 and once that lands it makes disabling authentication a lot easier.

PR #146, opened 2020-03-23
Doc fixes.

diogomatsubara:
* Fix google config doc to be consistent with var names in code.
* Fix jira config doc to signal sensitive var in password.
* Fix slack config doc to point to correct update incident endpoint.


kevgliss:
Thanks!

PR #147, opened 2020-03-23
Adding some vscode settings

kevgliss:


kevgliss:
@willbengtson I'm not sure if people set python.pythonPath in their project specific settings often, personally, I define it dynamically in my user settings.json (via something like: `"python.pythonPath": "/home/kglisson/.pyenv/versions/${workspaceFolderBasename}/bin/python",`)

PR #149, opened 2020-03-23
[WIP] Add a data-only dump to the tree for feedback

diogomatsubara:
Add initial commit with a data-only dump of demo dispatch

### DIFF
--- a/src/dispatch/cli.py
+++ b/src/dispatch/cli.py
@@ -503,11 +503,20 @@ def init_database():
 
 
 @dispatch_database.command("restore")
-def restore_database():
+@click.option(
+    "--data-only", help="Restore only the data in the dump.", default=False)
+@click.option(
+    "--dump-file", default='dispatch-backup.dump', help="Path to a PostgreSQL dump file.")
+def restore_database(dump_file, data_only=False):
     """Restores the database via pg_restore."""
     import sh
     from sh import psql, createdb
-    from dispatch.config import DATABASE_HOSTNAME, DATABASE_PORT, DATABASE_CREDENTIALS
+    from dispatch.config import (
+        DATABASE_HOSTNAME,
+        DATABASE_NAME,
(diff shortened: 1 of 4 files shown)


kevgliss:
Thanks for this, wondering where we should store this? I'm inclined to put this in the https://github.com/Netflix/dispatch-docker repo instead of the main application repo (with a docs on how to restore). 

What do you think?

diogomatsubara:
For me would be more useful to have it in this repo as I'm not using the dispatch-docker repo. I adapted the build configuration to be able to deploy on Platform.sh. I was thinking maybe I could add an additional `database` command to load sample data. That way for people trying they can just run `dispatch database load_sample_data` or something like that and they have an environment somewhat useful to play with. Would that be a good solution?

kevgliss:
I think that's fine, but I would at least want to create a `data` directory for it.

src/dispatch/cli.py
```diff
@@ -568,6 +568,10 @@ def dump_database():
         _env={"PGPASSWORD": password},
     )
 
+@dispatch_database.command("load_sample_data")
+def load_sample_data():
+    """Load sample data into the database."""
+    pass
```
kevgliss: Should we reuse the `restore` command/function with a different path?

diogomatsubara: I made changes to re-use the `restore` command. My sample data is still not working and I think I'll have to make some edits to it. I'll ping you here once I have something working. Thanks for the feedback!

kevgliss:
I'm going to merge this in as I think it will be helpful to lots of folks just trying to experiment with dispatch.

sakshigupta3:
I don't see any such option, "dispatch database load_sample_data" to load sample data. Was the design changed or could you please add a section for this in https://hawkins.gitbook.io/dispatch/cli

PR #152, opened 2020-03-23
Add KMS ability for Secret protection in env

willbengtson:
Adds the ability to encrypt environment secrets with KMS, similar to metatron.  Application would be responsible for setting `AWS_DEFAULT_REGION` environment variable.


kevgliss:
Looks good, the next one we add we should make it into a plugin though.

PR #161, opened 2020-03-24
Allows tags and terms to be associated with incidents via the admin UI.

kevgliss:


src/dispatch/incident/models.py
```diff
@@ -152,13 +162,15 @@ class IncidentCreate(IncidentBase):
...
-    visibility: Visibility
     incident_priority: IncidentPriorityBase
     incident_type: IncidentTypeBase
     reported_at: Optional[datetime] = None
     stable_at: Optional[datetime] = None
     commander: Optional[IndividualReadNested]
     reporter: Optional[IndividualReadNested]
+    tags: Optional[List[Any]] = []  # any until we figure out circular imports
+    terms: Optional[List[Any]] = []  # any until we figure out circular imports
+    visibility: Visibility
```
mvilanova: Not needed here if already in `IncidentBase`. Same in `IncidentRead`.

PR #163, opened 2020-03-24
Feature/incident forecast

kevgliss:
Initial work on using fbprophet for incident forecasting.

### DIFF
--- a/src/dispatch/incident/metrics.py
+++ b/src/dispatch/incident/metrics.py
@@ -73,4 +73,4 @@ def make_forecast(
     future = forecaster.make_future_dataframe(periods=periods, freq="M")
     forecast = forecaster.predict(future)
 
-    return forecast
+    return forecast.to_dict("series")
--- a/src/dispatch/incident/metrics.py
+++ b/src/dispatch/incident/metrics.py
@@ -1,53 +1,76 @@
+import json
 from itertools import groupby
 
 from datetime import date
+from dateutil.relativedelta import relativedelta
+
 from calendar import monthrange
 
 from fbprophet import Prophet
 import pandas as pd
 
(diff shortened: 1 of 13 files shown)


src/dispatch/incident/metrics.py
```diff
@@ -0,0 +1,53 @@
...
+    for (last_day, items) in groupby(incidents_sorted, grouper):
+        dataframe_dict["ds"].append(last_day)
+        dataframe_dict["y"].append(len(list(items)))
+
+    dataframe = pd.DataFrame.from_dict(dataframe_dict)
+
+    forecaster = Prophet()
+    forecaster.fit(dataframe, algorithm="LBFGS")
+
+    future = forecaster.make_future_dataframe(periods=2, freq="M")
```
snkilmartin: I'm currently using 24 for periods

src/dispatch/incident/metrics.py
```diff
@@ -0,0 +1,53 @@
...
+
+    forecast = forecaster.predict(future)
+
+    return forecast
```
snkilmartin: The [Prophet docs](https://facebook.github.io/prophet/docs/quick_start.html#python-api) are a great reference; this line from the Prophet docs might be good to include in the Dispatch docs:

> The forecast object here is a new dataframe that includes a column yhat with the forecast, as well as columns for components and uncertainty intervals.

kevgliss: Thank you! I was going to ask what all of the output values mean.

src/dispatch/incident/metrics.py
kevgliss: Yes, I will make this configurable and default to 24.

PR #170, opened 2020-03-27
Add Zoom conference plugin.

willbengtson:
Adds a Zoom conference plugin.

- Modifies the conference DB to allow a challenge code for zoom conference
- Updates the google meet conference plugin to send "" for challenge as it's not needed
- Set's all existing conference challenges fields in DB to N/A
- Updates messaging to include the conference challenge


src/dispatch/alembic/versions/5d4dee3e24fc_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
mvilanova: nit: add a message (e.g. adds column for conference challenge)

src/dispatch/messaging.py
```diff
@@ -280,7 +280,7 @@ class MessageType(str, Enum):
 INCIDENT_CONFERENCE = {
     "title": "Incident Conference",
     "title_link": "{{conference_weblink}}",
-    "text": INCIDENT_CONFERENCE_DESCRIPTION,
+    "text": INCIDENT_CONFERENCE_DESCRIPTION + " Password: {{conference_challenge}}",
```
mvilanova: Can you add the password in the `INCIDENT_CONFERENCE_DESCRIPTION`?

src/dispatch/alembic/versions/5d4dee3e24fc_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '5d4dee3e24fc'
+down_revision = '8b67c774279d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('conference', sa.Column('conference_challenge', sa.String(), server_default='N/A', nullable=False))
```
kevgliss: Why not let this be null? Null will return `None` which can be checked more easily that the string value of "N/A"

src/dispatch/incident/messaging.py
```diff
@@ -338,6 +340,8 @@ def send_incident_update_notifications(incident: Incident, previous_incident: In
                 ticket_weblink=incident.ticket.weblink,
                 faq_weblink=incident.incident_faq.weblink,
                 incident_id=incident.id,
+                incident_priority_old=previous_incident.incident_priority.name,
+                incident_priority_new=incident.incident_priority.name,
```
mvilanova: Without looking at the `notification_template`, looks like the following would also be missing:
```
incident_type_old=previous_incident.incident_type.name,
incident_type_new=incident.incident_type.name,
incident_status_old=previous_incident.status.value,
incident_status_new=incident.status,
```

src/dispatch/plugins/dispatch_zoom/plugin.py
```diff
@@ -0,0 +1,106 @@
...
+    title = "Zoom - Conference"
+    slug = "zoom-conference"
+    description = "Uses Zoom to manage conference meetings."
+
+    author = "Will Bengtson"
+    author_url = "https://github.com/netflix/dispatch.git"
+
+    # def __init__(self):
+    #     self.scopes = ["https://www.googleapis.com/auth/calendar"]
+
```
mvilanova: Remove

src/dispatch/incident/messaging.py
mvilanova: I'll take a look at it today.

src/dispatch/conference/models.py
```diff
@@ -11,6 +11,7 @@
 class Conference(Base, ResourceMixin):
     id = Column(Integer, primary_key=True)
     conference_id = Column(String)
+    conference_challenge = Column(String, nullable=False, server_default="N/A")
```
kevgliss: I'm not sure there is a huge value to using `server_default` vs `default` here.

src/dispatch/plugins/dispatch_zoom/plugin.py
```diff
@@ -0,0 +1,106 @@
...
+@apply(counter, exclude=["__init__"])
+class ZoomConferencePlugin(ConferencePlugin):
+    title = "Zoom - Conference"
+    slug = "zoom-conference"
+    description = "Uses Zoom to manage conference meetings."
+
+    author = "Will Bengtson"
+    author_url = "https://github.com/netflix/dispatch.git"
+
+    # def __init__(self):
```
kevgliss: Remove

src/dispatch/plugins/dispatch_zoom/plugin.py
```diff
@@ -0,0 +1,106 @@
...
+.. moduleauthor:: Will Bengtson <wbengtson@hashicorp.com>
+"""
+import logging
+import time
+import uuid
+import random
+from datetime import datetime, timedelta
+from typing import Any, List
+
+from googleapiclient.errors import HttpError
```
mvilanova: Remove this and any other imports that are not being used.

src/dispatch/plugins/dispatch_zoom/plugin.py
```diff
@@ -0,0 +1,106 @@
...
+            "id": conference_json.get('id', '1'),
+            "challenge": conference_json.get('password', '123')
+        }
+
+    def delete(self, event_id: str):
+        """Deletes an existing event."""
+        client = ZoomClient(ZOOM_API_KEY, str(ZOOM_API_SECRET))
+        delete_meeting(client, event_id)
+
+    def add_participant(self, event_id: str, participant: str):
```
kevgliss: How are folks added to a zoom? Anyone with the link and passcode? Are there private rooms that need an email?

src/dispatch/alembic/versions/5d4dee3e24fc_.py
willbengtson: when this was Null, there are errors on the loading up the incident table in the UI, so I set this to "N/A" instead

src/dispatch/alembic/versions/5d4dee3e24fc_.py
willbengtson: Oops, good catch.

src/dispatch/messaging.py
willbengtson: Sure

src/dispatch/incident/messaging.py
willbengtson: I'll take a look as well.  Only noticed that priority update emails had "** to **" instead of something like "*Low* to *Medium*"

src/dispatch/plugins/dispatch_zoom/plugin.py
willbengtson: Good catch

src/dispatch/conference/models.py
willbengtson: When I had default, it didn't migrate the existing conference rows and set a default value which then broke the UI.

src/dispatch/plugins/dispatch_zoom/plugin.py
willbengtson: 👍

src/dispatch/plugins/dispatch_zoom/plugin.py
willbengtson: Anyone with link and password.  Participants are only in webinars and things like that.

src/dispatch/plugins/dispatch_zoom/client.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, api_key, api_secret):
+        self.api_key = api_key
+        self.api_secret = api_secret
+        self.token = generate_jwt(self.api_key, self.api_secret)
+        self.headers = self._get_headers()
+        self.timeout = 15
+
+    def _get_headers(self):
+        headers = {}
+        headers["authorization"] = "Bearer {}".format(self.token)
```
mvilanova: nit: use `f-strings` instead

src/dispatch/incident/messaging.py
willbengtson: Just tested, will add these as well

mvilanova: Awesome, thanks!

src/dispatch/plugins/dispatch_zoom/client.py
willbengtson: 👍

src/dispatch/incident/messaging.py
mvilanova: Looking at `convo_plugin.send()` above, we can probably remove the following for this notification:
```
priority=incident.incident_priority.name,
document_weblink=incident.incident_document.weblink,
storage_weblink=incident.storage.weblink,
faq_weblink=incident.incident_faq.weblink,
```

mvilanova:
@willbengtson forgot to ask, please document the plugin when you have spare cycles. Thanks

PR #175, opened 2020-03-27
Allowing some heavy weight metric libraries to be opt-in.

kevgliss:
This allows our heavy weight metric dependencies to be optionally installed via `pip install -e .[metrics]`

### DIFF
--- a/requirements-metrics.txt
+++ b/requirements-metrics.txt
@@ -1,0 +1,3 @@
+numpy
+pystan
+fbprophet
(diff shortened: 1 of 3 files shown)


requirements-metrics.txt
```diff
@@ -0,0 +1,3 @@
+numpy
+pystan
+fbprophet
```
mvilanova: We should remove them from `requirements-base.txt`, which are commented at the moment.

PR #184, opened 2020-03-30
Fixes rendering the incident conference description

mvilanova:


mvilanova:
cc @willbengtson

willbengtson:
👍

mvilanova:
I thought that was already the case?

kevgliss:
Sorry @mvilanova you're correct, I mis-read the pydantic model as the sqlalchemy model.

PR #185, opened 2020-03-31
Incident Timeline Support (Part I)

mvilanova:
This PR adds the model, service, and tests.


src/dispatch/event/models.py
```diff
@@ -0,0 +1,68 @@
...
+from sqlalchemy import Column, Integer, String, ForeignKey, PrimaryKeyConstraint, Table
+from sqlalchemy.orm import relationship
+from sqlalchemy_utils import TSVectorType
+
+from dispatch.database import Base
+from dispatch.models import DispatchBase, TimeStampMixin
+from dispatch.plugins.base import plugins
+
+
+EventSource = Enum(
```
mvilanova: Think about other sources and add them to the Enum.

kevgliss: Why are these enums? Why can't we just use the slug strings themselves? Likely, we will have many different plugins adding things as events and will not want to manage enums.

mvilanova: It's only one `Enum`. The idea was to have a structure with all the different sources of events that we can use in different parts of the code.

kevgliss: If it's only one then I think a CONSTANT is better than an enum.

PR #187, opened 2020-04-01
Adds 'Get Involved' button on all incident conversation messages.

kevgliss:
Additionally this PR moves the logic for restricting joining closed incidents to the action handler and displays an error to the user.


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -476,15 +476,22 @@ def create_status_report_dialog(incident_id: int, command: dict = None, db_sessi
...
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
 
-    user_id = dispatch_slack_service.resolve_user(slack_client, user_email)["id"]
-
-    dispatch_slack_service.add_users_to_conversation(
-        slack_client, incident.conversation.channel_id, user_id
-    )
+    print(action)
+    if incident.status == IncidentStatus.closed:
+        message = f"Sorry we cannot add you to a closed incident. Please reach out to the incident commander ({incident.commander.name}) for details."
```
forestmonster, line 487: Nit: 
```suggestion
        message = f"Sorry, we cannot add you to a closed incident. Please reach out to the incident commander ({incident.commander.name}) for details."
```

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -476,15 +476,22 @@ def create_status_report_dialog(incident_id: int, command: dict = None, db_sessi
...
+):
     """Adds a user to a conversation."""
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
 
-    user_id = dispatch_slack_service.resolve_user(slack_client, user_email)["id"]
-
-    dispatch_slack_service.add_users_to_conversation(
-        slack_client, incident.conversation.channel_id, user_id
-    )
+    print(action)
```
mvilanova, line 485: Remove.

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -756,7 +763,7 @@ def verify_timestamp(timestamp: int):
 
     # Dispatch action functions to be executed in the background
     for f in action_functions(action_name):
-        background_tasks.add_task(f, user_email, incident_id, action)
+        background_tasks.add_task(f, user_id, user_email, incident_id, action)
```
mvilanova, line 766: Should `user_id` be a kwarg? It is not used in a lot of functions that are added as background tasks (e.g. `new_status_report_flow()`).

src/dispatch/messaging.py
```diff
@@ -364,9 +367,7 @@ class MessageType(str, Enum):
 INCIDENT_NOTIFICATION_COMMON = [INCIDENT_NAME, INCIDENT_TITLE]
 
 INCIDENT_NOTIFICATION = INCIDENT_NOTIFICATION_COMMON.copy()
-INCIDENT_NOTIFICATION.extend(
-    [INCIDENT_STATUS, INCIDENT_PRIORITY_FYI, INCIDENT_COMMANDER, INCIDENT_GET_INVOLVED_BUTTON]
```
mvilanova, line 368: Remove `INCIDENT_GET_INVOLVED_BUTTON` if we're not going to use it anymore.

src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -476,15 +476,22 @@ def create_status_report_dialog(incident_id: int, command: dict = None, db_sessi
...
-    )
+    print(action)
+    if incident.status == IncidentStatus.closed:
+        message = f"Sorry we cannot add you to a closed incident. Please reach out to the incident commander ({incident.commander.name}) for details."
+        dispatch_slack_service.send_ephemeral_message(
+            slack_client, action["container"]["channel_id"], user_id, message
+        )
+    else:
+        dispatch_slack_service.add_users_to_conversation(
+            slack_client, incident.conversation.channel_id, user_id
```
mvilanova, line 493: `add_users_to_conversation()` expects a list of users, not a `str`.

src/dispatch/messaging.py
kevgliss: I left it as I thought it might still be useful in the future. But will delete.

src/dispatch/plugins/dispatch_slack/views.py
kevgliss: It's not optional for the functions that do use it, so I don't think a kwarg makes sense.

src/dispatch/messaging.py
kevgliss: Actually, this commit does remove the button?

mvilanova: I meant, removing the `INCIDENT_GET_INVOLVED_BUTTON` `dict` and the `INCIDENT_GET_INVOLVED_BUTTON_DESCRIPTION`, but if you want to keep them in case they're needed in the future, fine by me.

PR #191, opened 2020-04-01
Adds get involved button to daily summary emails. 

kevgliss:

### DIFF
--- a/src/dispatch/incident/scheduled.py
+++ b/src/dispatch/incident/scheduled.py
@@ -96,7 +96,7 @@ def daily_summary(db_session=None):
                 },
             }
         )
-        for incident in active_incidents:
+        for idx, incident in enumerate(active_incidents):
             if incident.visibility == Visibility.open:
                 try:
                     blocks.append(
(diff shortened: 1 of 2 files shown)


forestmonster:
As a user, I would want to get involved at any point without having to scroll back. So daily buttons are better, but buttons on every message (other than `closed`) are best. If you tell me "hey, something happened on this incident, we moved it from stable back to active," then, give me a button, don't you think?

kevgliss:
@forestmonster the functionality you mention was included in #187

src/dispatch/incident/scheduled.py
```diff
@@ -110,6 +111,12 @@ def daily_summary(db_session=None):
                                     f"*Incident Commander*: <{incident.commander.weblink}|{incident.commander.name}>"
                                 ),
                             },
+                            "block_id": f"{ConversationButtonActions.invite_user}-{idx}",
+                            "accessory": {
+                                "type": "button",
+                                "text": {"type": "plain_text", "text": "Get Involved"},
+                                "value": f"{incident.id}",
```
mvilanova, line 118: Does this need to be an f-string?

kevgliss: I wasn't confident if we passed an int, slack would do the right thing. So I just forced it to be a string to be safe.

mvilanova: Ok

PR #193, opened 2020-04-02
Feature/get involved incident notifications

kevgliss:
In channel:

In notification channel:


forestmonster:
@mvilanova and I chatted about #187 versus #191, thanks. I personally like the more compressed format you are going for here, and how these only show up in the notifications channel. :+1: on my end

PR #198, opened 2020-04-02
Exposes incident_priorities in the UI

kevgliss:
This PR effectively moves us off of the legacy incident priority enum in favor of using the model exclusively. 

With this change we add metadata to the priority to determine:

- If a commander should be paged on incident creation
- How often a status report reminder should be sent (in hours)
- The order in which to list the priorities

It also adds the UI elements needed to create/update a priority from the Admin UI.


src/dispatch/incident_priority/service.py
```diff
@@ -19,6 +19,11 @@ def get_by_name(*, db_session, name: str) -> Optional[IncidentPriority]:
     return db_session.query(IncidentPriority).filter(IncidentPriority.name == name).one_or_none()
 
 
+def get_by_slug(*, db_session, slug: str) -> Optional[IncidentPriority]:
+    """Returns an incident type based on the given type slug."""
```
mvilanova, line 23:
```suggestion
    """Returns an incident priority based on the given type slug."""
```

PR #199, opened 2020-04-03
Fix restore and sample data

diogomatsubara:
This patch fixes the sample data ownership to the default user used by dispatch `postgres` and fix the restore command to use the DATABASE_NAME env var.


kevgliss:
Awesome thanks!

PR #203, opened 2020-04-03
Bugfix/participant add after stable

kevgliss:
Prevents new participants from automatically being added if the incident is not active.


src/dispatch/incident/flows.py
```diff
@@ -795,17 +795,19 @@ def incident_update_flow(
...
-
-    # we add the individuals as incident participants
-    for individual in individual_participants:
-        incident_add_or_reactivate_participant_flow(
-            individual.email, incident.id, db_session=db_session
+    # lets not attempt to add new participants for non-active incidents (it's confusing)
+    if incident.status == IncidentStatus.active:
+        # get the incident participants based on incident type and priority
+        individual_participants, team_participants = get_incident_participants(
+            db_session, incident.incident_type, incident.incident_priority, incident.description
```
mvilanova, line 802: Only need to pass `incident` and `db_session` after #204.
```suggestion
        individual_participants, team_participants = get_incident_participants(incident, db_session)
```

PR #204, opened 2020-04-03
Logs Dispatch and plugin-related events in incident flows

mvilanova:


src/dispatch/incident/flows.py
```diff
@@ -351,41 +477,40 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
-        incident.title,
-        incident.incident_type.name,
-        incident.incident_priority.name,
-        incident.commander.email,
-        incident.reporter.email,
-        incident.visibility,
+    event_service.log(
+        db_session=db_session,
+        source="Dispatch App",
+        description="Incident participants added to incident",
```
kevgliss, line 491: Is this an event? Wouldn't we more likely to know when person X was added?

src/dispatch/incident/flows.py
```diff
@@ -547,11 +688,21 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
             participant.individual.email, incident.id, db_session
         )
 
-    log.debug("Sent incident welcome and announcement notifications.")
+    event_service.log(
+        db_session=db_session,
+        source="Dispatch App",
+        description="Participants announced and welcome messages sent",
```
kevgliss, line 694: Same as above, not sure these more general messages are actual events.

kevgliss:
This looks good. I'd be careful to keep out some of the more general messaging out of the events and focus on things you'd really want to know, otherwise we will just have to figure out how to filter them after the fact. 

I'd also say that we should probably log all of the events (you could add a log statement in the event_create). Because even though we are now putting them into the database it's still useful to see those events in the log itself.

src/dispatch/incident/flows.py
mvilanova: We do it in the add participant flow.

src/dispatch/incident/flows.py
mvilanova: Yeah, I was on the fence with this one.

mvilanova:
> I'd also say that we should probably log all of the events (you could add a log statement in the event_create). Because even though we are now putting them into the database it's still useful to see those events in the log itself.
Good idea.

PR #205, opened 2020-04-03
Feature/google drive to folders

kevgliss:
This PR moves us off google shared drives in favor of using folders instead. 

This has a couple of advantages:

- A user's drive view is no longer cluttered with Google Shared drives.
- We don't have to move or delete any files when an incident is closed. Which could take a long time if there are many files.
- Google team drive tend to take longer to create than folders.

Additionally, we don't have to control the individual sharing permissions of every drive, instead of letting the top-level folder or top-level shared drive control general sharing permissions. 

This PR also adds two envvars and deprecates two others:

Adds:
`INCIDENT_STORAGE_FOLDER_ID`: Top level folder where all incident data is stored. Note: viewing actual incident data is still on a per-sub folder basis.
`INCIDENT_STORAGE_OPEN_ON_CLOSE`: After an incident is closed we as an organization try to be transparent and allow others to view incident data. This is not desirable in all organizations. The default is `True` or to open after close.

Deprecates:
`INCIDENT_STORAGE_ARCHIVAL_FOLDER_ID`: Is no longer needed, due to the fact that we won't me moving files for archival.
`INCIDENT_STORAGE_RESTRICTED`: Is no longer needed because we use whatever top-level permissions are for the folder or shared drive (instead of setting it for each not team drive).


docs/configuration/app.md
```diff
@@ -169,13 +169,13 @@ If you wish to disabled authentication set `DISPATCH_AUTHENTICATION_PROVIDER=""`
...
+#### `INCIDENT_STORAGE_FOLDER_ID`
 
-> Controls the folder where to archive incident information.
+> Top level folder where all incident data is stored. Note: viewing actual incident data is still on a per-sub folder basis.
 
-#### `INCIDENT_STORAGE_RESTRICTED` \[default: 'True'\]
+#### `INCIDENT_STORAGE_OPEN_ON_CLOSE` \[default: 'true'\]
 
-> Controls whether a set of restrictions and capabilities to prevent content sharing need to be applied.
+> After an incident is closed, Netflix as an organization, tries to be transparent and allow others within the organization to view incident data. This is may not desirable in all organizations. This open controls whether to open up incident data on incident close.
```
mvilanova, line 178:
```suggestion
> After an incident is closed, Netflix as an organization, tries to be transparent and allow others within the organization to view incident data. This is may not desirable in all organizations. This controls whether to open up incident data on incident close.
```

src/dispatch/incident/flows.py
```diff
@@ -291,44 +291,11 @@ def create_incident_storage(
 ):
     """Create an external file store for incident storage."""
     p = plugins.get(INCIDENT_PLUGIN_STORAGE_SLUG)
-    storage = p.create(incident.name, participant_group_emails)
+    storage = p.create_file(INCIDENT_STORAGE_FOLDER_ID, incident.title, participant_group_emails)
```
mvilanova, line 294:
```suggestion
    storage = p.create_file(INCIDENT_STORAGE_FOLDER_ID, incident.name, participant_group_emails)
```

mvilanova: Or 
```suggestion
    storage = p.create_file(INCIDENT_STORAGE_FOLDER_ID, f"{inicident.name} - {incident.title}", participant_group_emails)
```

src/dispatch/incident/flows.py
```diff
@@ -865,8 +832,13 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
     )
 
     if incident.visibility == Visibility.open:
-        # we archive the artifacts in the storage
-        archive_incident_artifacts(incident, db_session)
+
+        # open file to domain on closure
```
mvilanova, line 836:
```suggestion
        # open folder to domain on closure
```

src/dispatch/incident/flows.py
```diff
@@ -865,8 +832,13 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
...
+        if INCIDENT_STORAGE_OPEN_ON_CLOSE:
+            # we archive the artifacts in the storage
+            storage_plugin = plugins.get(INCIDENT_PLUGIN_STORAGE_SLUG)
+            storage_plugin.archive(folder_id=incident.storage.resource_id,)
+            log.debug("We have archived the incident artifacts.")
```
mvilanova, line 841: I'd rather have this in a function like before.

src/dispatch/plugins/dispatch_google/drive/drive.py
```diff
@@ -208,24 +208,36 @@ def restrict_team_drive(client: Any, team_drive_id: str):
             "canShare": False,
         },
     }
-    return make_call(client.drives(), "update", driveId=team_drive_id, body=body)
+    return make_call(client.drives(), "update", driveId=id, body=body)
```
mvilanova, line 211:
```suggestion
    return make_call(client.drives(), "update", driveId=folder_id, body=body)
```

src/dispatch/incident/flows.py
kevgliss: I don't think there is enough logic to justify a function.

kevgliss:
This PR is blocked until the sharing of drive folders reaches GA.

src/dispatch/incident/flows.py
mvilanova: It would improve flow code readability. See functions `delete_participant_groups()` and `delete_conference()` below.

PR #208, opened 2020-04-06
Logs events to console

mvilanova:


src/dispatch/incident/flows.py
```diff
@@ -910,18 +910,18 @@ def incident_update_flow(
 
     log.debug(f"Updated the external ticket {incident.ticket.resource_id}.")
 
+    # get the incident participants based on incident type and priority
+    individual_participants, team_participants = get_incident_participants(incident, db_session)
+
```
mvilanova, line 915: @kevgliss moved this here as `team_participants` was used before assigned if incident status != active.

PR #212, opened 2020-04-07
Creates events from reacted messages in Slack

mvilanova:
This PR adds the ability to create incident events from reacted messages in Slack. The `stopwatch` reaction is used by default. The event description will contain the text of the reacted message.

It also introduces a new env var called `SLACK_API_USER_TOKEN`, which stores the Slack User OAuth Access Token required to call the `conversations.history` Slack API method.


src/dispatch/event/service.py
```diff
@@ -113,14 +113,17 @@ def log(
     description: str,
     incident_id: int,
     individual_id: int = None,
-    started_at: datetime = datetime.datetime.now(),
+    started_at: datetime = None,
```
kevgliss, line 116: Can't this be `started_at: Optional[datetime] = datetime.datetime.now()`?

mvilanova: Would that behave differently from what I had before? I decided to move it to the body of the function, because I noticed that the value was the same for all events in the database. Moving it there solved the problem.

PR #216, opened 2020-04-08
Feature/incident table filter

kevgliss:
Adds basic filtering for the incident table. If this approach works well, we can likely expand to the other tables and add filtering there as well. 

This approach should work with filtering the `command` and `reporter` fields as well as soon as an underlying library supports hybrid properties. See: juliotrigo/sqlalchemy-filters#32

For other filters (e.g. visibility, cost, reported, etc.) these can be incrementally added as needed.

### DIFF
--- a/src/dispatch/static/dispatch/src/incident/Table.vue
+++ b/src/dispatch/static/dispatch/src/incident/Table.vue
@@ -4,7 +4,47 @@
     <!--<delete-dialog />-->
     <div class="headline">Incidents</div>
     <v-spacer />
-    <v-btn color="primary" dark class="mb-2" @click="createEditShow()">New</v-btn>
+    <v-btn color="primary" dark @click="createEditShow()">New</v-btn>
+    <v-menu v-model="filterMenu" :close-on-content-click="false" :nudge-width="300" offset-y>
+      <template v-slot:activator="{ on }">
+        <v-btn class="ml-2" color="secondary" dark v-on="on">Filters</v-btn>
+      </template>
+
+      <v-card>
+        <v-list>
+          <v-list-item>
+            <v-list-item-content>
+              <individual-combobox v-model="filters.commanders" label="Commanders"></individual-combobox>
+            </v-list-item-content>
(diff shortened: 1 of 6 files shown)


src/dispatch/static/dispatch/src/incident/store.js
```diff
@@ -54,13 +62,37 @@ const state = {
...
+    return state.table.options
+  }
 }
 
 const actions = {
   getAll: debounce(({ commit, state }) => {
     commit("SET_TABLE_LOADING", true)
-    return IncidentApi.getAll(state.table.options).then(response => {
+
+    let tableOptions = Object.assign({}, state.table.options)
```
kevgliss, line 76: This logic can likely be abstracted away when we implement additional tables.

src/dispatch/static/dispatch/src/incident/store.js
```diff
@@ -54,13 +62,37 @@ const state = {
...
+          tableOptions.fields.push(key + ".id")
+          tableOptions.values.push(value.id)
+        } else {
+          tableOptions.fields.push(key)
+          tableOptions.values.push(value)
+        }
+        tableOptions.ops.push("==")
+      })
+    })
+    return IncidentApi.getAll(tableOptions).then(response => {
```
kevgliss, line 95: This logic can be abstracted once we add filtering to more tables.

src/dispatch/static/dispatch/src/incident/Table.vue
```diff
@@ -4,7 +4,48 @@
...
+            <v-list-item-content>
+              <individual-combobox v-model="commander" label="Commanders"></individual-combobox>
+            </v-list-item-content>
+          </v-list-item>
+          <v-list-item>
+            <v-list-item-content>
+              <individual-combobox v-model="reporter" label="Reporters"></individual-combobox>
+            </v-list-item-content>
+          </v-list-item>
+          -->
```
kevgliss, line 29: We can add these back once we have hybrid property support.

PR #217, opened 2020-04-10
Basic Authentication 

stefanm8:
This PR refers to #95 

Is only the backend part


src/dispatch/auth/models.py
```diff
@@ -0,0 +1,24 @@
...
+from dispatch.models import TimeStampMixin, DispatchBase
+from pydantic import validator
+
+
+class DispatchUser(Base, TimeStampMixin):
+    email = Column(String, primary_key=True)
+
+
+class UserLoginForm(DispatchBase):
+    email: str
```
kevgliss, line 14: Where do you authenticate the user? I don't see where you are storing a password (that you would then exchange for a JWT token).

stefanm8: kevgliss, I omitted storing password, I will make the modifications necessary.

stefanm8:
I've included in this PR the frontend as well for this, currently there is no persistence of session. The way I see it, on the backend part expiry time must be added to token, and on the frontend save that token.

src/dispatch/static/dispatch/src/auth/store.js
```diff
@@ -14,6 +17,26 @@ const actions = {
...
+      })
+  },
+  register({ commit }) {
+    LoginApi.register(state.creds.email, state.creds.password)
+      .then(function(res) {
+        commit("SET_USER_LOGIN", res.data.token)
+        router.push("/dashboard")
+      })
+      .catch(function() {
+        commit("app/SET_SNACKBAR", { text: "Invalid credentials", color: "red" }, { root: true })
```
stefanm8, line 37: This was supposed to bring up a snack, it doesn't and. not sure why this happens.

.vscode/settings.json
```diff
@@ -27,5 +27,6 @@
     "editor.codeActionsOnSave": {
       "source.organizeImports": false
     }
-  }
+  },
+  "python.pythonPath": ".venv/bin/python"
```
kevgliss, line 31: This should be in your personal settings and not in the project settings.

src/dispatch/static/dispatch/src/api.js
```diff
@@ -20,4 +20,18 @@ instance.interceptors.request.use(
   }
 )
 
+instance.interceptors.response.use(
```
kevgliss, line 23: Today, we check for auth on route load (instead of 401) and then have a timer telling a user to refresh before their token expires. 
See here:
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/static/dispatch/src/router/index.js#L137
and
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/static/dispatch/src/auth/store.js#L31

I'm not sure how that fits in with the login feature? Perhaps we set the same expiration but send users to /login when basic authentication is enabled?

We will also need a way to tell the frontend to use PKCE or basic authentication via the `DISPATCH_AUTHENTICATION_PROVIDER_SLUG` here:

https://github.com/Netflix/dispatch/blob/develop/src/dispatch/static/dispatch/src/router/index.js#L31

stefanm8: I moved the auth check, in the router/index.js although I kept the check on the responses so I can effectively logout the user, this should not interact with any other auth provider, unless there are 401 returned (which is not a bad intent to logout user if 401 is received)

This should not interact with the current login feature. Exactly when no auth provider slug is provided basic auth will be used and this handler will be accessed "loginBasic" from router/index.js.

I decided to let it empty and modify value of `VUE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG` to set a default value in case no slug is present

src/dispatch/static/dispatch/src/auth/store.js
kevgliss: That's okay, we can take a look.

src/dispatch/static/dispatch/src/router/index.js
```diff
@@ -138,7 +152,7 @@ router.beforeEach((to, from, next) => {
     if (authProviderSlug === "dispatch-auth-provider-pkce") {
       loginwithPKCE(to, from, next)
     } else {
-      next()
```
kevgliss, line 141: I think we need to provide a `dispatch-auth-provider-basic` slug. There are users who don't wish to enabled any kind of auth. That said I'm okay making `dispatch-auth-provider-basic` the default here: https://github.com/Netflix/dispatch/blob/develop/src/dispatch/config.py#L75

stefanm8: I see what you are saying, I added to the config "dispatch-auth-provider-basic". I did not know this app is usable without any authentication.

kevgliss:
Awesome, this looks good. I'll merge this in and test it out with some other changes. I wanted to get a release out before merging this in, so thanks for hanging in there and making changes.

PR #218, opened 2020-04-11
Feature/incident details

kevgliss:
Refactors how incidents are view in the UI and adds additional details to incidents via Admin UI.


src/dispatch/participant/models.py
```diff
@@ -56,6 +55,8 @@ class ParticipantUpdate(ParticipantBase):
 
 class ParticipantRead(ParticipantBase):
     id: int
+    participant_role: Optional[Any]
```
mvilanova, line 58: Shouldn't this be of type `ParticipantRole` role?

PR #220, opened 2020-04-11
Adding service descriptions

kevgliss:
Adds a description field for services.


src/dispatch/service/models.py
```diff
@@ -51,6 +51,7 @@ class Service(TimeStampMixin, Base):
     is_active = Column(Boolean, default=True)
     name = Column(String)
     type = Column(String, default="pagerduty-oncall")
```
mvilanova, line 53: Should we define the slug for the oncall plugin in the config and use it as default instead of a hardcoded string?

kevgliss: Probably, but we can do that in a separate PR.

PR #225, opened 2020-04-15
Improves incident cost calculation

mvilanova:
The model now uses the time participants have spent on each incident role and an engagement multiplier based on the role to adjust time spent.


src/dispatch/incident/service.py
```diff
@@ -263,49 +263,80 @@ def delete(*, db_session, incident_id: int):
...
+
+    return engagement_mappings.get(participant_role, [])
+
+
+def calculate_cost(incident_id: int, db_session: SessionLocal, incident_review=True):
+    """Calculates the cost of a given incident."""
     incident = get(db_session=db_session, incident_id=incident_id)
 
-    participants_active_hours = 0
+    participants_total_response_time_seconds = 0.0
```
kevgliss, line 283: Does this need to be a float/decimal? Are you capturing milliseconds?

src/dispatch/incident/service.py
```diff
@@ -263,49 +263,80 @@ def delete(*, db_session, incident_id: int):
...
+            # we calculate the number of hours the participant has spent in the incident role
+            participant_role_time_hours = participant_role_time.total_seconds() / SECONDS_IN_HOUR
 
-    # we calculate the number of hours spent in incident review related activities
+            # we make the assumption that participants only spend 8 hours a day working on the incident,
+            # if the incident goes past 24hrs
+            # TODO(mvilanova): adjust based on incident priority
+            if participant_role_time_hours > HOURS_IN_DAY:
+                days, hours = divmod(participant_role_time_hours, HOURS_IN_DAY)
+                participant_role_time_hours = math.ceil(((days * HOURS_IN_DAY) / 3) + hours)
```
kevgliss, line 310: We could cast `int(math.ceil())` if we don't want to deal with decimals.

src/dispatch/incident/service.py
```diff
@@ -263,49 +263,80 @@ def delete(*, db_session, incident_id: int):
...
+    """Returns an engagement multiplier for a given incident role."""
+    engagement_mappings = {
+        ParticipantRoleType.incident_commander: 1,
+        ParticipantRoleType.scribe: 0.75,
+        ParticipantRoleType.liaison: 0.75,
+        ParticipantRoleType.participant: 0.5,
+        ParticipantRoleType.reporter: 0.5,
+    }
+
+    return engagement_mappings.get(participant_role, [])
```
kevgliss, line 276: Should this default to a list? I'm not sure having a default here makes sense.

src/dispatch/participant/models.py
```diff
@@ -46,7 +48,10 @@ class ParticipantBase(DispatchBase):
 
 
 class ParticipantCreate(ParticipantBase):
-    participant_role: Optional[List[ParticipantRoleCreate]] = []
+    participant_roles: Optional[List[ParticipantRoleCreate]] = []
```
kevgliss, line 51: This will affect some frontend UI elements, also we should modify `ParticipantRead` as well.

src/dispatch/incident/service.py
mvilanova: No, not trying to capture milliseconds. The result of adjusting the time with the engagement multiplier could result in a `float`. I'll make it an `int` and cast the result to `int` as well.

src/dispatch/incident/service.py
mvilanova: Good catch. No, I'll remove the default.

PR #223, opened 2020-04-15
Initial work on decoupling plugins from creation flow

kevgliss:
This PR includes the initial foundations for allowing us to selectively disable plugins. Starts to address the concerns uncovered in #55 

Includes:
- New table for `Plugins`
- UI and API for viewing an editing plugin state
- Selectively disables the 'ticket' plugin and uses the dispatch admin UI instead
- Links a plugins `.enabled()` function with what is stored in the database
- Adds newly installed plugins to the database and enables them

 Not Included in this PR but will be in future PRs:
- Any ability to store the _configuration_ of the plugins or a way to interactively test whether they are working.

PR #233, opened 2020-04-17
Feature/incident auto tagger

kevgliss:
This is initial work required for a scheduled auto tagger. It searching through incident resources (google docs at the moment) and analyzes the text, associating known tags with an incident.

I've also added a `discoverable` flag to the tag itself, if the tag should never be automatically associated (for example if it's a common English word). 

The included `nlp.py` is a more generalized version of what is available via the route `service.py` these will be refactored in a separate PR.


src/dispatch/incident/scheduled.py
```diff
@@ -32,6 +43,47 @@
...
+    tags = tag_service.get_all(db_session=db_session).all()
+    log.debug(f"Fetched {len(tags)} tags from database.")
+
+    tag_strings = [t.name.lower() for t in tags if t.discoverable]
+    phrases = build_term_vocab(tag_strings)
+    matcher = build_phrase_matcher("dispatch-tag", phrases)
+
+    p = plugins.get(
+        INCIDENT_PLUGIN_STORAGE_SLUG
+    )  # this may need to be refactored if we support multiple document types
```
mvilanova, line 59: Just adding a note that we will need to refactor this once https://github.com/Netflix/dispatch/pull/223 lands.

src/dispatch/incident/scheduled.py
```diff
@@ -32,6 +43,47 @@
...
+        matched_tags = (
+            db_session.query(Tag)
+            .filter(func.upper(Tag.name).in_([func.upper(t) for t in extracted_tags]))
+            .all()
+        )
+
+        incident.tags.extend(matched_tags)
+        db_session.commit()
+
+        print(f"Associating tags with incident. Incident: {incident.name}, Tags: {extracted_tags}")
```
mvilanova, line 84: Remove or replace it with `log.debug()`.

src/dispatch/nlp.py
```diff
@@ -0,0 +1,52 @@
+import logging
```
mvilanova, line 1: Should we create a directory for nlp and rename this module to service?

kevgliss: Not yet, I don't think. A dir with one file seems like overkill, i.e I don't see nlp ever needing view or models.

src/dispatch/incident/scheduled.py
kevgliss: Yes eventually, but #223  itself won't require a refactor of this code. We will move each plugin type one by one and refactor all the slug references.

PR #241, opened 2020-04-21
Making some minor fixes around basic auth

kevgliss:
This PR tweaks some things about the basic auth implementation. 

- Modifies the login and register forms to look more like the rest of the application
- Ensures that we don't advertise /login or /register API endpoints if we don't have `dispatch-auth-provider-basic` enabled.
- Adds missing database migration for `DispatchUser`

Other considerations, should we add another config value to enable/disable registration? Given the nature of the app, I wouldn't expect most folks to allow registration without approval. 

cc @stefanm8


stefanm8:
Hi, looks really good, gg! 

Disabling registration does make sense, as most probably it will be closed when this runs in production. This can be achieved with minimal effort on the backend side.

src/dispatch/static/dispatch/src/router/index.js
```diff
@@ -134,12 +134,10 @@ function loginBasic(to, from, next) {
   let token = localStorage.getItem("token")
   if (token) {
     store.dispatch("account/loginWithToken", token)
-    return
```
stefanm8, line 137: In case token exists and user is authorized, I think he should continue navigating where he wanted. Eliminating the return statement will cause the user to be redirected to "login" regardless of the token existence.

src/dispatch/alembic/versions/62465f508f69_.py
```diff
@@ -0,0 +1,34 @@
+"""empty message
```
mvilanova, line 1: Add description.

docs/configuration/app.md
```diff
@@ -57,14 +57,32 @@ In general, do not include any quotation marks when adding configuration values.
 
 ### Authentication
 
-#### `DISPATCH_AUTHENTICATION_PROVIDER` \['default': dispatch-auth-provider-pkce\]
+#### `DISPATCH_AUTHENTICATION_PROVIDER` \['default': dispatch-auth-provider-basic\]
 
 > Used by Dispatch to determine which authentication provider to use, by default Dispatch ships with a PKCE authentication provider.
 
 {% hint style="info" %}
 If you wish to disabled authentication set `DISPATCH_AUTHENTICATION_PROVIDER=""`
```
mvilanova, line 65: Is this still valid? I don't see `DISPATCH_AUTHENTICATION_PROVIDER` being used anywhere in the code.

src/dispatch/alembic/versions/62465f508f69_.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision = '1221a4d60f03'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('dispatch_user',
+    sa.Column('email', sa.String(), nullable=False),
+    sa.Column('password', sa.Binary(), nullable=True),
```
mvilanova, line 23: Should we allow people to create accounts without password?

docs/configuration/app.md
```diff
@@ -57,14 +57,32 @@ In general, do not include any quotation marks when adding configuration values.
 
 ### Authentication
 
-#### `DISPATCH_AUTHENTICATION_PROVIDER` \['default': dispatch-auth-provider-pkce\]
+#### `DISPATCH_AUTHENTICATION_PROVIDER` \['default': dispatch-auth-provider-basic\]
```
mvilanova, line 60:
```suggestion
#### `DISPATCH_AUTHENTICATION_PROVIDER_SLUG` \['default': dispatch-auth-provider-basic\]
```

docs/configuration/app.md
kevgliss: Ah, should have `slug` added to the end. Will add.

src/dispatch/alembic/versions/62465f508f69_.py
kevgliss: Probably not.

src/dispatch/static/dispatch/src/router/index.js
kevgliss: That makes sense, although I'd rather we explicitly send them to their original destination. I was also getting a redirect routing loop when I was navigating to `/login` initially (and it kept sending me back to `/login` since that was my original destination (and I already had a token). 

Let me try to make it a bit more explicit.

stefanm8: Yep, the loop comes from calling multiple times "next()" at L140 and L142

PR #259, opened 2020-04-23
Adds flags to plugins allowing them to specify if they are required

kevgliss:
Adds two new columns to the plugins table:

- required - Allows plugins to specify if they are required.
- multiple - Allows plugins to specify if they can have multiple exist together. 

Both options will be usually set by the base plugin class of a given type. 

This PR then uses that new information to validate when and how plugins can be safely enabled and disabled.


src/dispatch/plugin/service.py
```diff
@@ -51,15 +62,19 @@ def update(*, db_session, plugin: Plugin, plugin_in: PluginUpdate) -> Plugin:
...
+        if not plugin.multiple:
+            enabled_plugins = get_enabled_by_type(db_session=db_session, plugin_type=plugin.type)
+            if enabled_plugins:
+                enabled_plugins[0].enabled = False
+                db_session.add(enabled_plugins[0])
+
+    if not plugin_in.enabled:
+        if plugin.required:
+            raise InvalidConfiguration(
+                "Cannot disable plugin, it is required and no other plugins of it's type are enabled."
```
mvilanova, line 76:
```suggestion
                f"Cannot disable plugin: {plugin.title}. It is required and no other plugins of type {plugin.type} are enabled."
```

src/dispatch/plugin/service.py
```diff
@@ -28,6 +29,16 @@ def get_by_type(*, db_session, plugin_type: str) -> List[Optional[Plugin]]:
     return db_session.query(Plugin).filter(Plugin.type == plugin_type).all()
 
 
+def get_enabled_by_type(*, db_session, plugin_type: str) -> List[Optional[Plugin]]:
+    """Fetches all plugins for a given type."""
```
mvilanova, line 33:
```suggestion
    """Fetches all enabled plugins for a given type."""
```

PR #262, opened 2020-04-24
Adds more timeline events

mvilanova:
Screen Shot 2020-04-24 at 10 34 35 AM
Screen Shot 2020-04-24 at 10 40 44 AM
Screen Shot 2020-04-24 at 10 43 37 AM


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -155,7 +155,7 @@ def handle_reaction_added_event(
         event_service.log(
             db_session=db_session,
             source=convo_plugin.title,
-            description=message_text,
+            description=f'"{message_text}," said {individual.name}',
```
kevgliss, line 158: nit: This feels a bit awkward, perhaps we could just do "Author: " instead of "Said"?

src/dispatch/status_report/flows.py
```diff
@@ -46,9 +46,10 @@ def save_status_report(
 
     event_service.log(
         db_session=db_session,
-        source="Dispatch Core App",
-        description=f"New status report created by {participant.individual.name}",
+        source="Incident Participant",
+        description=f"{participant.individual.name} created a new status report. Conditions: {conditions}; Actions: {actions}; Needs: {needs}",
```
kevgliss, line 50: This might eventually stress the display of the timeline. I wonder if we could add "metadata" to the event table and store any structured data for the event there?

mvilanova: Agree. I'll find another way to display large piece of text and add the column to the model in a separate PR.

src/dispatch/plugins/dispatch_slack/views.py
mvilanova: Fine :)

PR #266, opened 2020-04-24
Initial work adding RBAC controls

kevgliss:
This is initial work on adding rbac controls to Dispatch. Including the creation of a `DispatchUser` object regardless of the way the user is authenticated. We then add a default role to this user object that can then be used for permissioning. 

We will initially have three different roles:
- user 
- poweruser
- admin

There is currently no provision of anonymous access to any of Dispatch's resources. 

Adds new UI elements as well:


src/dispatch/auth/views.py
```diff
@@ -1,47 +1,40 @@
...
 def login_user(
-    form: UserLoginForm, db_session: Session = Depends(get_db),
+    user_in: UserLogin, db_session: Session = Depends(get_db),
 ):
-    user = fetch_user(db_session, form.email)
-    if user and check_password(form.password, user.password):
-        return {"token": gen_token({"email": user.email})}
-    return {"error": "Invalid username or password"}
+    user = get(db_session=db_session, email=user_in.email)
+    if user and user.check_password:
```
mvilanova, line 19: Shouldn't we pass `user_in.password` to `user.check_password`?

src/dispatch/auth/models.py
```diff
@@ -1,32 +1,107 @@
...
 class DispatchUser(Base, TimeStampMixin):
     email = Column(String, primary_key=True)
     password = Column(Binary, nullable=False)
+    role = Column(String, nullable=False, default=UserRoles.user)
+
+    def check_password(self, password):
+        return bcrypt.checkpw(password.encode("utf-8"), self.password)
 
+    def token(self):
+        now = datetime.now()
```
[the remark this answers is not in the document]
kevgliss: Yup.

src/dispatch/alembic/versions/9a3478cbe76c_.py
```diff
@@ -0,0 +1,48 @@
...
+    for r in results:
+        if len(r) == 5:
+            session.add(DispatchUser(email=r[0], password=r[1], role=r[4]))
+        else:
+            session.add(DispatchUser(email=r[0], password=r[1]))
+    session.commit()
+
+
+def downgrade():
+    pass
```
mvilanova, line 48: Should this be `op.drop_table("dispatch_user")`?

src/dispatch/auth/models.py
```diff
@@ -1,32 +1,121 @@
...
         return v
 
 
+class UserRegister(UserLogin):
+    password: Optional[str]
+    role: UserRoles = UserRoles.user
+
+    @validator("password", pre=True, always=True)
+    def password_required(cls, v):
+        # we generate a password for those that don't have one
```
mvilanova, line 96: Do we present the password to the user in the Web UI in this case?

src/dispatch/auth/service.py
```diff
@@ -29,36 +29,64 @@
...
-    log.debug("No authentication provider. Default one will be used")
-    return DISPATCH_AUTHENTICATION_DEFAULT_USER
 
+def get_by_email(*, db_session, email: str) -> Optional[DispatchUser]:
+    """Returns an user object based on user email."""
+    return db_session.query(DispatchUser).filter(DispatchUser.email == email).one_or_none()
+
+
+def create(*, db_session, user_in: UserRegister) -> DispatchUser:
+    """Creates a new dispatch user."""
```
mvilanova, line 43:
```suggestion
    """Creates a new Dispatch user."""
```

src/dispatch/incident/views.py
```diff
@@ -47,13 +61,25 @@ def get_incidents(
...
     Retrieve details about a specific incident.
     """
     incident = get(db_session=db_session, incident_id=incident_id)
     if not incident:
         raise HTTPException(status_code=404, detail="The requested incident does not exist.")
+
+    # we want to provide additional protections around restricted incidents
+    if not has_permission(current_user.principals, "view", IncidentRead.__acl__):
+        raise HTTPException(
+            status_code=401, detail="You do no have permission to view this incident."
```
mvilanova, line 80:
```suggestion
            status_code=401, detail="You do no have permissions to view this incident."
```

src/dispatch/static/dispatch/src/auth/editSheet.vue
```diff
@@ -0,0 +1,69 @@
+<template>
```
mvilanova, line 1: Do we prevent non-admins to change users' roles?

src/dispatch/auth/models.py
kevgliss: No, this is typically for the SSO use case, I felt it was better to generate a random password instead of leaving the field empty.

src/dispatch/static/dispatch/src/auth/editSheet.vue
kevgliss: Not yet, will land that in another PR.

src/dispatch/alembic/versions/9a3478cbe76c_.py
kevgliss: Technically yes? But in practice, I don't think that's what we should do. I tried to do a clean upgrade but migrating from one primary key (email) to another (id) was pretty difficult to do with alembic, so since these are all intermediate upgrades I think it will be okay. When do releases I think we should attempt to consolidate revisions if possible?

PR #267, opened 2020-04-25
Adding Conference tests.

forestmonster:
Test the `create` and `get`

### DIFF
--- a/src/dispatch/conference/service.py
+++ b/src/dispatch/conference/service.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, List
 
 from .models import Conference, ConferenceCreate
 
@@ -11,10 +11,9 @@ def get_by_resource_id(*, db_session, resource_id: str) -> Optional[Conference]:
     return db_session.query(Conference).filter(Conference.resource_id == resource_id).one_or_none()
 
 
-def get_by_resource_type(*, db_session, resource_type: str) -> Optional[Conference]:
-    return (
-        db_session.query(Conference).filter(Conference.resource_type == resource_type).one_or_none()
-    )
+def get_by_resource_type(*, db_session, resource_type: str) -> List[Conference]:
+    """Return a List of all Conferences matching a given resource type."""
+    return db_session.query(Conference).filter(Conference.resource_type == resource_type).all()
 
 
 def get_by_conference_id(db_session, conference_id: str) -> Optional[Conference]:
--- a/src/dispatch/conference/service.py
+++ b/src/dispatch/conference/service.py
@@ -11,8 +11,9 @@ def get_by_resource_id(*, db_session, resource_id: str) -> Optional[Conference]:
     return db_session.query(Conference).filter(Conference.resource_id == resource_id).one_or_none()
 
 
-def get_by_resource_type(*, db_session, resource_type: str) -> List[Conference]:
-    """Return a List of all Conferences matching a given resource type."""
+def get_by_resource_type(*, db_session, resource_type: str) -> List[Optional[Conference]]:
+    """Return a list of all conferences matching a given resource type.
+       May return an empty list if no conferences are available."""
     return db_session.query(Conference).filter(Conference.resource_type == resource_type).all()
 
 
--- a/src/dispatch/conference/service.py
+++ b/src/dispatch/conference/service.py
@@ -23,11 +23,7 @@ def get_by_conference_id(db_session, conference_id: str) -> Optional[Conference]
 
 
 def get_by_incident_id(*, db_session, incident_id: str) -> Optional[Conference]:
-    return (
-        db_session.query(Conference)
-        .filter(Conference.incident_id == incident_id)
-        .one()
-    )
+    return db_session.query(Conference).filter(Conference.incident_id == incident_id).one()
 
 
 def get_all(*, db_session):
--- a/tests/conference/test_conference_service.py
+++ b/tests/conference/test_conference_service.py
@@ -56,9 +56,15 @@ def test_conference_get_by_resource_type(session, conference):
 
 
 def test_conference_get_by_conference_id(session, conference):
-    raise NotImplementedError
+    from dispatch.conference.service import get_by_conference_id
+
+    test_conference = get_by_conference_id(session, conference.conference_id)
+
+    assert test_conference.conference_id == conference.conference_id
+    assert test_conference.conference_challenge == conference.conference_challenge
 
 
+@pytest.mark.skip
 def test_conference_get_by_incident_id(session, conference):
     raise NotImplementedError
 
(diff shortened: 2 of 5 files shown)


tests/conftest.py
```diff
@@ -70,7 +71,7 @@ def pytest_runtest_makereport(item, call):
             parent._previousfailed = item
 
 
-@pytest.yield_fixture(scope="session")
+@pytest.fixture(scope="session")
```
mvilanova, line 74: Thanks for updating it.

tests/conference/test_conference_service.py
```diff
@@ -0,0 +1,31 @@
...
+
+    name = "fakename"
+    resource_id = "000000"
+    resource_type = "resourcetype"
+    weblink = "https://www.example.com"
+    conference_id = "12345"
+    conference_challenge = "a0v0a0v9a"
+
+    conference_in = ConferenceCreate(
+        name=name,
```
mvilanova, line 16: Remove. `name` does not exist in the conference model.

forestmonster: You're right, thanks :taco:

forestmonster: Removed in 08fe5c5.

tests/conference/test_conference_service.py
```diff
@@ -44,17 +44,24 @@ def test_conference_get_by_resource_id(session, conference):
...
 
 
-def test_conference_get_by_resource_type():
-    raise NotImplementedError
+def test_conference_get_by_resource_type(session, conference):
+    """The service method returns a List of Conferences that match a given resource type. We'll test
+    to ensure that the first and last items returned match the desired resource type."""
+    from dispatch.conference.service import get_by_resource_type
+
+    conferences = get_by_resource_type(db_session=session, resource_type=conference.resource_type)
```
mvilanova, line 52: How many objects of type Conference are you getting as a result? Looks like you're only passing one to `test_conference_get_by_resource_type()`.

mvilanova: You could create a fixture that returns a list of Conferences (example [here](https://github.com/Netflix/dispatch/blob/develop/tests/conftest.py#L381)), and then add each of them to two separate incidents.

tests/conference/test_conference_service.py
```diff
@@ -44,17 +44,24 @@ def test_conference_get_by_resource_id(session, conference):
...
+
+    assert conferences[0].resource_type == conference.resource_type
+    assert conferences[-1].resource_type == conference.resource_type
 
 
-def test_conference_get_by_conference_id():
+def test_conference_get_by_conference_id(session, conference):
```
mvilanova, line 58: Is this test passing? If not, you can decorate it with `@pytest.mark.skip()` to skip it until it's implemented.

tests/conference/test_conference_service.py
```diff
@@ -0,0 +1,67 @@
...
+        resource_id=resource_id,
+        resource_type=resource_type,
+        weblink=weblink,
+        conference_id=conference_id,
+        conference_challenge=conference_challenge,
+    )
+    conference = create(db_session=session, conference_in=conference_in)
+    assert conference
+    assert conference.resource_id == "000000"
+    assert resource_type == "resourcetype"
```
mvilanova, line 24: Did you forget to prefix `resource_type` with `conference.`?

src/dispatch/conference/service.py
```diff
@@ -11,10 +11,9 @@ def get_by_resource_id(*, db_session, resource_id: str) -> Optional[Conference]:
     return db_session.query(Conference).filter(Conference.resource_id == resource_id).one_or_none()
 
 
-def get_by_resource_type(*, db_session, resource_type: str) -> Optional[Conference]:
-    return (
-        db_session.query(Conference).filter(Conference.resource_type == resource_type).one_or_none()
-    )
+def get_by_resource_type(*, db_session, resource_type: str) -> List[Conference]:
```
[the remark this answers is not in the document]
forestmonster: Yes I guess if no Conferences are present in the database we should return an empty `List`.

tests/conference/test_conference_service.py
forestmonster: Since it only tests whether `get_by_resource_type` returns a list of conferences matching the desired `resource_type`, this particular test really only needs one value: the contents of the `resource_type` attribute attached to the incoming `conference` object. However the fixture is a suitable workaround for testing the `get_all()` method... thanks. :smile:

tests/conference/test_conference_service.py
forestmonster: Yes thanks, that would always succeed, so incorrect. Fixed in 6703f56.

src/dispatch/conference/service.py
```diff
@@ -11,10 +11,9 @@ def get_by_resource_id(*, db_session, resource_id: str) -> Optional[Conference]:
...
+    """Return a List of all Conferences matching a given resource type."""
```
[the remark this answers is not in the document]
forestmonster: Just differentiating a `Conference` from a "conference." But it's fine, it doesn't really matter for readability.

tests/conference/test_conference_service.py
```diff
@@ -0,0 +1,88 @@
...
+    assert test_conference.conference_challenge == conference.conference_challenge
+
+
+@pytest.mark.skip
+def test_conference_get_by_incident_id(session, conference):
+    from dispatch.conference.service import get_by_incident_id
+
+    """By the time we get here, several conferences have been added to the database. So this call to
+    get_by_incident_id will fail, as we assume only one() result there."""
+    test_conference = get_by_incident_id(db_session=session, incident_id=conference.incident_id)
```
forestmonster, line 73: @mvilanova Here's this problem that I have. When running this test in isolation (`pytest -sv tests/conference/test_conference_service.py::test_conference_get_by_incident_id`), it succeeds, because in the [invocation](https://github.com/Netflix/dispatch/pull/267/files#diff-6147207fd13defcb5b55d0abe8ddf963R68), one, and only one, `conference` has been created. But when running the series of tests, the previous `conference`s defined earlier in the file have already been created. So the call to `one()` in `get_by_incident_id` will fail. Should I just move the test earlier in the file? Seems like the easy solution, but hides the problem. Tests should be idempotent and should work when run in isolation. They shouldn't depend on the results of the tests immediately before or after them.

kevgliss: You should create a new incident for each invocation of `ConferenceFactory` you can either do this manually in your test or do it via `SubFactory`. There should never be more than one conference per incident. Also, I'm not sure where you are even associating an incident at all? Are you sure that the value isn't `None`? Quickly printing out the value of `conference` would indicate it hasn't been set.

```
  File "/home/kglisson/Workspace/dispatch/tests/conference/test_conference_service.py", line 73, in test_conference_get_by_incident_id
    assert conference.incident_id == "foo"
AssertionError: assert None == 'foo'
 +  where None = <dispatch.conference.models.Conference object at 0x7fe033779ba8>.incident_id
```

forestmonster: Used a `SubFactory` to fix this in d13420f

PR #270, opened 2020-04-28
Fixed empty enabled column in table 'plugin'

stefanm8:
Encountering pydantic errors when trying access plugin page as enabled column is empty. This PR assures that enabled column will always have a value.

### DIFF
--- a/src/dispatch/plugins/base/manager.py
+++ b/src/dispatch/plugins/base/manager.py
@@ -76,7 +76,7 @@ class PluginManager(InstanceManager):
                 required=cls.required,
                 multiple=cls.multiple,
                 description=cls.description,
-                enabled=getattr(cls, "enabled", False)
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/base/manager.py
```diff
@@ -76,6 +76,7 @@ def register(self, cls):
                 required=cls.required,
                 multiple=cls.multiple,
                 description=cls.description,
+                enabled=getattr(cls, "enabled", False)
```
kevgliss, line 79: I think this should be `enabled=cls.enabled` instead all plugins should have this attribute if they are using the plugin system. See: https://github.com/Netflix/dispatch/blob/develop/src/dispatch/plugins/base/v1.py#L62

mvilanova:
Merging as it's causing me issues as well.

PR #272, opened 2020-04-28
Ensures we always use the right time for renounced_at when we calculate the incident cost

mvilanova:


mvilanova:
I'll create a hotfix for it instead.

PR #275, opened 2020-04-29
fix typo in setup.py option

mlioo:
This was causing errors for setup tools, looks like it was a missed typo since the option is `zip_safe`.

`/usr/local/lib/python3.8/distutils/dist.py:274: UserWarning: Unknown distribution option: 'zip_save'
  warnings.warn(msg)
`


mvilanova:

PR #276, opened 2020-04-29
Fixes metric query, and reduces bundle size by allowing for treeshaki…

kevgliss:
…ng of lodash.

### DIFF
--- a/src/dispatch/database.py
+++ b/src/dispatch/database.py
@@ -1,4 +1,5 @@
 import re
+import logging
 from typing import Any, List
 from itertools import groupby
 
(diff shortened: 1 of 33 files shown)


src/dispatch/database.py
```diff
@@ -98,7 +101,22 @@ def create_filter_spec(model, fields, ops, values):
...
-        filter_spec.append({"or": list(g)})
+        # force 'and' for operations other than equality
+        filters = list(g)
+        force_and = False
+        for f in filters:
+            if ">" in f["op"]:
+                force_and = True
+
+            if "<" in f["op"]:
+                force_and = True
```
mvilanova, line 112:
```suggestion
            if ">" in f["op"] or "<" in f["op"]:
                force_and = True
```

PR #277, opened 2020-04-30
Bugfix router login redirect

stefanm8:
Currently if auth provider is set to dispatch-auth-provider-basic if the requestor is not authenticated, the requestor will remain hanging trying to access /login page.

e.g 
User is unauthorized gets redirect to "/login" from L143 after this the hook on routerBefore will be called again but this time the path.to is "/login" thus the router will actually have no resolve at all, and next() is never called.


mvilanova:
Thanks @stefanm8

kevgliss:
Ah, I had fixed this in #266 but you beat me to it @stefanm8, thanks!

PR #279, opened 2020-04-30
Allowing resources to be optionally created

kevgliss:


kevgliss:
Unless we hit bugs I'm pretty sure we will always have at least one of these resources.

PR #282, opened 2020-05-01
Timeline event details

mvilanova:
This PR adds a new column of type JSON to the event model called `details`, which can be used to store detailed information about an event in a structured format. It also adds a switch at the top left of the timeline tab in the Web UI, that when the user turns it on, event details are displayed (e.g. incident description, status report). The switch is off by default.

Screen Shot 2020-05-01 at 11 27 27 AM

Screen Shot 2020-05-01 at 12 03 53 PM


src/dispatch/static/dispatch/src/incident/TimelineTab.vue
```diff
@@ -19,6 +34,13 @@ import { mapFields } from "vuex-map-fields"
 
 export default {
   name: "IncidentTimelineTab",
+
+  data() {
+    return {
+      show: false
```
kevgliss, line 40: nit: would be clearer if this var was `showDetails`

src/dispatch/static/dispatch/src/incident/TimelineTab.vue
```diff
@@ -1,10 +1,25 @@
 <template>
   <v-container>
+    <v-switch v-model="show" label="Show details"></v-switch>
```
kevgliss, line 3: nit: this would look a bit cleaner if it was right aligned, I think you could just add <v-spacer />

PR #284, opened 2020-05-01
Refactors incident table filter

kevgliss:


src/dispatch/static/dispatch/src/incident/TableFilterDialog.vue
```diff
@@ -0,0 +1,93 @@
...
+            <v-list-item-content>
+              <individual-combobox v-model="commander" label="Commanders"></individual-combobox>
+            </v-list-item-content>
+          </v-list-item>
+          <v-list-item>
+            <v-list-item-content>
+              <individual-combobox v-model="reporter" label="Reporters"></individual-combobox>
+            </v-list-item-content>
+          </v-list-item>
+          -->
```
mvilanova, line 24: Are we planning to use this in the future? If not, we should remove it.

kevgliss: We will as soon as #124 is merged.

PR #288, opened 2020-05-03
Fixes help links for multi-messages and adds task resolution directions.

kevgliss:


src/dispatch/messaging.py
```diff
@@ -38,7 +38,7 @@ class MessageType(str, Enum):
 INCIDENT_TASK_REMINDER_DESCRIPTION = """
 You are assigned to the following incident tasks.
 This is a reminder that these tasks have *passed* their due date.
-Please review and update as appropriate.""".replace(
+Please review and update as appropriate. To supress this message please resolve the listed tasks.""".replace(
```
mvilanova, line 41:
```suggestion
Please review and update them as appropriate. Resolving them will stop the reminders.""".replace(
```

src/dispatch/plugins/dispatch_google/gmail/templates/base.html
```diff
@@ -115,9 +115,11 @@ <h1 style="color:#ffffff;display:block;font-family:Helvetica,Arial,sans-serif;fo
...
                                                     always reach us at
                                                     <a href="mailto:{{dispatch_help_email}}"
-                                                        target="_blank"><span
+                                                        target="_blank">
+                                                        <span
                                                             class="lG">{{dispatch_help_email}}
-                                                    or Slack channel
+                                                        </span>
+                                                    </a>or Slack channel
                                                     # {{dispatch_help_slack_channel}}.
```
mvilanova, line 123:
```suggestion
                                                    #{{dispatch_help_slack_channel}}.
```

kevgliss:
Yeah, we can delete `route/flows.py` looks like a copy-paste issue. We can add the title... but they can't be quite long, it may make the emails look terrible.

mvilanova:
Yeah, not the title, the name (e.g. SEC-1234)

kevgliss:
Actually, it looks like we already include the titles under the incident name:

https://github.com/Netflix/dispatch/blob/develop/src/dispatch/plugins/dispatch_google/gmail/templates/task_notification.html#L17

PR #299, opened 2020-05-08
Adds support for incident updates

mvilanova:
This PR adds support for writing and distributing incident updates. These updates are aimed at leadership and stakeholders rather than status updates (aka tactical updates), which are aimed at incident participants/ responders.

This PR introduces a new env var called `INCIDENT_STORAGE_INCIDENT_UPDATE_FILE_ID`.


kevgliss:
With so much overlap between status reports and these higher level status updates is there an opportunity here to consolidate between them into a "reports" module?  It seems rather limiting that every time we want to create a new kind of report we would create an entirely new service/table/view for that data, when essentially we would only change the fields themselves and maybe a template?

mvilanova:
It's more than fields and a template, but I see what you're saying. Let me rethink it.

PR #301, opened 2020-05-11
Feature/task filtering

kevgliss:


src/dispatch/task/flows.py
```diff
@@ -79,36 +84,46 @@ def send_task_notification(conversation_id, message_template, assignees, descrip
...
+                    INCIDENT_TASK_RESOLVED_NOTIFICATION,
+                    assignees,
+                    description,
+                    weblink,
+                )
+
+        # always sync assignees
+        incident_task.assignees = assignees
+        db_session.add(incident_task)
+        db_session.commit()
```
mvilanova, line 126: I think the logic we had before was correct. You need to check the status of the task that gets passed, not the one in the database. If the status is open, then we don't need to update it and we return. If already resolved in the doc and database, then we return. If not resolved in the db, then we resolve it.

PR #320, opened 2020-05-15
Add PKCE overrides and support the key rotation

willbengtson:
Adds the ability to include an audience in `jwt.decode` as well as not rely on email address to be in `email` key of the JWT.  This also adds the support for key rotation when the IDP sends back more than one key in the key fetch.

### DIFF
--- a/src/dispatch/plugins/dispatch_core/config.py
+++ b/src/dispatch/plugins/dispatch_core/config.py
@@ -1,6 +1,20 @@
+import logging
+
 from starlette.config import Config
+
+
+log = logging.getLogger(__name__)
+
 
 config = Config(".env")
 
-DISPATCH_JWT_AUDIENCE = config("DISPATCH_JWT_AUDIENCE")
-DISPATCH_JWT_EMAIL_OVERRIDE = config("DISPATCH_JWT_EMAIL_OVERRIDE")
+
+DISPATCH_JWT_AUDIENCE = config("DISPATCH_JWT_AUDIENCE", default=None)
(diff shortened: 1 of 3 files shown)


src/dispatch/plugins/dispatch_core/config.py
```diff
@@ -1,3 +1,6 @@
 from starlette.config import Config
 
 config = Config(".env")
+
+DISPATCH_JWT_AUDIENCE = config("DISPATCH_JWT_AUDIENCE")
```
kevgliss, line 5: Can these be gated by if the plugin PCKE plugin is actually in use? See [here](https://github.com/Netflix/dispatch/blob/develop/src/dispatch/config.py#L85) for an example. 

We could also probably move those values to the dispatch_core config as well... But I can tackle that in a separate PR.

src/dispatch/plugins/dispatch_core/config.py
```diff
@@ -1,3 +1,20 @@
...
 
+
+log = logging.getLogger(__name__)
+
+
 config = Config(".env")
+
+
+DISPATCH_JWT_AUDIENCE = config("DISPATCH_JWT_AUDIENCE", default=None)
+DISPATCH_JWT_EMAIL_OVERRIDE = config("DISPATCH_JWT_EMAIL_OVERRIDE", default=None)
```
mvilanova, line 13: Can we document these envvars in https://hawkins.gitbook.io/dispatch/configuration/app?

willbengtson: Updated

PR #322, opened 2020-05-16
Adds as join incident button to the incident status page

kevgliss:


src/dispatch/incident/views.py
```diff
@@ -154,6 +158,26 @@ def update_incident(
     return incident
 
 
+@router.post("/{incident_id}/join", summary="Join an incident.")
+def join_incident(
+    *,
+    db_session: Session = Depends(get_db),
+    incident_id: str,
```
mvilanova, line 165: nit: shouldn't `incident_id` be of type `int`? Same goes for the one in `service.get()`

PR #327, opened 2020-05-18
Add custom profile field mapping for slack to get Team, Department, W…

willbengtson:
Adds the ability to populate the Department, Team, and weblink fields when listing incident participants.

These are returned in the `users.profile.get` API call to slack in the `fields` dictionary.  

Adds 3 new config variables (optional):
- `SLACK_PROFILE_DEPARTMENT_FIELD_ID`
- `SLACK_PROFILE_TEAM_FIELD_ID`
- `SLACK_PROFILE_WEBLINK_FIELD_ID`

Updates the documentation to add the new scope required (breaking change): `users.profile:read`


docs/configuration/plugins/configuring-slack.md
```diff
@@ -72,6 +72,18 @@ The `Slack` plugin relies on the [Events API](https://api.slack.com/events-api)
 
 > List resources command as displayed in Slack.
 
+#### `SLACK_DEPARTMENT_PROFILE_ID`
```
mvilanova, line 75: nit: I think it would be easier to know what info needs to be specified if we name them like `SLACK_PROFILE_DEPARTMENT_FIELD_ID`

willbengtson: 👍 will update

willbengtson: Updated

mvilanova:
Looks like a `docker-compose.yml` and an `.env.bak` file were included in this commit: https://github.com/Netflix/dispatch/pull/327/commits/78895308ff007198c6b3f4ccbff61239efce1a06

willbengtson:
derp, should be fixed

src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -42,6 +45,7 @@
     get_user_email,
     get_user_info_by_id,
     get_user_info_by_email,
```
mvilanova, line 47: Looks like we don't need this import here anymore. Mind removing it? We can merge the code afterwards.

willbengtson: removed

PR #328, opened 2020-05-18
Decoupling some components 

stefanm8:
Hello this PR tries to resolve some of the "tightly-coupled components" issue. 

- Contact plugin 
Removes the need for slack integration for ContactPlugin, and uses dispatch itself to resolve a individual
- Opsgenie plugin 
Second oncall plugin 

Opinions welcomed:
- If individual is not found (fetched from oncall-plugin) should it be created within dispatch, for further use?
- Should I include in this PR google hangouts, and google meet to get rid of slack necessity, leveraging a much easier dispatch installation as you would only have to configure and integrate it with google.

@kevgliss what do you think?

### DIFF
--- a/src/dispatch/individual/service.py
+++ b/src/dispatch/individual/service.py
@@ -7,6 +7,7 @@ from dispatch.incident_priority import service as incident_priority_service
 from dispatch.incident_type import service as incident_type_service
 from dispatch.plugins.base import plugins
 from dispatch.term import service as term_service
+from dispatch.plugin import service as plugin_service
 
 from .models import IndividualContact, IndividualContactCreate, IndividualContactUpdate
 
@@ -43,7 +44,7 @@ def get_or_create(*, db_session, email: str, **kwargs) -> IndividualContact:
     contact = get_by_email(db_session=db_session, email=email)
 
     if not contact:
-        contact_plugin = plugins.get(INCIDENT_PLUGIN_CONTACT_SLUG)
+        contact_plugin = plugin_service.get_active(db_session=db_session, plugin_type="contact")
         individual_info = contact_plugin.get(email, db_session=db_session)
         kwargs["email"] = individual_info.get("email", email)
         kwargs["name"] = individual_info.get("fullname", "unknown")
--- a/src/dispatch/individual/service.py
+++ b/src/dispatch/individual/service.py
@@ -44,10 +44,10 @@ def get_or_create(*, db_session, email: str, **kwargs) -> IndividualContact:
 
     if not contact:
         contact_plugin = plugins.get(INCIDENT_PLUGIN_CONTACT_SLUG)
-        individual_info = contact_plugin.get(email)
-        kwargs["email"] = individual_info["email"]
-        kwargs["name"] = individual_info["fullname"]
-        kwargs["weblink"] = individual_info["weblink"]
+        individual_info = contact_plugin.get(email, db_session=db_session)
+        kwargs["email"] = individual_info.get("email", email)
+        kwargs["name"] = individual_info.get("fullname", "unknown")
+        kwargs["weblink"] = individual_info.get("weblink", "unknown")
         individual_contact_in = IndividualContactCreate(**kwargs)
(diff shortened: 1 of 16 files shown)


src/dispatch/service/service.py
```diff
@@ -41,6 +42,7 @@ def create(*, db_session, service_in: ServiceCreate) -> Service:
         incident_types=incident_types,
         terms=terms,
     )
+    service.type = ONCALL_PLUGIN_SLUG
```
stefanm8, line 45: Not sure if this is the only place where "pagerduty_oncall" is referenced, if you would like to help me with this @kevgliss would be awesome.

kevgliss:
> If individual is not found (fetched from oncall-plugin) should it be created within dispatch, for further use?

Yes, creating a new individual based on the on call (if they don't exist) makes sense (we do the same thing with contact plugins). If fact, I believe we already do this for existing oncall-plugins. We create a new participant which will, in turn, create a new individual if one doesn't exist. 

For reference see:
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/incident/service.py#L200
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/incident/service.py#L28
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/participant/flows.py#L26

> Should I include in this PR google hangouts, and google meet to get rid of slack necessity, leveraging a much easier dispatch installation as you would only have to configure and integrate it with google.

I think the changes should be in separate PRs to keep readability high.

src/dispatch/service/service.py
kevgliss: afaik we only use the oncall_plugins in two places:

Here:
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/incident/service.py#L42

and:
https://github.com/Netflix/dispatch/blob/develop/src/dispatch/incident/flows.py#L1088

Recently, we made a change that allows plugins to be configurable via the UI (instead of strictly via the config). I think in both places we should replace the _plugin.get_ calls with 
`plugin = plugin_service.get_active(db_session=db_session, plugin_type="service")` to enable this behavior (we're slowly migrating plugins as we add more of the same type).

Additionally, if you're up for it (this is specifically for services) we have on our backlog to be able to _specify_ the service's plugin (allowing you to mix and match services/plugins). Today we had just hardcoded pagerduty, but if we want folks to be able to specify ops-genie without knowing it's slug we should provide them a UI dropdown.

stefanm8: Thank you for suggestions and hints, I've implemented the dropdown as you suggested

src/dispatch/individual/service.py
```diff
@@ -44,10 +44,10 @@ def get_or_create(*, db_session, email: str, **kwargs) -> IndividualContact:
 
     if not contact:
         contact_plugin = plugins.get(INCIDENT_PLUGIN_CONTACT_SLUG)
```
kevgliss, line 46: I think we need something like: `plugin_service.get_active(db_session=db_session, plugin_type="contact")`

src/dispatch/plugins/dispatch_core/plugin.py
```diff
@@ -162,6 +163,23 @@ def get(
...
+    title = "Dispatch Plugin - Contact plugin"
+    slug = "dispatch-contact"
+    description = "Uses dispatch itself to resolve incident participants."
+    version = dispatch_plugin.__version__
+
+    author = "Netflix"
+    author_url = "https://github.com/netflix/dispatch.git"
+
+    def get(self, email, db_session=None):
+        return getattr(
```
kevgliss, line 176: Why not just access the values here directly like:
```
i =  individual_service.get_by_email(db_session=db_session, email=email),
return {"email": i.email, "fullname": i.email}
```

stefanm8: It may return None in case the individual does not exist, e.g when it tries to create the individual as it doesn't exist, it retrieves data from contact plugin (which this case individual table)

kevgliss:
lgtm thanks for the PR!

PR #336, opened 2020-05-19
Add the ability to use a different Jira project depending on Incident…

willbengtson:
Adds the ability to use a different Jira depending on the incident type. Adds a JSON column to the `incident_type` DB table which allows you to place JSON data for the given plugin. The JSON first level keys are plugin slug.

Example data:

```
{
  "dispatch-jira-plugin": {
    "project_key": "TEST123",
    "issue_type_id": 10244
  }
}
```

Also add a name to the unique constraint from the previous db migration.


src/dispatch/incident_type/models.py
```diff
@@ -15,6 +15,7 @@ class IncidentType(Base):
     slug = Column(String)
     description = Column(String)
     visibility = Column(String, default=Visibility.open)
+    plugin_metadata = Column(JSON)
```
kevgliss, line 18: If we set the default to be an empty dict via `plugin_metadata = db.Column(JSON, nullable=False, default=dict, server_default='{}')` I think we can remove a lot of checks for `None` and returning empty dicts.

src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -222,6 +223,7 @@ def update(
         storage_weblink: str = None,
         labels: List[str] = None,
         cost: int = None,
+        incident_type_plugin_metadata: dict = {},
```
kevgliss, line 226: Is this still a WIP? Looks like we haven't let Jira read the incident metadata.

willbengtson:
Updated to default dict `{}` for `plugin_metadata`.  Will PR generic Jira plugin in another PR, does that work?

src/dispatch/incident_type/models.py
willbengtson: updated

src/dispatch/plugins/dispatch_jira/plugin.py
willbengtson: Can port generic Jira plugin in another PR?

src/dispatch/incident/flows.py
```diff
@@ -650,6 +659,10 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
     set_conversation_topic(incident)
 
     # we update the incident ticket
+    incident_type_data = incident_type_service.get_by_name(
+        db_session=db_session, name=incident.incident_type.name
+    )
+
```
mvilanova, line 665: Can we move it inside `update_incident_ticket`, so we don't have to add this code to all the places where we call `update_incident_ticket`? Same for when we call `create_incident_ticket`.

src/dispatch/alembic/versions/9b4a5ff08170_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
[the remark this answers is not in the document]
willbengtson: Dang I always forget this :(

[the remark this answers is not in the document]
mvilanova: No worries :)

src/dispatch/incident/flows.py
willbengtson: updated. Had to add passing in the incident type name in a few places.

src/dispatch/alembic/versions/9b4a5ff08170_.py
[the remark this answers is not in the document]
willbengtson: updated

src/dispatch/incident/flows.py
```diff
@@ -499,6 +510,10 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
     )
 
     # create the incident ticket
+    incident_type_data = incident_type_service.get_by_name(
+        db_session=db_session, name=incident.incident_type.name
+    )
+
```
mvilanova: Forgot to remove this one?

willbengtson: yeah I updated it

mvilanova: All good. Looks like I wasn't looking at the latest version. Thanks

src/dispatch/plugins/dispatch_jira/plugin.py
kevgliss: Separate PR is fine.

PR #341, opened 2020-05-19
switch gpg keyserver due to missing keys, fix syntax error in dockerfile

### LINKED ISSUE #340: Dockerfile issues
The dockerfile in master/develop is broken, it has invalid syntax.

Beyond the syntax fix, there are also some issues building due to some GPG keys seemingly not being available on the chosen keyserver.

bsundsrud:
Fix a syntax issue in the Dockerfile, switch to sks-pools for the keyserver as several keys were missing from keys.openpgp.org (including the one for `gosu`)

Removed the bits messing with `$GPUGPHOME` since it didn't seem necessary and none of the other steps did anything similar.

Closes #340

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,18 +36,17 @@ RUN for key in \
     A48C2BEE680E841632CD4E44F07496B3EB3C1762 \
     B9E2F5981AA6E0CD28160D9FF13993A75599653C \
     ; do \
-    gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys "$key"; \
+    gpg --batch --keyserver hkps://hkps.pool.sks-keyservers.net --recv-keys "$key"; \
     done
 
 # grab gosu for easy step-down from root
 ENV GOSU_VERSION 1.12
 RUN set -x \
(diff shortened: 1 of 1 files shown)


kevgliss:
Thanks for the PR.

mvilanova:
Thanks for the fixes. I wanted to clarify that the reason why verification was failing was not because the keys don't exist in `keys.openpgp.org`, but because `The identity information in an OpenPGP key is only distributed with consent` according to https://keys.openpgp.org/about/. Looks like some of the key owners have not consented, which is causing current versions of GnuPG to skip their keys: `key 9A84159D7001A4E5: new key but contains no user ID - skipped`.

Servers in the SKS pool are not recommended for various reasons (see https://keys.openpgp.org/about/faq#sks-pool, https://code.firstlook.media/the-death-of-sks-pgp-keyservers-and-how-first-look-media-is-handling-it).

We should use `keyserver.ubuntu.com` until GnuPG has fixed [this](https://keys.openpgp.org/about/faq#older-gnupg) issue, then switch back to `keys.openpgp.org`.

PR #356, opened 2020-05-21
Generalizes report data model and adds support for executive reports

mvilanova:
This PR adds support for writing and distributing executive reports. These reports are aimed at leadership and stakeholders.

This PR introduces a new env var called `INCIDENT_STORAGE_EXECUTIVE_REPORT_FILE_ID`.


src/dispatch/report/models.py
```diff
@@ -0,0 +1,57 @@
...
+from dispatch.database import Base
+from dispatch.models import DispatchBase
+
+from .enums import ReportTypes
+
+
+class Report(Base):
+    id = Column(Integer, primary_key=True)
+    created_at = Column(DateTime, default=datetime.utcnow)
+    conditions = Column(String)
```
kevgliss, line 17: This feels likes like we just combined two models (instead of making something more generic/extensible).

src/dispatch/incident/models.py
```diff
@@ -134,8 +137,19 @@ def incident_faq(self):
 
     @hybrid_property
     def last_status_report(self):
-        if self.status_reports:
-            return sorted(self.status_reports, key=lambda r: r.created_at)[-1]
+        if self.reports:
+            status_reports = [
```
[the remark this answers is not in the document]
mvilanova: I'm filtering them and returning the last one in the list, which should be the most recent one?

src/dispatch/report/models.py
mvilanova: Yeah, perhaps we could store the report details in a JSON field?

src/dispatch/incident/models.py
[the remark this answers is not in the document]
kevgliss: I don't believe that order it's guaranteed, besides insertion order !=  created_at order.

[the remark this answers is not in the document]
mvilanova: Ok, I'll sort them.

src/dispatch/alembic/versions/dd3df6a3af3c_.py
```diff
@@ -0,0 +1,81 @@
...
+    op.drop_column("report", "actions")
+    op.drop_column("report", "needs")
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("ix_report_search_vector", table_name="report")
+    op.drop_constraint(None, "document", type_="foreignkey")
+    op.drop_column("document", "report_id")
```
kevgliss, line 66: We don't have a good story here, but a downgrade would cause you to loose data, maybe it's better off not to drop the old columns?

src/dispatch/incident/models.py
```diff
@@ -133,9 +136,22 @@ def incident_faq(self):
                     return d
 
     @hybrid_property
-    def last_status_report(self):
-        if self.status_reports:
-            return sorted(self.status_reports, key=lambda r: r.created_at)[-1]
+    def last_tactical_report(self):
+        if self.reports:
```
kevgliss, line 140: Do we need a property to fetch all tactical reports (not just last)? Perhaps we break this function into two hybrid properties that call each other (tactical_reports and last_tactical_reports).

src/dispatch/report/flows.py
```diff
@@ -0,0 +1,216 @@
...
+
+    return tactical_report
+
+
+@background_task
+def create_executive_report(
+    user_id: str, user_email: str, incident_id: int, action: dict, db_session=None
+):
+    """Creates an executive report."""
+    current_date = (date.today()).strftime("%B %d, %Y")
```
kevgliss, line 81: Is there an extra set of parens here?

src/dispatch/report/models.py
```diff
@@ -0,0 +1,63 @@
...
+    participant_id = Column(Integer, ForeignKey("participant.id"))
+    document = relationship("Document", uselist=False, backref="report")
+
+    # full text search capabilities
+    search_vector = Column(TSVectorType("details_raw"))
+
+    @staticmethod
+    def _details_raw(mapper, connection, target):
+        target.details_raw = "; ".join(
+            [f"{k.capitalize()}: {v}" for k, v in target.details.items()]
```
kevgliss, line 34: I'm not sure if it will affect searching, but I would rather we just concat all the raw strings together, i.e. if I search for `conditions` it shouldn't return every tactical report.

src/dispatch/alembic/versions/dd3df6a3af3c_.py
mvilanova: Yeah, let's keep them.

src/dispatch/incident/models.py
mvilanova: Sure, I can break them into two.

src/dispatch/report/flows.py
mvilanova: Yup, will remove the extra ones.

src/dispatch/report/models.py
mvilanova: `details_raw` will contain something like this, for example: `Conditions: These are the conditions; Actions: These are the actions; Needs: These are the needs`. Isn't that what we want?

kevgliss: Yes, but if you include the word `condition` if I search for `condition` all status reports will be returned (since you've included that word) even though it never showed up in the original condition text. Additionally, that column will never be used for anything other than indexing so we don't really need to add that metadata.

mvilanova: Ah, got it. I'll remove the key.

PR #368, opened 2020-05-27
Adds UI and refactors the metadata json format

kevgliss:
Adds a UI to incident types to be able to edit incident type plugin metadata:

Also modifies the json structure a bit (and adds appropriate pydantic validation). See:
```
       [  
         {
            slug: "dispatch-jira-plugin",
            metadata: [
              { key: "project_key", value: "TEST123" },
              { key: "issue_type_id", value: "10244" }
            ]
          }
        ]
```


src/dispatch/alembic/versions/cfd62f719c84_.py
```diff
@@ -0,0 +1,29 @@
+"""empty message
```
mvilanova, line 1: Add description plz.

src/dispatch/alembic/versions/cfd62f719c84_.py
```diff
@@ -0,0 +1,29 @@
...
+revision = "cfd62f719c84"
+down_revision = "5693756fd192"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("incident_type", "plugin_metadata")
+    op.add_column("incident_type", sa.Column("plugin_metadata", sa.JSON(), nullable=True))
```
mvilanova, line 22: Why not using [`alter_column()`](https://alembic.sqlalchemy.org/en/latest/ops.html?highlight=insert#alembic.operations.Operations.alter_column)? Aren't we going to lose the data if we drop the column?

src/dispatch/alembic/versions/cfd62f719c84_.py
```diff
@@ -0,0 +1,29 @@
+"""empty message
+
+Revision ID: cfd62f719c84
+Revises: 5693756fd192
+Create Date: 2020-05-26 20:17:47.707953
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
```
mvilanova, line 10: This import is not necessary.

src/dispatch/incident_type/models.py
```diff
@@ -15,7 +18,7 @@ class IncidentType(Base):
     slug = Column(String)
     description = Column(String)
     visibility = Column(String, default=Visibility.open)
-    plugin_metadata = Column(JSON, nullable=False, default=dict, server_default="{}")
+    plugin_metadata = Column(JSON, default=[])
```
mvilanova, line 21: Shouldn't the default be an empty dict here?

src/dispatch/alembic/versions/cfd62f719c84_.py
willbengtson: I think he did this to ensure any data that existed gets dropped since the format of the JSON has changed.

willbengtson:
I should be able to test this today

src/dispatch/incident_type/models.py
kevgliss: No, the new json format is a list of ducts.

PR #370, opened 2020-05-27
Adds the ability to filter tasks by associated incidents.

kevgliss:

### DIFF
--- a/src/dispatch/static/dispatch/src/incident/IncidentCombobox.vue
+++ b/src/dispatch/static/dispatch/src/incident/IncidentCombobox.vue
@@ -1,0 +1,109 @@
+<template>
+  <v-combobox
+    v-model="incident"
+    :items="items"
+    item-text="name"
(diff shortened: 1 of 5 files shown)


src/dispatch/static/dispatch/src/incident/IncidentCombobox.vue
```diff
@@ -0,0 +1,109 @@
...
+      <v-list-item-content>
+        <v-list-item-title v-text="item.name"></v-list-item-title>
+        <v-list-item-subtitle v-text="item.title"></v-list-item-subtitle>
+      </v-list-item-content>
+    </template>
+    <template v-slot:no-data>
+      <v-list-item>
+        <v-list-item-content>
+          <v-list-item-title>
+            No Incidents matching "
```
mvilanova, line 32: nit
```suggestion
            No incidents matching "
```

PR #376, opened 2020-05-28
Checks Slack profile fields for None before accessing them

mvilanova:
This PR fixes a `NoneType` exception when Slack fields are `None`.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/plugin.py
+++ b/src/dispatch/plugins/dispatch_slack/plugin.py
@@ -198,14 +198,16 @@ class SlackContactPlugin(ContactPlugin):
 
     def get(self, email: str, **kwargs):
         """Fetch user info by email."""
-        team = department = weblink = ""
+        team = department = weblink = "Unknown"
 
         profile = get_user_profile_by_email(self.client, email)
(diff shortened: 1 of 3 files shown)


mvilanova:
We could also move the logic down to the `get_user_profile_by_email()` function in the service module.

src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -198,20 +198,23 @@ def __init__(self):
 
     def get(self, email: str, **kwargs):
         """Fetch user info by email."""
+        team = department = weblink = ""
+
         profile = get_user_profile_by_email(self.client, email)
+        profile_fields = profile.get("fields")
+        if profile_fields:
+            team = profile_fields.get(SLACK_PROFILE_TEAM_FIELD_ID, {}).get("value", "")
```
kevgliss, line 206: This type of think this has come up in a couple of different places, is empty string good or is something more explicit like "Unknown" better?

mvilanova: Let's make it explicit here like we do in the participant and individual service modules.

PR #383, opened 2020-05-30
Adds last executive report to active incidents status page

mvilanova:
Screen Shot 2020-06-01 at 2 48 20 PM


kevgliss:
Do you have a screenshot? Should we put the reports into tabs?

mvilanova:
@kevgliss see attached screenshot. I think it looks good w/o tabs for now.

PR #399, opened 2020-06-02
Adds data model for incident cost

mvilanova:


src/dispatch/alembic/versions/0797208b9eee_.py
```diff
@@ -0,0 +1,85 @@
...
+        sa.Column("details", sqlalchemy_utils.types.json.JSONType(), nullable=True),
+        sa.Column("incident_id", sa.Integer(), nullable=True),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
+        sa.Column("updated_at", sa.DateTime(), nullable=True),
+        sa.ForeignKeyConstraint(["incident_id"], ["incident.id"],),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_table(
+        "pci_disclosure_fines_judgements",
+        sa.Column("id", sa.Integer(), nullable=False),
```
kevgliss, line 34: Didn't you say these cost ranges are mostly standardized? If so, does it make sense to have them as tables instead of constants? Or are we just allowing flexibility for other industries?

mvilanova: Yeah, I'd like to give folks the flexibility of changing them if they want.

mvilanova:
We're going to take a different approach.

PR #408, opened 2020-06-04
Allows for customizing contact information in email notifications

mvilanova:


src/dispatch/incident/scheduled.py
```diff
@@ -221,14 +221,14 @@ def daily_summary(db_session=None):
             }
         )
 
-    # NOTE INCIDENT_DAILY_SUMMARY_ONCALL_SERVICE_ID is optional
-    if INCIDENT_DAILY_SUMMARY_ONCALL_SERVICE_ID:
+    # NOTE INCIDENT_ONCALL_SERVICE_ID is optional
+    if INCIDENT_ONCALL_SERVICE_ID:
```
kevgliss, line 225: In the future, I think we could move this to a flag on the service object.

src/dispatch/plugins/dispatch_google/gmail/templates/executive_report.html
```diff
@@ -26,8 +26,6 @@
   <li>{{ next_steps }}</li>
 </ul>
 <p style="font-weight:normal!important;line-height:140%;margin-top:0">
-  A Google document version of the report can be found <a href="{{ weblink }}">here</a>. Please, use the document for questions and feedback.<br><br>
-  Thanks,<br>
-  {{ commander_fullname}}
+  A Google document version of the report can be found <a href="{{ weblink }}">here</a>. Please, use the document for questions and feedback.
```
kevgliss, line 29: Should we be more generic and drop Google here? There could be different plugins at play (today we just have google).

kevgliss:
Just so I'm clear, this PR allows you to specify a default contact (`dispatch_help`) or allows you to provide a service that resolves the contact depending on when the message is sent?

Also, please modify the docs for the new configuration variable. Also since we are deprecating a configuration item we should first warn (in the docs and the code) that we are deprecating it for a bit before removing it. 

I think in this case since it's just a rename we can just alias them until we remove.

mvilanova:
It allows you to provide the contact information you want to use in the base template footer. In the case of the task reminder, we're using the current oncall (if `INCIDENT_ONCALL_SERVICE_ID` is set), as it seems to be the best option. For all the other cases, the incident commander, so we keep discussions contained within the incident.

PR #409, opened 2020-06-05
Dockerfile bugfix: Add backslash on line 52

jakestenger:
newline wasn't escaped on line 52, causing parse error for line 53


mvilanova:
@jakestenger looks like master is in a bad state after our last release. Let me try to fix it.

mvilanova:
Ok, master is fixed. Thanks for the PR.

PR #417, opened 2020-06-05
Adding a new command to display only the requesting user's tasks

kevgliss:


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -223,8 +224,44 @@ def after_hours(user_email: str, incident_id: int, event: dict = None, db_sessio
...
+            creator_email = t.creator.individual.email
+            if creator_email == by_creator:
+                filtered_tasks.append(t)
+
+        if by_assignee:
+            assignee_emails = [a.individual.email for a in t.assignees]
+            if by_assignee in assignee_emails:
+                filtered_tasks.append(t)
+
+    return filtered_tasks
```
mvilanova, line 241: Should we move this to the task service and make it a db query instead?

kevgliss: We could but the query will be fairly complex and the task lists are normally small so filtering after the fact seems reasonable.

PR #418, opened 2020-06-05
Adding feedback when a user uses the join incident button

kevgliss:


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -561,6 +561,10 @@ def add_user_to_conversation(
         dispatch_slack_service.add_users_to_conversation(
             slack_client, incident.conversation.channel_id, [user_id]
         )
+        message = f"Success! We've added you {incident.name}. Please check your side bar for the new channel."
```
mvilanova, line 564:
```suggestion
        message = f"Success! We've added you to incident {incident.name}. Please check your side bar for the new channel."
```

PR #426, opened 2020-06-08
Don't try to load images for large incidents

kevgliss:


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -339,25 +339,24 @@ def list_participants(incident_id: int, command: dict = None, db_session=None):
...
+                        f"*Team*: {participant_team}, {participant_department}\n"
+                        f"*Location*: {participant_location}\n"
+                        f"*Incident Role(s)*: {(', ').join(participant_roles)}\n"
+                    ),
+                },
+                "accessory": {"type": "image", "alt_text": participant_name},
+            }
+
+            if len(participants) < 20:
+                block["accessory"].update({"image_url": participant_avatar_url})
```
mvilanova, line 357: I'd suggest we either add the whole accessory (`type`, `image_url`, and `image_url`) or we don't, rather than updating it if the condition is met. Or, is there a particular reason why you're doing it this way?

mvilanova: We could also lower the image resolution: https://github.com/Netflix/dispatch/blob/develop/src/dispatch/plugins/dispatch_slack/service.py#L218

kevgliss: No particular reason. Will add/remove the accessory data itself.

PR #427, opened 2020-06-08
Allowing closed incidents to be updated.

kevgliss:


src/dispatch/incident/flows.py
```diff
@@ -943,8 +942,8 @@ def incident_update_flow(
         )
 
     if conversation_topic_change:
-        # we update the conversation topic
-        set_conversation_topic(incident)
+        if incident.status != IncidentStatus.closed:
+            set_conversation_topic(incident)
 
     if notify:
         send_incident_update_notifications(incident, previous_incident)
```
mvilanova, line 949: Looks like the default value for `notify` is `True`, which means that we will also try to send a message to the conversation. We should protect against it as well.

PR #439, opened 2020-06-12
Fix misspelling of `received`

caitlin615:
Fixes misspelling of `received` in src/dispatch/static/dispatch/src/api.js


mvilanova:
Thanks for the fix!

Issue #406, opened 2020-06-04
Getting Internal server error 500 for many api calls


Martin1196155:
I run the ./install.sh and i refer develop branch of the dispatch. When i am loading web UI i am getting below error
![image](https://user-images.githubusercontent.com/46367679/83797359-2735f280-a6c0-11ea-8318-98b3a9adda67.png)

In ngrok console i found few api call got internal server error 500.

![image](https://user-images.githubusercontent.com/46367679/83797426-403ea380-a6c0-11ea-95b2-d9c2ec9b975f.png)

after that i click every option in the left hand side and i am getting the same error.

![image](https://user-images.githubusercontent.com/46367679/83797535-71b76f00-a6c0-11ea-979b-5b40a161a669.png)

Can anyone please help me to resolve this issue.

mvilanova:
@Martin1196155 the error in the first screen is expected when there's no data in the database. @kevgliss we may want to provide a less scary, more informative message here.

kevgliss:
The errors could be lack of data or something else entirely, we would need the error logs from the server to know. @Martin1196155 could you provide those logs?

Martin1196155:
@mvilanova @kevgliss Please find the attachment contains the docker logs for the dispatch web container.
[Logs.txt](https://github.com/Netflix/dispatch/files/4737910/Logs.txt)

kevgliss:
I don't see the server startup in those logs, but based on `2020-06-05T15:26:58.800406049Z KeyError: 'dispatch-auth-provider-basic'` it's likely that your database is not up to date. Please try and run `dispatch database upgrade`

gargrag:
Hi I'm having the same error on a fresh install, I did run the `dispatch database upgrade` 

![image](https://user-images.githubusercontent.com/967311/84549126-1424b300-acbc-11ea-80f5-ecc8c41de64e.png)

I'm still debugging to see if I can find something else.

gargrag:
@Martin1196155  OK! got it, you DO NEED to set `DISPATCH_JWT_SECRET=something` if you want to get rid of the error.

Martin1196155:
I think the new update in the dispatch-docker works. I registering by using https://XXXXXX:8000/register. Now i am not getting this issue.

Martin1196155:
But still i am getting error while creating incident. 
![image](https://user-images.githubusercontent.com/46367679/84672471-37fb2980-af46-11ea-9759-15c699209585.png)
[...]

Issue #360, opened 2020-05-25
Issues with Basic Auth


bsundsrud:
I'm trying to get basic auth set up and I have a few issues.  I'm not using the dispatch-docker project as I'm trying to deploy to kubernetes.

1) I've set `DISPATCH_AUTHENTICATION_PROVIDER_SLUG=dispatch-auth-provider-basic`, I was running into issues with the Vue app not redirecting to /login until I realized I had to set `VUE_APP_DISPATCH_AUTHENTICATION_PROVIDER_SLUG` at container build time, as well.  Even with this, `/incidents/report` fails to get Type and Priority because the JWT token is not sent to the API, even though it is present in local storage.  It seems the Authorization header isn't being set correctly or in all places?

When visiting `/incidents/report` with a valid token in local storage:
```
INFO:     172.16.7.67:58766 - "GET /api/v1/incident_priorities/?sortBy[]=view_order&descending[]=false HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 102, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 26, in __call__
    await response(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/responses.py", line 197, in __call__
    async for chunk in self.body_iterator:
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 56, in body_stream
    task.result()
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 38, in coro
    await self.app(scope, receive, send)
[...]

PR #440, opened 2020-06-12
Mention that the basic auth plugins, requires DISPATCH_JWT_TOKEN

gargrag:
https://github.com/Netflix/dispatch/issues/360
https://github.com/Netflix/dispatch/issues/406


mvilanova:
Thanks for the PR!

PR #441, opened 2020-06-15
Fixing google docs permissions issue

tomarv2:
I noticed that the permissions should be: `https://www.googleapis.com/auth/documents` and not `https://www.googleapis.com/auth/document`


mvilanova:
@tomarv2 thanks for the PR. The doc fixes look good. Would you mind removing all the other changes, so we can merge them?

mvilanova:
Closes https://github.com/Netflix/dispatch/issues/401.

PR #447, opened 2020-06-16
Updating Jira Plugin to pass as a variable in .env file

kstringlo:
I have modified the Jira plugin to be able to log in for a particular UserID in Jira for all the reported incidents in dispatch.


kevgliss:
I don't think this would be desirable, we often have dynamic assignee's and reporters (as opposed static ids). What was your use case for assigning these statically? 

We could have an optional flag to use static values but I don't think that should be the default for most user.

kstringlo:
We can have scenarios where I want to use dispatch to use it as a tool to report the incident. And it triggers a Jira issue in my workflow. However, I do not want to provide a Jira user account to all of my reporters. (If I'm expecting many reporters, that will cause that I will have to provide to all the reporters a Jira account; that will affect my budget just to use this tool). But I can provide them a way to report incidents (trough dispatch) without adding any additional cost.
@kevgliss

kevgliss:
Okay; if it purely a cost issue, then yes you could have static users for assignees and reporters. But I think this should be optional for most folks... Perhaps we create a .env item `JIRA_STATIC_USERS` and only if that is true then we assign the static users?

kevgliss:
Feel free to re-open this PR with the requested changes. Closing for now.

PR #461, opened 2020-06-23
Initial work. Refactors how common documents are resolved.

kevgliss:
This starts us down the road of making some of this optional in order to help ease adoption. For instance, with these changes, the FAQ document and Conversation Reference documents are now optional and will not be rendered in messaging. 

Additional PRs will address making the incident document optional.
This PR also includes some small UI improvements:


src/dispatch/report/flows.py
```diff
@@ -80,6 +78,11 @@ def create_executive_report(
     user_id: str, user_email: str, incident_id: int, action: dict, db_session=None
 ):
     """Creates an executive report."""
+    report_template = document_service.get_executive_report_template(db_session=db_session)
+
+    if not report_template:
```
kevgliss, line 83: I don't think we have a good story around reporting errors back to a slack user. cc @mvilanova for ideas.

docs/configuration/app.md
```diff
@@ -193,14 +193,6 @@ Dispatch [calculates](https://github.com/Netflix/dispatch/blob/develop/src/dispa
...
 ### Incident Resource Configuration
 
-#### `INCIDENT_FAQ_DOCUMENT_ID`
-
-> Controls which document id to use as the FAQ.
-
-#### `INCIDENT_CONVERSATION_COMMANDS_REFERENCE_DOCUMENT_ID`
-
-> Controls which document id to use for the conversation commands reference document.
-
```
mvilanova, line 203: Shouldn't we note that they're optional rather than removing them? Otherwise, how are users going to know they're at their disposal?

src/dispatch/config.py
```diff
@@ -190,24 +183,35 @@ def __str__(self) -> str:
 INCIDENT_RESOURCE_NOTIFICATIONS_GROUP = config(
     "INCIDENT_RESOURCE_NOTIFICATIONS_GROUP", default="google-group-participant-notifications-group"
 )
+INCIDENT_RESOURCE_INVESTIGATION_SHEET_TEMPLATE = config(
+    "INCIDENT_RESOURCE_INVESTIGATION_SHEET_TEMPLATE", default="dispatch-incident-sheet-template"
+)
```
mvilanova, line 188: Should we take the opportunity and prefix them with `DISPATCH` rather than `INCIDENT`, so we don't use `INCIDENT` twice and are consistent with the default value?
```suggestion
DISPATCH_RESOURCE_INCIDENT_SHEET_TEMPLATE = config(
    "DISPATCH_RESOURCE_INCIDENT_SHEET_TEMPLATE", default="dispatch-incident-sheet-template"
)
```

src/dispatch/config.py
```diff
@@ -190,24 +183,35 @@ def __str__(self) -> str:
...
+    "INCIDENT_RESOURCE_INCIDENT_REVIEW_DOCUMENT_TEMPLATE",
+    default="dispatch-incident-review-document-template",
+)
+INCIDENT_RESOURCE_EXECUTIVE_REPORT_DOCUMENT_TEMPLATE = config(
+    "INCIDENT_RESOURCE_EXECUTIVE_REPORT_DOCUMENT_TEMPLATE",
+    default="dispatch-executive-report-document-template",
+)
 INCIDENT_RESOURCE_INVESTIGATION_DOCUMENT = config(
-    "INCIDENT_RESOURCE_INVESTIGATION_DOCUMENT", default="google-docs-investigation-document"
+    "INCIDENT_RESOURCE_INVESTIGATION_DOCUMENT", default="dispatch-incident-document"
```
mvilanova, line 198: Changing the default value will break older incidents, as Dispatch won't be able to locate the documents using the resource type. Are you planning to update them?

src/dispatch/incident/messaging.py
```diff
@@ -63,28 +61,41 @@ def send_welcome_ephemeral_message_to_participant(
...
+        "ticket_weblink": incident.ticket.weblink,
+        "conference_weblink": incident.conference.weblink,
+        "conference_challenge": incident.conference.conference_challenge,
+    }
+
+    faq_doc = document_service.get_incident_faq_document(db_session=db_session)
+    if faq_doc:
+        message_kwargs.update({"faq_weblink": faq_doc.weblink})
+
+    reference_doc = document_service.get_conversation_reference_document(db_session=db_session)
```
mvilanova, line 86: In some functions you're using `faq_doc` and `reference_doc` and in others `incident_faq` and `conversation_reference `. Can you use the same everywhere?

src/dispatch/report/flows.py
mvilanova: You can call the following function to provide feedback to the user: https://github.com/Netflix/dispatch/blob/develop/src/dispatch/conversation/messaging.py#L5

docs/configuration/app.md
kevgliss: They have been removed entirely and are not used in the codebase. They will be defined by their resource type in the document table.

src/dispatch/config.py
kevgliss: Yeah, we can write a migration task.

src/dispatch/config.py
kevgliss: I'm not actually sure these should even be config values? I'm not sure why we would allow users to change them.

mvilanova: Yeah, now that you're mentioning it, I also don't see why they should be.

PR #464, opened 2020-06-25
Fixing Slack properties

sochkasov:
There is not 'users.profile:read' in the Bot Token Scopes.


mvilanova:
Yes, there is. See last one in the screenshot and list of supported token types in https://api.slack.com/scopes/users.profile:read.

Screen Shot 2020-06-25 at 9 09 21 AM

PR #477, opened 2020-07-02
make SQLALCHEMY_DATABASE_URI more generalized

chapmanjacobd:
to support connections on file sockets

looks like this:

```
postgresql+psycopg2://{DATABASE_CREDENTIALS}@/dispatch?host=/file_socket
```


chapmanjacobd:
ah wait this still won't work because ":" will be there even if you don't specify `port` hmmmm....

chapmanjacobd:
can we override config( default() ) with an empty string ?

kevgliss:
I think this might be easiest to support with a new config item and documentation? Defaulting to the current schematics?

chapmanjacobd:
yes. I agree. something more generic will be better and a breaking change
will allow for a cleaner solution

kevgliss:
Cool, I'm going to close this PR feel free to open another.

PR #478, opened 2020-07-06
Bugfix/participant reactivate

kevgliss:


src/dispatch/incident/scheduled.py
```diff
@@ -32,7 +32,13 @@
 
 from .enums import IncidentStatus
 from .flows import update_external_incident_ticket
-from .service import calculate_cost, get_all, get_all_by_status, get_all_last_x_hours_by_status
+from .service import (
+    calculate_cost,
+    get_all,
+    get_all_by_status,
+    get_all_last_x_hours_by_status,
+    get_by_name,
```
mvilanova, line 40: Looks like we added `get_by_name`, but it's actually not being used?

src/dispatch/participant/flows.py
```diff
@@ -135,7 +135,7 @@ def reactivate_participant(user_email: str, incident_id: int, db_session: Sessio
     participant_role = participant_role_service.create(
         db_session=db_session, participant_role_in=participant_role_in
     )
-    participant.participant_role.append(participant_role)
+    participant.participant_roles.append(participant_role)
```
mvilanova, line 138: Nice catch

PR #482, opened 2020-07-06
Prevents slack from throwing a fit if we suggest a lot of documents.

kevgliss:


src/dispatch/incident/messaging.py
```diff
@@ -178,7 +178,9 @@ def send_incident_suggested_reading_messages(
     if suggested_documents:
         # we send the ephemeral message
         items = []
-        for i in suggested_documents:
+        # lets grab the first 5 documents
+        # TODO add more intelligent ranking
+        for i in suggested_documents[:5]:
```
mvilanova, line 183: 5 seems a good number. I'd suggest we define it in a constant.

kevgliss: A constant or a config value? I don't know where else we might use a constant? Also, as I'm not sure what slack message limits might be it's probably safer to not let a user set that value.

mvilanova: Constant, since its value won't likely change.

PR #488, opened 2020-07-08
Adding reminder for incidents that have been stable for over a week.

kevgliss:
Reminds every week on Monday, I chose not to make this configurable at this time.

### DIFF
--- a/src/dispatch/incident/messaging.py
+++ b/src/dispatch/incident/messaging.py
@@ -597,7 +597,7 @@ def send_incident_resources_ephemeral_message_to_participant(
 
 def send_incident_close_reminder(incident: Incident):
     """Sends a direct message to the incident reminding them to close the incident if possible."""
-    message_text = f"Incident Status Reminder"
+    message_text = "Incident Status Reminder"
     message_template = INCIDENT_STATUS_REMINDER
 
     convo_plugin = plugins.get(INCIDENT_PLUGIN_CONVERSATION_SLUG)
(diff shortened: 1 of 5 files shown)


src/dispatch/incident/messaging.py
```diff
@@ -591,3 +593,32 @@ def send_incident_resources_ephemeral_message_to_participant(
...
     log.debug(f"List of incident resources sent to {user_id} via ephemeral message.")
+
+
+def send_incident_close_reminder(incident: Incident):
+    """Sends a direct message to the incident reminding them to close the incident if possible."""
+    message_text = f"Incident Status Reminder"
+    message_template = INCIDENT_STATUS_REMINDER
+
+    convo_plugin = plugins.get(INCIDENT_PLUGIN_CONVERSATION_SLUG)
+    update_command = convo_plugin.get_command_name(ConversationCommands.edit_incident)
```
mvilanova, line 604: nit: we should change the name of the enum to `update_incident` to be consistent with the command name.

src/dispatch/incident/scheduled.py
```diff
@@ -289,3 +288,17 @@ def calculate_incidents_cost(db_session=None):
             # we shouldn't fail to update all incidents when one fails
             log.error(e)
             sentry_sdk.capture_exception(e)
+
+
+@scheduler.add(every(1).day.at("18:00"), name="incident-stable-reminder")
```
mvilanova, line 293: `incident-close-reminder` or `incident-status-reminder`?

src/dispatch/incident/messaging.py
kevgliss: Yeah, I had `update_incident` first until I realized that wasn't the enum.

mvilanova: Yeah, looks like we never updated the enum when we renamed the command from edit to update.

mvilanova: Don't forget to update it here as well: https://github.com/Netflix/dispatch/blob/develop/src/dispatch/conversation/enums.py#L6

PR #491, opened 2020-07-08
Adds Slack command to add events to the incident timeline

mvilanova:
This PR adds a new Slack slash command to add events to the incident timeline.

Screen Shot 2020-07-10 at 12 35 27 PM


src/dispatch/plugins/dispatch_slack/modals.py
```diff
@@ -482,3 +498,132 @@ def update_notifications_group_from_submitted_form(action: dict, db_session=None
...
+        "type": "input",
+        "block_id": AddTimelineEventBlockFields.date,
+        "label": {"type": "plain_text", "text": "Date"},
+        "element": {"type": "datepicker"},
+    }
+    modal_template["blocks"].append(date_picker_block)
+
+    hour_picker_options = []
+    for h in range(0, 24):
+        hour_picker_options.append(create_block_option_from_template(text=h, value=str(h).zfill(2)))
```
kevgliss, line 535: Is zfill used to handle the case when no hour is selected? Can the field not have a default value instead?

src/dispatch/plugins/dispatch_slack/modals.py
```diff
@@ -482,3 +498,132 @@ def update_notifications_group_from_submitted_form(action: dict, db_session=None
...
+    event_minute = parsed_form_data.get(AddTimelineEventBlockFields.minute)["value"]
+    event_description = parsed_form_data.get(AddTimelineEventBlockFields.description)
+
+    incident_id = action["view"]["private_metadata"]["incident_id"]
+
+    participant = participant_service.get_by_incident_id_and_email(
+        db_session=db_session, incident_id=incident_id, email=user_email
+    )
+
+    event_datetime = f"{event_date} {event_hour}:{event_minute}"
```
kevgliss, line 616: nit: this seems a little roundabout you could simply construct the DateTime object directly:

https://www.w3schools.com/python/gloss_python_date_create.asp

src/dispatch/plugins/dispatch_slack/modals.py
mvilanova: No, `zfill` only fills out the value with zeros until it reaches the character length provided. Let me see if I can make the field required, and if not, if I can add a default.

PR #500, opened 2020-07-13
Mark as duplicate

kevgliss:
Adds the ability to mark an incident as a duplicate, and if so which incident is the correct incident.


src/dispatch/static/dispatch/src/incident/IncidentFilterCombobox.vue
```diff
@@ -0,0 +1,125 @@
...
+  props: {
+    value: {
+      type: Array,
+      default: function() {
+        return []
+      }
+    },
+    label: {
+      type: String,
+      defualt: "Add Incidents"
```
mvilanova, line 61:
```suggestion
      default: "Add Incidents"

```

PR #502, opened 2020-07-14
[WIP] Test the incident creation flow.

forestmonster:
What do you think is the best way to handle the assignment of [incident_type_plugin_metadata](https://github.com/Netflix/dispatch/blob/579435f94aac2c1ca60d64514ea189f53e607f33/src/dispatch/incident/flows.py#L109)? I'd rather just have a plain dict assignment. Return a Query object from [incident_type.service.get_by_name()](https://github.com/Netflix/dispatch/blob/9f42c43b63205f4abfde3271b04e85e734f84d09/src/dispatch/incident_type/service.py#L26), then mock get_by_name()? Is there a way to ignore the assignment altogether?

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -626,6 +626,18 @@ class IncidentFactory(BaseFactory):
     title = FuzzyText()
     description = FuzzyText()
     status = FuzzyChoice(["Active", "Stable", "Closed"])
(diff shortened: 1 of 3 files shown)


forestmonster:
It's not picking up recent commits to the branch.

PR #504, opened 2020-07-15
Improves plugin docs

mvilanova:
Improves plugin docs and adds missing doc for Opsgenie.


alexa-kelley-vim:
These are looking a lot clearer. As a quick note for configuring-jira.md and configuring-opsgenie.md you may want to include at the very top on if those plugins are required for core functionality? Otherwise the required field may come across incorrectly. It's a little wordy but something like "required for plugin" or something equally specific. 

An alternative to that idea that I also just thought of is creating a reference table either in the installation doc, or under configuration > app, that lists the bare minimum requirements for standing up the server as far as plugins. That way the what plugin is explicitly required vs not is listed without having mention it in each plugin section later. Hopefully that made sense.

PR #519, opened 2020-07-28
Remove INCIDENT_PLUGIN_STORAGE_SLUG and fix group plugin typo

willbengtson:
Related to https://github.com/Netflix/dispatch/pull/507, fix a typo for `group` plugin type and remove the `INCIDENT_PLUGIN_STORAGE_SLUG` from document scheduler.


kevgliss:
Thanks for this, for anything still using `get` vs `get_active` we also need to use the `plugin.instance` to get the actual plugin class. But I can add those changes in another PR.

PR #522, opened 2020-07-29
Task ownership

kevgliss:


src/dispatch/task/flows.py
```diff
@@ -100,76 +97,41 @@ def send_task_notification(
...
-        incident_task.tickets = tickets
-
         # only notify if it's newly resolved
-        if status == TaskStatus.resolved:
+        prev_status = incident_task.status
+        task = task_service.update(
+            db_session=db_session, task=incident_task, task_in=TaskUpdate(**task)
+        )
+        if prev_status == TaskStatus.resolved:
             if incident_task.status != TaskStatus.resolved:
```
mvilanova, line 111: I'm confused about this. If I'm reading it correctly, we're setting the status of the task to resolved when we go from resolved to open status, assuming `incident_task.status` has the updated status, and then if notify is `True` we send a notification that the task has been resolved? Shouldn't we check if the new status is resolved and the previous one open, and then notify if notify is `True`?

Also, renaming `task` to `updated_incident_task` and using that instead of `incident_task` would make it easier to read.

src/dispatch/task/service.py
```diff
@@ -57,42 +58,86 @@ def get_overdue_tasks(*, db_session) -> List[Optional[Task]]:
...
         tickets=tickets,
-        resource_id=resource_id,
-        resource_type=resource_type,
-        weblink=weblink,
     )
+
+    event_service.log(
+        db_session=db_session,
+        source="Dispatch Core App",
+        description="New task created",
```
mvilanova, line 105:
```suggestion
        description="New incident task created",
```

src/dispatch/task/service.py
```diff
@@ -57,42 +58,86 @@ def get_overdue_tasks(*, db_session) -> List[Optional[Task]]:
...
+        incident_id=incident.id,
+    )
```
mvilanova, line 107: You could use the `details` column in the event model to add the description of the task or any other details, which then could be displayed in the timeline view. For example:
```
details={"description": description},
```

src/dispatch/decorators.py
```diff
@@ -45,7 +48,8 @@ def wrapper(*args, **kwargs):
             print(traceback.format_exc())
             sentry_sdk.capture_exception(e)
         finally:
-            db_session.close()
+            if background:
+                kwargs["db_session"].close()
```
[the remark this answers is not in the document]
kevgliss: There are two flows, as a web request (inline) that has an established db_session for the web request and and background task (spawns a new thread and does not have a db_session). If we don't check to see we are calling it in the background (by checking that db_session doesn't exist) we run into issues where the session gets closed by this function, even though functions call it may still need to use the session.

src/dispatch/individual/models.py
```diff
@@ -57,7 +57,7 @@ class IndividualContact(ContactMixin, Base):
         secondary=assoc_individual_contact_incident_priorities,
         backref="individuals",
     )
-    participant = relationship("Participant", lazy="subquery", backref="individual")
+    # participant = relationship("Participant", lazy="subquery", backref="individual")
```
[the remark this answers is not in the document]
kevgliss: We can, I wanted to to the whole suite of testing to make sure it didn't cause any issues before totally removing.

src/dispatch/task/service.py
kevgliss: That feels like duplicating data on the event timeline (the description could be quite big). If we felt that data was useful for the timeline I would rather merge it instead of duplicating actual data (what happens if a task changes?), at most I would add a link to the task in question here.

mvilanova: A link to the task would be great. I wasn't sure we could generate them w/o adding more code.

src/dispatch/task/flows.py
kevgliss: afaik, I don't believe the logic was changed. I added a few comments to hopefully make the logic a bit clearer.

src/dispatch/task/service.py
kevgliss: Added.

src/dispatch/task/flows.py
```diff
@@ -100,76 +97,40 @@ def send_task_notification(
...
+        task = task_service.update(
+            db_session=db_session, task=existing_task, task_in=TaskUpdate(**task)
+        )
 
-                if notify:
+        if notify:
+            # determine if task was perviously resolved
+            if existing_status == TaskStatus.resolved:
+                # determine if we have a newly resolved task (since last sync)
+                if existing_task.status != TaskStatus.resolved:
```
mvilanova, line 114: I still don't think this is going to work as intended. If a task goes from open (`existing_status == TaskStatus.open`) to resolved (`existing_task.status == TaskStatus.resolved`), you're not notifying. You're notifying when a task goes from resolved to open. Previously, we [checked](https://github.com/Netflix/dispatch/blob/develop/src/dispatch/task/flows.py#L132) if the new status was resolved and then if the previous one was open (or not resolved).

src/dispatch/task/service.py
mvilanova: Hmm where did you add it? I don't see it.

src/dispatch/task/flows.py
kevgliss: I've inverted the statements.

PR #524, opened 2020-07-29
Add AioHttpIntegration and SqlalchemyIntegration to Sentry

willbengtson:
Add AioHttpIntegration and SqlalchemyIntegration to Sentry

### DIFF
--- a/src/dispatch/extensions.py
+++ b/src/dispatch/extensions.py
@@ -1,7 +1,11 @@
 import logging
 
 import sentry_sdk
-from sentry_sdk.integrations.logging import LoggingIntegration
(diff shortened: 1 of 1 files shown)


willbengtson:
https://docs.sentry.io/platforms/python/sqlalchemy/

https://docs.sentry.io/platforms/python/aiohttp/

kevgliss:
Do we need to add `aiocontextvars` to the requirements? The aiohttp integration docs mention it.

kevgliss:
Ah nvm, looks like it's only 3.6 since we target 3.7 and above we should be good. Thanks!

PR #530, opened 2020-08-06
Update slash commands in public docs

forestmonster:
Also there were a number of redundant screenshots, not sure exactly why. Removed those and fixed links.


docs/user-guide/incident-commander.md
```diff
@@ -6,61 +6,104 @@ description: What to expect as an incident commander.
 
 ## Reporting
 
-Within Dispatch Incident Commander's \(ICs\) are also participants and will receive all of the messaging that a participant does. When resolved as incident commander you are assigned that role by Dispatch and your identity is propagated via various messaging. Additionally for `High` severity events, the Incident Commander is automatically paged \(via PagerDuty\).
+Within Dispatch, Incident Commanders \(ICs\) are also participants, and will receive all of the participant messaging as well. When resolved as the Incident Commander, you are assigned that role by Dispatch, and your identity is propagated via various messaging. For `High` priority incidents, the Incident Commander is automatically paged \(via PagerDuty\).
```
mvilanova, line 9: IC is only paged via the oncall service defined if the `Page Commander` checkbox is checked for the chosen incident priority when the incident is reported.
```suggestion
Within Dispatch, Incident Commanders \(ICs\) are also participants, and will receive all of the participant messaging as well. When resolved as the Incident Commander, you are assigned that role by Dispatch, and your identity is propagated via various messaging.
```

docs/user-guide/incident-commander.md
```diff
@@ -6,61 +6,104 @@ description: What to expect as an incident commander.
...
 
-## During
+All Slack commands are listed below, or you may view _groups_ of commands relating to [People](#people), [Communications](#communications), [Tasks](#tasks), and [Incident Resources & Metadata](#incident-resources-and-metadata).
```
mvilanova, line 11: Should we point people to the list of commands [here](https://hawkins.gitbook.io/dispatch/configuration/plugins/configuring-slack#commands), which are up to date,  rather than listing them again here, which doubles the amount of maintenance?

forestmonster: Yeah I'm a fan of DRY so that's a great idea. Erm... and... to do that we should recognize that these serve two different purposes. One seeks to educate incident commanders on how and why to use the commands in Slack, and the other helps people when configuring the app. Seems like we could either merge these, or double down on a separate section to help people use the software after configuration. Do you anticipate a lot of further changes to our commands? That could change how we approach it.

forestmonster: Yeah the User Guide is definitely a different audience from those configuring the app. Thinking that a little duplication may just be a necessary evil in this case, to help users. I hope to anticipate and head off support questions, you know?

mvilanova: Sounds good.

docs/user-guide/incident-commander.md
```diff
@@ -6,61 +6,104 @@ description: What to expect as an incident commander.
...
 
-You're not always incident commander forever, sometimes a situation changes or you just need a break. The `dispatch-assign-role` command makes it easy and more importantly _clear_ to everyone involved in the incident who the current incident commander is.
+It's especially important to use this command when handing off responsibility for incident leadership. Doing so will help avoid any confusion about the identity of the current Incident Commander.
 
-![](../.gitbook/assets/slack-conversation-assign-role%20%281%29.png)
+![](../.gitbook/assets/slack-conversation-assign-role.png)
 
-### Status
+### /dispatch-engage-oncall
+You'll need the help of various teams to resolve an incident. To quickly page an on-call member of another team, use `/dispatch-engage-oncall` to resolve their identity, and optionally page them using PagerDuty.
```
mvilanova, line 39:
```suggestion
You'll need the help of various teams to resolve an incident. To quickly engage an on-call member of another team, use `/dispatch-engage-oncall` to resolve their identity, and optionally page them.
```

mvilanova: Removing PagerDuty as we support other oncall services.

docs/user-guide/incident-commander.md
```diff
@@ -6,61 +6,104 @@ description: What to expect as an incident commander.
...
+### /dispatch-update-participant
+Participants in an incident, or the Incident Commander, may want to know a participant's area of expertise or their expected contribution to resolving an incident. Use `/dispatch-update-participant` to update the reason a participant was added. The dialog appears like this:
+
+![](../.gitbook/assets/slack-conversation-update-participant.png)
+
+## Communications
+These commands help manage incident communications.
+
+### /dispatch-notifications-group
+An incident notifications group consists of individuals, or a distribution list. Manage this group by using `/dispatch-notifications-group`.
```
mvilanova, line 59:
```suggestion
An incident notifications group consists of individuals, or distribution lists. Manage this group by using `/dispatch-notifications-group`.
```

docs/user-guide/incident-commander.md
```diff
@@ -6,61 +6,104 @@ description: What to expect as an incident commander.
...
 
-This command marks the incident as stable, updating external tickets as necessary.
+![](../.gitbook/assets/google-docs-task-comment.png)
 
-It also creates a Post Incident Review \(PIR\) document to help you track the post incident review process.
+The following commands help manage these tasks associated with an incident.
 
-`/dispatch-mark-closed`
+### /dispatch-list-my-tasks
+Any individual who issues the `/dispatch-list-my-tasks` command will see a list of only their assigned tasks.
```
mvilanova, line 94:
```suggestion
Any individual who issues the `/dispatch-list-my-tasks` command will see a list of tasks created by or assigned to them.
```

PR #532, opened 2020-08-15
Fix Doc Typos

chickenandpork:
While reading docs, fixed a few typos/syntax and a bad link.

Not worth a new release.  I'm sorry if I come off as a "Linter" (no impact beyond spellcheck) but it stands out to me and this may be what little I can contribute :)


mvilanova:
Thanks for the fixes!

PR #553, opened 2020-09-09
Feature/sentry tags

kevgliss:
Allows you to specify env tags that will be gathered and sent to sentry. An example:

`ENV_TAGS="foo:bar"`

Will attempt to pull `bar` from the current ENV and set it's value to the tag `foo`

### DIFF
--- a/docs/configuration/app.md
+++ b/docs/configuration/app.md
@@ -37,7 +37,7 @@ In general, do not include any quotation marks when adding configuration values.
 
 #### `ENV_TAGS` \[defaut: ""\]
 
-> A comma seperated list of tags that Dispatch will attempt to pull from the environment. As an example the string `foo:bar,baz:blah` will create two tags, `foo` with the environment value for `bar` and `baz with the environments values for `blah`.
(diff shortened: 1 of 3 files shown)


docs/configuration/app.md
```diff
@@ -35,7 +35,11 @@ In general, do not include any quotation marks when adding configuration values.
 
 > A comma separated list of metric providers Dispatch will send key system metrics to.
 
-#### `SENTRY_DSN` \[default: none\] \[secret: True\]
+#### `ENV_TAGS` \[defaut: ""\]
+
+> A comma seperated list of tags that Dispatch will attempt to pull from the environment. As an example the string `foo:bar,baz:blah` will create two tags, `foo` with the environment value for `bar` and `baz with the environments values for `blah`.
```
mvilanova, line 40:
```suggestion
> A comma separated list of tags that Dispatch will attempt to pull from the environment. As an example the string `foo:bar,baz:blah` will create two tags: `foo` with the environment value for `bar` and `baz` with the environment value for `blah`.
```

mvilanova (approved):

PR #569, opened 2020-09-16
Adding the ability to run external flows

kevgliss:
Adds:

- A new command `/dispatch-run-external-workflow`
- A new modal (with dynamic workflow parameters)
- New plugin type `external-workflow`


src/dispatch/plugins/bases/external_flow.py
```diff
@@ -0,0 +1,20 @@
...
+    :platform: Unix
+    :copyright: (c) 2019 by Netflix Inc., see AUTHORS for more
+    :license: Apache, see LICENSE for more details.
+.. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
+"""
+from dispatch.plugins.base import Plugin
+from dispatch.models import PluginOptionModel
+
+
+class ExternalFlowPlugin(Plugin):
```
mvilanova, line 12: Should we just call them Workflows?

src/dispatch/plugins/dispatch_slack/dialogs.py
```diff
@@ -19,6 +21,47 @@
 log = logging.getLogger(__name__)
 
 
+@background_task
+def create_run_external_flow_dialog(incident_id: int, command: dict = None, db_session=None):
```
mvilanova, line 25: Why not using a modal?

kevgliss: Will switch.

PR #581, opened 2020-09-23
Slack list incidents command

mvilanova:
This PR adds a new Slack slash command to list current active and stable incidents, and closed incidents in the last 24 hours.
Screen Shot 2020-09-23 at 11 12 06 AM


kevgliss:
Can we modify the query to exclude simulations?

Issue #594, opened 2020-10-05
Can't properly use default set of plugins


nedvna:
**Describe the bug**
After I set up the basic minimalistic install of Dispatch, I can't actually use it.
All the plug ins are enabled by default.
When you try to create the incident (after creating a document, service, service type, priority and an individual), the application fails to do so. You can't see your incidents in the incident list (/incidents).

**To Reproduce**
1. Set up basic .env - change only DISPATCH_UI_URL, DISPATCH_HELP_EMAIL. SECRET_KEY and DISPATCH_JWT_SECRET will be replaced automatically. Other fields are left 'as is', AFAIK.
2. Run ./install.sh
2. docker-compose up -d
3. Register (and auto-login) with basic auth
4. Create the incident (and all its prerequisites before)
5. Observe the trace in logs and not-working GUI

**Expected behavior**
Incident is created successfully, and can be viewed in the incidents list.
Or maybe an error message stating 'you can not use multiple plugins of type X!' of some kind.

**Additional context**
What we have here is that Dispatch doesn't quite work 'out of the box'. It took quite some time to find a workaround.
Obviously, the described setup is a 'bare bones' one, only useful for demonstrational purposes / some basic testing. But, well, that was what I was trying to achieve before I try to set up plugins etc. Unfortunately, that doesn't work. According to database data, the incident is created but you can't view it.
What worked was - disabling all plugins before creating anything in the UI, except for dispatch-* ones.
Also, even though it worked, the UI is kind of unresponsive. I mean, it looks like the incident is being created forever. I have to switch to another tab from the side pane, and the back to 'Incidents'. Then I can see and access my incident.
Can't create tasks though, but that should be another bug report.
(edited) If after the failed creation, I disable the plugins, restart docker-compose and then try to create another incident, I still can't view the list. I guess some info gets halfway-written there, and fetching the incidents from DB is broken.

Trace examples:
[trace_examples.log](https://github.com/Netflix/dispatch/files/5327772/trace_examples.log)

nedvna:
> I will take a look, did you happen to install the example data? Or was this just based on an empty database?

It was an empty database, I only added my data.
I'll try to populate with data from [here](https://raw.githubusercontent.com/Netflix/dispatch/master/data/dispatch-sample-data.dump) a bit later but I doubt it's going to help, might have also done that in the past.

nedvna:
So I dropped the database and loaded data from dump.
[...]

PR #597, opened 2020-10-06
Better emtpy database support. 

### LINKED ISSUE #595: Can't create task in UI, code 422
**Describe the bug**
After creating the minimalistic install of Dispatch (see #594 ), I can create an incident using a workaround but still can't create a task.

**To Reproduce**
1. Create the minimalistic install with basic auth, disable most plugins
2. Create an incident
3. Try to create a task (fill in all fields, bind to that incident)

**Expected behavior**
A task is created and listed

**Additional context**
After clicking on the 'save' icon what I'm getting is
`INFO:     172.17.41.116:48794 - "POST /api/v1/tasks/ HTTP/1.1" 422 Unprocessable Entity`

kevgliss:
Small quality of life improvements, including better support for an empty database starting point.

Should fix #595 and hopefully address some of the issues in #594

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,14 +36,14 @@ RUN for key in \
     A48C2BEE680E841632CD4E44F07496B3EB3C1762 \
     B9E2F5981AA6E0CD28160D9FF13993A75599653C \
     ; do \
-        # Let's try several servers to ensure we get all the keys
-        for server in \
-            keys.openpgp.org \
-            keyserver.ubuntu.com \
-            hkps.pool.sks-keyservers.net \
-        ; do \
-	        gpg2 --batch --keyserver "$server" --recv-keys "$key"; \
-        done \
+    # Let's try several servers to ensure we get all the keys
+    for server in \
+    keys.openpgp.org \
+    keyserver.ubuntu.com \
+    hkps.pool.sks-keyservers.net \
+    ; do \
+    gpg2 --batch --keyserver "$server" --recv-keys "$key"; \
+    done \
     done
 
 # grab gosu for easy step-down from root
@@ -124,7 +124,7 @@ RUN set -x \
     && apt-get update \
     && apt-get install -y --no-install-recommends $buildDeps \
     # remove internal index when internal plugins are seperated
-    && pip install /tmp/dist/*.whl \
+    && pip install -U /tmp/dist/*.whl \
     && apt-get purge -y --auto-remove $buildDeps \
     # We install run-time dependencies strictly after
     # build dependencies to prevent accidental collusion.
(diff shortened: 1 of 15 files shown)


docker/Dockerfile
```diff
@@ -36,14 +36,14 @@ RUN for key in \
...
     B9E2F5981AA6E0CD28160D9FF13993A75599653C \
     ; do \
-        # Let's try several servers to ensure we get all the keys
-        for server in \
-            keys.openpgp.org \
-            keyserver.ubuntu.com \
-            hkps.pool.sks-keyservers.net \
-        ; do \
-	        gpg2 --batch --keyserver "$server" --recv-keys "$key"; \
-        done \
```
mvilanova, line 46: I'd prefer if we leave this nested `for` code block indented, as it makes it easier to read/ troubleshoot.

src/dispatch/auth/service.py
```diff
@@ -79,6 +79,11 @@ def get_current_user(*, db_session: Session = Depends(get_db), request: Request)
         log.debug("No authentication provider. Default user will be used")
         user_email = DISPATCH_AUTHENTICATION_DEFAULT_USER
 
+    if not user_email:
+        log.exception(
+            f"Unable to determine user email based on configured auth provider. Provider: {DISPATCH_AUTHENTICATION_PROVIDER_SLUG}"
```
mvilanova, line 84: Adding verbiage in case we tried to get the user email from the `DISPATCH_AUTHENTICATION_DEFAULT_USER` env var and this was not defined.
```suggestion
            f"Unable to determine user email based on configured auth provider or no default auth user email defined. Provider: {DISPATCH_AUTHENTICATION_PROVIDER_SLUG}"
```

src/dispatch/cli.py
```diff
@@ -52,34 +52,55 @@ def list_plugins():
     for p in plugins.all():
         record = plugin_service.get_by_slug(db_session=db_session, slug=p.slug)
 
-        installed = True
         if not record:
-            installed = False
-
-        table.append([p.title, p.slug, p.version, installed, p.type, p.author, p.description])
+            log.warning(
+                f"{p.name} Plugin available, but not installed. Run `dispatch plugins install` to install it."
```
mvilanova, line 57:
```suggestion
                f"Plugin {p.name} available, but not installed. Run `dispatch plugins install` to install it."
```

src/dispatch/metrics.py
```diff
@@ -12,7 +12,7 @@ class Metrics(object):
 
     def __init__(self):
         if not METRIC_PROVIDERS:
-            log.warning("No metric providers specified metrics will not be sent.")
+            log.info("No metric providers specified metrics will not be sent.")
```
mvilanova, line 15:
```suggestion
            log.info("No metric providers defined via METRIC_PROVIDERS env var. Metrics will not be sent.")
```

docker/Dockerfile
kevgliss: This was auto formatted for me, with what I assume is the defaults.

mvilanova: I suspect the autoformatter you're using is not able to detect it's a nested for.

kevgliss: Yes that is likely the case (or it's preference is to not nest them), I guess my point is that I wouldn't want to give up formatting for the whole file (if I can't disable this one behavior).

PR #604, opened 2020-10-08
Adds support for tag types

mvilanova:
This PR adds support for creating types of tags. Now you will be able to select from the list of tag types available when you create a new tag. For example, you can create a new tag with name `foo@example.com` and type `Email Address` or name `127.0.0.1` and type `IP Address`. In the future, we may allow to map tag types to actions or workflows that we can suggest (or programmatically run) when an incident is tagged.

Screen Shot 2020-10-09 at 3 05 46 PM

Screen Shot 2020-10-09 at 3 05 37 PM


src/dispatch/static/dispatch/src/api/menu.js
```diff
@@ -110,18 +110,25 @@ const Menu = [
...
   {
-    title: "Workflows",
-    group: "workflows",
-    component: "Workflows",
-    icon: "work",
-    href: "/workflows"
+    title: "Tag Types",
+    group: "Configuration",
+    name: "Tag Types",
+    icon: "settings",
```
kevgliss, line 116: Shouldn't we pick a different icon?

kevgliss (approved):

src/dispatch/static/dispatch/src/api/menu.js
mvilanova: I'll change it to label, so we're consistent with the other tag section above.

PR #605, opened 2020-10-09
Refactor the route/service module

tomerm98:


kevgliss:
Hey thanks for this PR, I think it looks cleaner that what I had. We'll test this out and merge it in.

PR #620, opened 2020-10-14
Stops an error from being logged for duplicate google groups

kevgliss:


src/dispatch/plugins/dispatch_google/groups/plugin.py
```diff
@@ -78,6 +78,9 @@ def add_member(client: Any, group_key: str, email: str, role: str):
         except HttpError as e:
             #  we are okay with duplication errors upon insert
             if e.resp.status in [409]:
+                log.debug(
+                    f"Member already exists in google group. GroupKey={group_key} Body={body}"
```
mvilanova, line 82: Can we use the error string we get from Google instead?

kevgliss: We could, but dumping the json into the logs make it less readable imo.

mvilanova: I guess that's fine as long as the response status 409 is only used for this type of error.

kevgliss: It looks like it's pretty standard: https://developers.google.com/webmaster-tools/search-console-api-original/v3/errors#:~:text=CONFLICT%20(409),-Error%20code&text=The%20API%20request%20cannot%20be,identified%20with%20more%20specific%20errors.

PR #623, opened 2020-10-14
Allows for incidents to be correctly re-opened

kevgliss:


src/dispatch/incident/flows.py
```diff
@@ -630,25 +630,22 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
 
-@background_task
-def incident_active_flow(incident_id: int, command: Optional[dict] = None, db_session=None):
+def incident_active_status_flow(incident: Incident, db_session=None):
     """Runs the incident active flow."""
-    # we load the incident instance
-    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
-
-    # we remind the incident commander to write a tactical report
-    send_incident_report_reminder(incident, ReportTypes.tactical_report, db_session)
```
mvilanova, line 640: I think we should keep the reminder, as we want the IC to inform everyone about why it was re-opened?

src/dispatch/incident/flows.py
```diff
@@ -630,25 +630,22 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
+    # we un-archive the conversation
+    convo_plugin = plugin_service.get_active(db_session=db_session, plugin_type="conversation")
+    if convo_plugin:
+        convo_plugin.instance.unarchive(incident.conversation.channel_id)
```
mvilanova, line 638: Should we only do this when we go from closed -> active? Slack will throw a `not_archived` error if the channel is not archived: https://api.slack.com/methods/conversations.unarchive#errors. Or is the plan to ignore the error?

src/dispatch/incident/flows.py
```diff
@@ -630,25 +630,22 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
 
-    # we update the status of the external ticket
+    # we update the external ticket
     update_external_incident_ticket(incident, db_session)
 
+    db_session.add(incident)
+    db_session.commit()
```
mvilanova, line 644: I don't think this is necessary as we're not updating the incident, unless I'm missing something?

src/dispatch/incident/flows.py
```diff
@@ -772,16 +765,44 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
...
+):
+    """Runs the correct flows depending on a incident's current and previous status."""
+
+    # we have a currently active incident
+    if current_status == IncidentStatus.active:
+        # re-activate incident
+        if previous_status == IncidentStatus.closed:
+            incident_active_status_flow(incident=incident, db_session=db_session)
+        elif previous_status == IncidentStatus.stable:
+            incident_active_status_flow(incident=incident, db_session=db_session)
```
mvilanova, line 786: Do we need the `if / elif` here if we always call `incident_active_status_flow()`?
```suggestion
        incident_active_status_flow(incident=incident, db_session=db_session)
```

src/dispatch/incident/flows.py
```diff
@@ -772,16 +765,44 @@ def incident_closed_flow(incident_id: int, command: Optional[dict] = None, db_se
...
+            incident_stable_status_flow(incident=incident, db_session=db_session)
+
+    # we currently have a closed incident
+    elif current_status == IncidentStatus.closed:
+        if previous_status == IncidentStatus.active:
+            incident_stable_status_flow(incident=incident, db_session=db_session)
+            incident_closed_status_flow(incident=incident, db_session=db_session)
+
+        if previous_status == IncidentStatus.stable:
```
mvilanova, line 802:
```suggestion
        elif previous_status == IncidentStatus.stable:
```

src/dispatch/incident/flows.py
```diff
@@ -876,15 +897,8 @@ def incident_update_flow(
...
-        if incident.status == IncidentStatus.active:
-            incident_active_flow(incident_id=incident.id, db_session=db_session)
-        elif incident.status == IncidentStatus.stable:
-            incident_stable_flow(incident_id=incident.id, db_session=db_session)
-        elif incident.status == IncidentStatus.closed:
-            if previous_incident.status.value == IncidentStatus.active:
-                incident_stable_flow(incident_id=incident.id, db_session=db_session)
-            incident_closed_flow(incident_id=incident.id, db_session=db_session)
+    # run whatever flows we need
+    status_flow_dispatcher(incident, incident.status, previous_incident.status.value)
```
mvilanova, line 901: Missing db_session

src/dispatch/incident/flows.py
kevgliss: In the case of going from "closed" -> "stable" we run the active and then stable flows. Since stable is already sending a reminder, I wanted to avoid duplicate reminders.

If we want this reminder for all status changes except "closed" I can move it out of a specific flow.

src/dispatch/incident/flows.py
kevgliss: I think we can handle this error in the slack plugin, i.e. if we ask for it to be unarchived and it's already unarchived we should swallow that exception.

src/dispatch/incident/flows.py
mvilanova: Yeah, I think it makes sense to send the reminder for both active and stable.

src/dispatch/incident/flows.py
```diff
@@ -630,140 +631,127 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
 
-        # incident review document is optional
-        if template:
-            incident_review_document = storage_plugin.instance.copy_file(
-                folder_id=incident.storage.resource_id,
-                file_id=template.resource_id,
-                name=incident_review_document_name,
-            )
+    # incident review document is optional
+    if template:
```
mvilanova, line 665: We may want to check if the template exists, and if not, let the user know using a warning.

src/dispatch/incident/flows.py
```diff
@@ -850,41 +825,88 @@ def incident_update_flow(
...
 
-    # we update the external ticket
-    update_external_incident_ticket(incident, db_session)
+def status_flow_dispatcher(
+    incident: Incident,
+    current_status: IncidentStatus,
+    previous_status: IncidentStatus,
+    db_session=SessionLocal,
+):
+    """Runs the correct flows depending on a incident's current and previous status."""
```
mvilanova, line 835:
```suggestion
    """Runs the correct flows depending on the incident's current and previous status."""
```

src/dispatch/incident/flows.py
```diff
@@ -850,41 +825,88 @@ def incident_update_flow(
...
+    # we have a currently active incident
+    if current_status == IncidentStatus.active:
+        # re-activate incident
+        incident_active_status_flow(incident=incident, db_session=db_session)
+        send_incident_report_reminder(incident, ReportTypes.tactical_report, db_session)
```
mvilanova, line 840:
```suggestion
        send_incident_report_reminder(incident, ReportTypes.tactical_report.value, db_session)
```

mvilanova: Same for the other calls below.

mvilanova:
Looks great. Just a few minor, non-blocking things.

PR #624, opened 2020-10-15
Updates Python requirements

mvilanova:
Installs sentry sdk with flask to solve the issue described [here](https://github.com/getsentry/sentry-docs/issues/371). Also, removes a duplicate.


kevgliss:
But... we don't use the flask?

mvilanova:
I can try with only adding `blinker`, but if they add other dependencies we could have the same problem.

PR #631, opened 2020-10-16
Asks the commander to review the incident's information on incident close

mvilanova:
This PR adds a reminder for the incident commander to review the incident's information and to tag the incident when the incident is closed.

Screen Shot 2020-10-16 at 1 24 59 PM

### DIFF
--- a/src/dispatch/messaging.py
+++ b/src/dispatch/messaging.py
@@ -503,7 +503,7 @@ INCIDENT_COMMANDER_READDED_NOTIFICATION = [
 INCIDENT_CLOSED_INFORMATION_REVIEW_REMINDER_NOTIFICATION = [
     {
         "title": "{{name}} Incident - Information Review Reminder",
-        "title_link": "{{ticket_weblink}}",
+        "title_link": "{{dispatch_ui_url}}",
         "text": INCIDENT_CLOSED_INFORMATION_REVIEW_REMINDER_DESCRIPTION,
     }
 ]
--- a/src/dispatch/messaging.py
+++ b/src/dispatch/messaging.py
@@ -20,17 +20,18 @@ from .config import (
 
 
 class MessageType(str, Enum):
+    incident_closed_information_review_reminder = "incident-closed-information-review-reminder"
     incident_daily_summary = "incident-daily-summary"
     incident_daily_summary_no_incidents = "incident-daily-summary-no-incidents"
     incident_executive_report = "incident-executive-report"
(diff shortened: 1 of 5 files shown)


src/dispatch/messaging.py
```diff
@@ -252,6 +253,9 @@ class MessageType(str, Enum):
 {% endif %}
 """
 
+INCIDENT_CLOSED_INFORMATION_REVIEW_REMINDER_DESCRIPTION = """
+Thanks for closing incident {{name}}. Is the following incident information up to date? If not, please edit the incident in the <{{dispatch_ui_url}}|Dispatch Web UI>. Also, consider tagging the incident using the Web UI if appropiate.\n\n*Title:* {{title}}\n*Description:* {{description}}\n*Incident Type:* {{type}}\n*Incident Priority:* {{priority}}"""
```
forestmonster, line 257:
```suggestion
Thanks for closing incident {{name}}. Is the following incident information up to date? If not, please edit the incident in the <{{dispatch_ui_url}}|Dispatch Web UI>. If it's appropriate, consider adding relevant tags to the incident using the Web UI. This will help us solve similar incidents faster.\n\n*Title:* {{title}}\n*Description:* {{description}}\n*Incident Type:* {{type}}\n*Incident Priority:* {{priority}}"""
```

forestmonster: I'm not sure quite how to say it, but I think we want folks to know _why_ they should add tags to something, even if they're not familiar with the concept of "tagging."

forestmonster (approved):
Feedback inline!

src/dispatch/messaging.py
```diff
@@ -496,6 +500,14 @@ class MessageType(str, Enum):
     {"title": "Incident Commander Re-Added", "text": INCIDENT_COMMANDER_READDED_DESCRIPTION}
 ]
 
+INCIDENT_CLOSED_INFORMATION_REVIEW_REMINDER_NOTIFICATION = [
+    {
+        "title": "{{name}} Incident - Information Review Reminder",
+        "title_link": "{{ticket_weblink}}",
```
kevgliss, line 506: This link would go to Jira afaik. Is that what we want? Or do we always want to take them to the Dispatch UI (since we're asking them to modify it there).

mvilanova: I'll change it to the Dispatch UI and work on adding the ability to link to tickets directly in Dispatch.

PR #634, opened 2020-10-19
Adds support for capturing feedback from participants on incident close

mvilanova:
This PR adds a new data model for feedback and logic to message incident participants on incident close to request them to rate and provide feedback about the incident.

Screen Shot 2020-10-19 at 2 28 52 PM
Screen Shot 2020-10-19 at 2 29 09 PM


kevgliss:
This looks good, do you think we should provide an option (a checkbox) to give anonymous feedback?

mvilanova:
@kevgliss yeah, I think it's a good idea. We may receive more candid feedback if we provide the option.

src/dispatch/api.py
```diff
@@ -61,6 +62,7 @@
 )
 authenticated_api_router.include_router(workflow_router, prefix="/workflows", tags=["workflows"])
 authenticated_api_router.include_router(plugin_router, prefix="/plugins", tags=["plugins"])
+authenticated_api_router.include_router(feedback_router, prefix="/feedbacks", tags=["feedbacks"])
```
forestmonster, line 65: FYI, "feedback" is the correct plural form...

src/dispatch/feedback/views.py
```diff
@@ -0,0 +1,97 @@
...
+    FeedbackUpdate,
+)
+from .service import create, delete, get, update
+
+
+router = APIRouter()
+
+
+@router.get("/", response_model=FeedbackPagination)
+def get_feedbacks(
```
forestmonster, line 21: I see, you're using the method name to remind you that you're getting multiple feedback objects, not a single piece of feedback. Carry on then.

forestmonster (approved):

PR #635, opened 2020-10-20
Adds error handling at the decorator level for actions.

kevgliss:
If we like this approach we can do something similar with /commands and /events so we can better reporting to the user when an error occurs and next steps.


src/dispatch/plugins/dispatch_slack/decorators.py
```diff
@@ -0,0 +1,64 @@
...
+            elapsed_time = time.perf_counter() - start
+            metrics_provider.timer(
+                "function.elapsed.time",
+                value=elapsed_time,
+                tags={"function": fullname(func), "slack": True},
+            )
+            return result
+        except Exception as e:
+            # we generate our own guid for now, maybe slack provides us something we can use?
+            slack_interaction_guid = str(uuid.uuid4())
```
mvilanova, line 51: Perhaps the `trigger_id`?

mvilanova (approved):

PR #638, opened 2020-10-21
Adds new section for incident feedback in the Web UI

mvilanova:
This PR adds a new section in the Web UI to visualize incident feedback.

Screen Shot 2020-10-22 at 2 51 48 PM


src/dispatch/feedback/models.py
```diff
@@ -19,8 +22,11 @@ class Feedback(Base, TimeStampMixin):
 
 # Pydantic models
 class FeedbackBase(DispatchBase):
+    created_at: Optional[datetime]
     rating: str
     feedback: Optional[str]
+    incident: Optional[IncidentReadNested]
```
kevgliss, line 28: Do we expect to have feedback not tied to an incident?

kevgliss (approved):
I might have missed this in another PR, but do we hide/filter feedback from sensitive incidents?

src/dispatch/feedback/models.py
mvilanova: No, the incident provides the context to help us understand the feedback.

kevgliss: So why is it optional?

mvilanova: Must be a leftover from when I was testing adding feedback through the UI, which I will tackle in a separate PR.

mvilanova:
No, I was going to look into that when I was going to add additional column filters. We could also filter feedback out based on incident participation for the user role (i.e. you only see feedback about incidents you have participated).

PR #642, opened 2020-10-22
Ensures that incident type documents are truly optional. 

kevgliss:


src/dispatch/incident/service.py
```diff
@@ -35,23 +35,33 @@ def resolve_incident_commander_email(
...
+        if page_commander:
+            p.instance.page(
+                service_id=commander_service.external_id,
+                incident_name=incident_name,
+                incident_title=incident_title,
+                incident_description=incident_description,
+            )
 
-    return p.instance.get(service_id=commander_service.external_id)
+        return p.instance.get(service_id=commander_service.external_id)
```
mvilanova, line 61: Why creating a new plugin instance if you have `oncall_plugin`?

src/dispatch/incident/service.py
```diff
@@ -35,23 +35,33 @@ def resolve_incident_commander_email(
...
-            incident_description=incident_description,
-        )
+    # We resolve the incident commander email
+    # default to reporter if we don't have an oncall plugin enabled
+    oncall_plugin = plugin_service.get_active(db_session=db_session, plugin_type="oncall")
+    if oncall_plugin:
+        incident_type = incident_type_service.get_by_name(db_session=db_session, name=incident_type)
+
+        if not incident_type.commander_service:
+            return reporter_email
```
mvilanova, line 45: Why not keeping this before the `if`, if we don't rely on `oncall_plugin` for it?

src/dispatch/incident/service.py
```diff
@@ -35,23 +35,33 @@ def resolve_incident_commander_email(
...
-        p.instance.page(
-            service_id=commander_service.external_id,
-            incident_name=incident_name,
-            incident_title=incident_title,
-            incident_description=incident_description,
-        )
+    # We resolve the incident commander email
+    # default to reporter if we don't have an oncall plugin enabled
+    oncall_plugin = plugin_service.get_active(db_session=db_session, plugin_type="oncall")
+    if oncall_plugin:
```
mvilanova, line 41: I think it will be easier to read if you check if we don't have the `oncall_plugin`
```suggestion
    if not oncall_plugin:
        return reporter_email
```

src/dispatch/incident/service.py
```diff
@@ -35,23 +35,33 @@ def resolve_incident_commander_email(
...
+        incident_type = incident_type_service.get_by_name(db_session=db_session, name=incident_type)
+
+        if not incident_type.commander_service:
+            return reporter_email
+
+        commander_service = incident_type.commander_service
+
+        p = plugin_service.get_active(db_session=db_session, plugin_type="oncall")
```
mvilanova, line 49: Why creating a new plugin instance if you have `oncall_plugin`?

PR #644, opened 2020-10-26
Fixes issue with possible XSS

kevgliss:


mvilanova:
Looks good. I scanned the static directory and did not find more instances.

PR #645, opened 2020-10-26
Fixes an issue were users can self-assign roles

kevgliss:


src/dispatch/auth/views.py
```diff
@@ -65,17 +67,26 @@ def get_user(*, db_session: Session = Depends(get_db), user_id: int):
...
+        if not user:
+            raise HTTPException(status_code=404, detail="The user with this id does not exist.")
 
-    user = update(db_session=db_session, user=user, user_in=user_in)
+        user = update(db_session=db_session, user=user, user_in=user_in)
 
-    return user
+        return user
+
+    raise HTTPException(status_code=401, detail="You do no have permission to view this incident.")
```
mvilanova, line 89: We should update the text in `detail` to reflect the operation that the user is trying to perform here.

kevgliss: Missed this. Will resolve in another PR.

PR #646, opened 2020-10-26
Only allows admin and commander to update an incident if visibility is restricted

mvilanova:


src/dispatch/incident/views.py
```diff
@@ -138,6 +138,19 @@ def update_incident(
...
         raise HTTPException(status_code=404, detail="The requested incident does not exist.")
 
+    # we want to provide additional protections around restricted incidents
+    if incident.visibility == Visibility.restricted:
+        # reject if the user isn't an admin or commander
+        if incident.commander.email == current_user.email:
+            return incident
+
+        if current_user.role == UserRoles.admin:
+            return incident
```
kevgliss, line 148: I think we want to run the update code here.

kevgliss (changes requested):

src/dispatch/incident/views.py
mvilanova: Yup, I just noticed the error. Added a commit that fixes it.

src/dispatch/incident/views.py
```diff
@@ -138,6 +138,16 @@ def update_incident(
     if not incident:
         raise HTTPException(status_code=404, detail="The requested incident does not exist.")
 
+    # we want to provide additional protections around restricted incidents
+    if incident.visibility == Visibility.restricted:
+        # reject if the user isn't an admin or commander
+        if current_user.email != incident.commander.email or current_user.role != UserRoles.admin:
+            return incident
```
kevgliss, line 145: We should throw the exception here instead of returning.

mvilanova: Fixed.

kevgliss (approved):

PR #650, opened 2020-10-27
Fixes issue where a user can register an account as admin

mvilanova:


src/dispatch/auth/views.py
```diff
@@ -107,6 +107,8 @@ def register_user(
 ):
     user = get_by_email(db_session=db_session, email=user_in.email)
     if not user:
+        # we only allow users to register as a regular user
+        user_in.role = UserRoles.user.value
```
kevgliss, line 111: I think we should remove "role" from the UserRegister model and then push this logic into "create" (defaulting to user if no role is provided).

mvilanova: Sounds good.

PR #655, opened 2020-11-05
Fixes filtering on several columns

kevgliss:
Enables/Fixes filtering on:

- Tag Table -> Tag Type
- Task Table -> Creator, Owner, Due By, Incident Name, Incident Priority, Incident Type

### DIFF
--- a/src/dispatch/static/dispatch/src/task/Table.vue
+++ b/src/dispatch/static/dispatch/src/task/Table.vue
@@ -30,7 +30,7 @@
               :sort-by.sync="sortBy"
               :sort-desc.sync="descending"
             >
-              <template v-slot:item.creator="{ item }">
+              <template v-slot:item.creator.individual_contact.name="{ item }">
                 <v-chip
                   v-if="item.creator"
                   class="ma-2"
@@ -44,7 +44,7 @@
                   Unknown
                 </v-chip>
               </template>
-              <template v-slot:item.owner="{ item }">
+              <template v-slot:item.owner.individual_contact.name="{ item }">
                 <v-chip
                   v-if="item.owner"
                   class="ma-2"
@@ -57,6 +57,12 @@
                 <v-chip v-else class="ma-2" pill small>
                   Unknown
                 </v-chip>
+              </template>
+              <template v-slot:item.incident_priority.name="{ item }">
+                {{ item.incident.incident_priority.name }}
+              </template>
+              <template v-slot:item.incident_type.name="{ item }">
+                {{ item.incident.incident_type.name }}
               </template>
               <template v-slot:item.tickets="{ item }">
                 <a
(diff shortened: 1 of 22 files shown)


src/dispatch/static/dispatch/src/task/Table.vue
```diff
@@ -30,7 +30,7 @@
               :sort-by.sync="sortBy"
               :sort-desc.sync="descending"
             >
-              <template v-slot:item.creator="{ item }">
+              <template v-slot:item.creator.individual_contact.name="{ item }">
```
mvilanova, line 33: Shouldn't this be `individual` not `individual_contact` based on the Participant model?

src/dispatch/static/dispatch/src/task/Table.vue
```diff
@@ -44,7 +44,7 @@
                   Unknown
                 </v-chip>
               </template>
-              <template v-slot:item.owner="{ item }">
+              <template v-slot:item.owner.individual_contact.name="{ item }">
```
mvilanova, line 47: Idem here.

src/dispatch/static/dispatch/src/task/Table.vue
```diff
@@ -58,6 +58,12 @@
                   Unknown
                 </v-chip>
               </template>
+              <template v-slot:item.incident_priority.name="{ item }">
```
mvilanova, line 61: I think `incident` is missing after `item.`

src/dispatch/static/dispatch/src/task/Table.vue
```diff
@@ -58,6 +58,12 @@
...
+                {{ item.incident.incident_priority.name }}
+              </template>
+              <template v-slot:item.incident_type.name="{ item }">
```
mvilanova, line 64: Same here.

src/dispatch/static/dispatch/src/task/Table.vue
```diff
@@ -136,17 +142,17 @@ export default {
...
-        { text: "Incident Priority", value: "incident.incident_priority.name", sortable: false },
-        { text: "Incident Type", value: "incident.incident_type.name", sortable: false },
+        { text: "Incident Name", value: "incident.name", sortable: true },
+        { text: "Incident Priority", value: "incident_priority.name", sortable: true },
+        { text: "Incident Type", value: "incident_type.name", sortable: true },
         { text: "Status", value: "status", sortable: true },
-        { text: "Creator", value: "creator", sortable: true },
-        { text: "Owner", value: "owner", sortable: true },
+        { text: "Creator", value: "creator.individual_contact.name", sortable: true },
+        { text: "Owner", value: "owner.individual_contact.name", sortable: true },
```
mvilanova, line 150: `individual` not `individual_contact`?

kevgliss: No, the actual model is individual contact (which is then resolved and used for filtering). We have a disconnect in that the actual model is called `individual_contact` but we refer to it as `individual` in most places.

src/dispatch/static/dispatch/src/task/Table.vue
kevgliss: No, the slot name has to match the value defined as the value in the header value. It has no reflection on the actual item data structure.

src/dispatch/static/dispatch/src/task/Table.vue
kevgliss: Same as above

src/dispatch/static/dispatch/src/task/Table.vue
kevgliss: Same as above.

src/dispatch/static/dispatch/src/task/Table.vue
kevgliss: Same as above.

src/dispatch/static/dispatch/src/task/Table.vue
mvilanova: Got it.

src/dispatch/static/dispatch/src/task/Table.vue
mvilanova: Yeah, it's a little bit confusing.

mvilanova (approved):

PR #660, opened 2020-11-06
Enables evergreen notifications for documents.

kevgliss:
Adds support for marking documents as eligible for evergreen notifications:

PR #661, opened 2020-11-10
Adds the ability to map/engage the liaison role.

kevgliss:


src/dispatch/incident/service.py
```diff
@@ -196,35 +207,15 @@ def create(
...
-        reporter_email,
-        incident_type.name,
-        "",
-        title,
-        description,
-        incident_priority.page_commander,
+    # Add other incident roles (e.g. commander and liaison)
+    assign_incident_role(
+        db_session, incident, reporter_email, ParticipantRoleType.incident_commander
     )
```
mvilanova, line 217: nit: we could avoid passing `reporting_email` and use the incident `reporter` hybrid property inside `assign_incident_role()`

src/dispatch/incident/service.py
```diff
@@ -25,41 +24,53 @@
...
         oncall_plugin.instance.page(
-            service_id=commander_service.external_id,
-            incident_name=incident_name,
-            incident_title=incident_title,
-            incident_description=incident_description,
+            service_id=service.external_id,
+            incident_name=incident.name,
+            incident_title=incident.title,
+            incident_description=incident.description,
         )
```
mvilanova, line 72: Should we move it inside the `if role == ParticipantRoleType.incident_commander` above and check if we have an `oncall_plugin` before paging?

src/dispatch/incident/service.py
```diff
@@ -25,40 +24,58 @@
...
-    incident_description: str,
-    page_commander: bool,
+    role: ParticipantRoleType,
 ):
-    """Resolves the correct incident commander email based on given parameters."""
-    # We resolve the incident commander email
+    """Assigns incident roles."""
+    # We resolve the incident role email
     # default to reporter if we don't have an oncall plugin enabled
+    assignee_email = None
```
mvilanova, line 36: If you assign `assignee_email = reporter_email` here you could avoid all the assignments below by checking if the service exists vs not.

src/dispatch/alembic/versions/c6db1c7655db_.py
```diff
@@ -0,0 +1,32 @@
+"""Adds evergreen data for documents.
```
mvilanova, line 1: This file should not be part of this PR. It's already in #660

kevgliss: Yeah, I backported it so that I could generate the new fields I needed on the database. It should merge cleanly?

mvilanova: I guess so.

mvilanova (approved):

PR #662, opened 2020-11-12
Hardens the incident creation flow against failures in plugins.

kevgliss:
As almost all plugins are optional at this point, we are able to harden against their individual failures. Meaning that one plugin's failure should not completely stop the flow (and thus other resources) from being created. This will allow individuals to still get some value from submitting an incident (even if the overall experience is degraded due to a failing plugin).


src/dispatch/incident/flows.py
```diff
@@ -503,80 +514,107 @@ def incident_create_flow(*, incident_id: int, checkpoint: str = None, db_session
...
-        )
-        incident.conference = conference_service.create(
-            db_session=db_session, conference_in=conference_in
-        )
+        try:
+            if group_plugin:
+                conference = create_conference(incident, [tactical_group["email"]], db_session)
+            else:
+                # we don't have a group to manage this resource
+                conference = create_conference(incident, participant_emails, db_session)
```
mvilanova, line 554:
```suggestion
            participants = participant_emails
            
            if group_plugin:
                # we use the tactical group email if the group plugin is enabled
                participants = [tactical_group["email"]]
                
            conference = create_conference(incident, participants, db_session)
```

mvilanova (approved):

PR #664, opened 2020-11-15
Unused db connection fix

Liamvdv:
Both get_me (/auth/views.py) and get_current_user() (/auth/service.py) depend on db_session, but auth/me doesn't pass it when calling get_current_user(), hence creating two db_session for handeling one request. Sorry for not bumping dispatch/__about__.py, this is my first open source commit and I'm unable to locate that file.


mvilanova:
Nice catch. Thanks for the improvement!

PR #676, opened 2020-11-18
Adds sample data migration script.

kevgliss:
This should make it easier to update the sample data for each release.


data/update-example-data.sh
```diff
@@ -0,0 +1,8 @@
+echo "Dropping existing database..."
```
mvilanova, line 1: `#!/usr/bin/env bash` is missing.

PR #682, opened 2020-11-19
Feature/incident feedback digest

mvilanova:
This PR adds a daily digest for incident feedback. The digest is emailed to the commander.

Screen Shot 2020-11-19 at 3 36 28 PM

### DIFF
--- a/src/dispatch/feedback/messaging.py
+++ b/src/dispatch/feedback/messaging.py
@@ -1,4 +1,5 @@
 import logging
+from typing import List
 
 from dispatch.config import DISPATCH_HELP_EMAIL
 from dispatch.database import SessionLocal
@@ -8,12 +9,14 @@ from dispatch.messaging import (
 )
 from dispatch.plugin import service as plugin_service
 
+from .models import Feedback
+
 
 log = logging.getLogger(__name__)
 
 
 def send_incident_feedback_daily_digest(
-    commander_email: str, feedback: dict, db_session: SessionLocal
+    commander_email: str, feedback: List[Feedback], db_session: SessionLocal
 ):
     """
     Sends an incident feedback daily digest to all incident commanders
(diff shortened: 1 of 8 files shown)


src/dispatch/feedback/service.py
```diff
@@ -14,6 +15,19 @@ def get_all(*, db_session):
...
 
+def get_all_last_x_hours(
+    *, db_session, hours: int = 24, skip=0, limit=100
+) -> List[Optional[Feedback]]:
+    """Returns all feedback provided in the last x hours."""
+    return (
+        db_session.query(Feedback)
+        .filter(Feedback.created_at >= datetime.utcnow() - timedelta(hours=hours))
+        .offset(skip)
+        .limit(limit)
```
kevgliss, line 26: Why do we limit this? You don't seem to use pagination anywhere. Same with skip/offset.

src/dispatch/feedback/messaging.py
```diff
@@ -0,0 +1,59 @@
...
+                "rating": piece.rating,
+                "feedback": piece.feedback,
+                "participant": participant,
+                "created_at": piece.created_at,
+            }
+        )
+
+    name = subject = "Incident Feedback Daily Digest"
+    plugin.instance.send(
+        commander_email,
```
kevgliss, line 49: Did we want to add support to cc the security team?

kevgliss (approved):

src/dispatch/feedback/messaging.py
mvilanova: Yes, I was going to work on it in a separate PR.

src/dispatch/feedback/service.py
mvilanova: Correct, not needed, will remove.

PR #691, opened 2020-11-23
Refactoring and improving engage oncall messaging

kevgliss:
This cleans up the engage oncall flow and removes slack specific dependencies.

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -45,7 +45,6 @@ from dispatch.participant.models import Participant
 from dispatch.participant_role import flows as participant_role_flows
 from dispatch.participant_role.models import ParticipantRoleType
 from dispatch.plugin import service as plugin_service
-from dispatch.plugins.base import plugins
 from dispatch.report.enums import ReportTypes
 from dispatch.report.messaging import send_incident_report_reminder
 from dispatch.service import service as service_service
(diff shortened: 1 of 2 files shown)


mvilanova (approved):

mvilanova:
I think we should replace `oncall_plugin.title` with `oncall_plugin.instance.title` [here](https://github.com/Netflix/dispatch/pull/691/files#diff-9f1e8b67dd91333ce78c91db2036e284006cf54fa05595c9e43b0b66451d628fR1096) and [here](https://github.com/Netflix/dispatch/pull/691/files#diff-9f1e8b67dd91333ce78c91db2036e284006cf54fa05595c9e43b0b66451d628fR1109).

src/dispatch/incident/flows.py
```diff
@@ -1107,6 +1111,8 @@ def incident_engage_oncall_flow(
             incident_id=incident.id,
         )
 
+    return oncall_email
```
mvilanova, line 1114: Suggestion: we could return the individual and oncall_service, so the message we create in `handle_engage_oncall_action` could say `You have successfully engage X from team Y`.

kevgliss:
> I think we should replace `oncall_plugin.title` with `oncall_plugin.instance.title` [here](https://github.com/Netflix/dispatch/pull/691/files#diff-9f1e8b67dd91333ce78c91db2036e284006cf54fa05595c9e43b0b66451d628fR1096) and [here](https://github.com/Netflix/dispatch/pull/691/files#diff-9f1e8b67dd91333ce78c91db2036e284006cf54fa05595c9e43b0b66451d628fR1109).

I believe the model will be the most up to date, so I would rather use it's metadata.

PR #698, opened 2020-11-30
Feature/export wizard

kevgliss:
Initial work on an export wizard. Some pieces will need to be refactored if we extend this pattern to other data models. 

Screenshot from 2020-11-30 10-24-50
Screenshot from 2020-11-30 10-24-32
Screenshot from 2020-11-30 10-24-13


forestmonster:
Nice to clearly see where you are in the multi-step process. I also like the indicator showing the number of rows to be exported.

src/dispatch/static/dispatch/src/incident/TableExportDialog.vue
```diff
@@ -0,0 +1,245 @@
...
+              <v-list-item>
+                <v-list-item-content>
+                  <incident-priority-combobox v-model="filters.incident_priority" />
+                </v-list-item-content>
+              </v-list-item>
+              <v-list-item>
+                <v-list-item-content>
+                  <incident-status-multi-select v-model="filters.status" />
+                </v-list-item-content>
+              </v-list-item>
```
mvilanova, line 50: Are we missing a label for these three?

src/dispatch/static/dispatch/src/incident/TableExportDialog.vue
```diff
@@ -0,0 +1,245 @@
...
+import IncidentStatusMultiSelect from "@/incident/IncidentStatusMultiSelect.vue"
+import TagFilterCombobox from "@/tag/TagFilterCombobox.vue"
+import IncidentTypeCombobox from "@/incident_type/IncidentTypeCombobox.vue"
+import IncidentPriorityCombobox from "@/incident_priority/IncidentPriorityCombobox.vue"
+export default {
+  name: "IncidentTableExportDialog",
+  data() {
+    return {
+      e1: 1,
+      selectedFields: [
```
mvilanova, line 117: Should type and perhaps priority be also selected by default?

mvilanova (approved):
Echoing @forestmonster feedback. Nice work!

src/dispatch/static/dispatch/src/incident/TableExportDialog.vue
kevgliss: I believe they all have default labels within the components themselves.

PR #709, opened 2020-12-04
More oncall service fixes

mvilanova:


kevgliss (approved):
It looks like we do the same logic quite a bit, should the plugin active checking logic exist within `get_by_slug`?

mvilanova:
Yes, I think that would be a good place to encapsulate it. We can tackle it in a separate PR.

PR #713, opened 2020-12-09
Oncall plugin-related UI warnings and improvements

mvilanova:
Screen Shot 2020-12-09 at 10 47 08 AM

### DIFF
--- a/src/dispatch/plugin/service.py
+++ b/src/dispatch/plugin/service.py
@@ -49,14 +49,14 @@ def get_enabled_by_type(*, db_session, plugin_type: str) -> List[Optional[Plugin
     )
 
 
-def get_all(*, db_session) -> List[Optional[Plugin]]:
-    """Returns all plugins."""
-    return db_session.query(Plugin)
-
-
 def get_by_slug(*, db_session, slug: str) -> Plugin:
     """Fetches a plugin by slug."""
     return db_session.query(Plugin).filter(Plugin.slug == slug).one_or_none()
+
+
+def get_all(*, db_session) -> List[Optional[Plugin]]:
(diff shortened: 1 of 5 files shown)


src/dispatch/plugin/service.py
```diff
@@ -85,6 +88,15 @@ def update(*, db_session, plugin: Plugin, plugin_in: PluginUpdate) -> Plugin:
                     f"Cannot disable plugin: {plugin.title}. It is required and no other plugins of type {plugin.type} are enabled."
                 )
 
+        if plugin.type == OncallPlugin.type:
+            oncall_services = service_service.get_all_by_type_and_status(
+                db_session=db_session, service_type=plugin.slug, is_active=True
+            )
+            if oncall_services:
+                raise InvalidConfiguration(
+                    f"Cannot disable plugin: {plugin.title}. One or more oncall services depend on it. "
```
kevgliss, line 97: Should we include which services are dependent so the user doesn't have to go and find them?

kevgliss (approved):

src/dispatch/plugin/service.py
mvilanova: I thought about it, but decided not to for two reasons: 1) the error message could get very long if there are a lot of services that depend on it 2) the error message disappears after a few seconds. Instead I make it easier for the user by adding a column for enabled or not.

kevgliss: We could always make the error message last longer (or until dismissed).

mvilanova: I think we should do it regardless. I find it annoying that it disappears so quickly.

PR #714, opened 2020-12-10
UI errors are visible until closed

mvilanova:
Screen Shot 2020-12-14 at 10 39 27 AM

### DIFF
--- a/src/dispatch/static/dispatch/src/app/store.js
+++ b/src/dispatch/static/dispatch/src/app/store.js
@@ -4,16 +4,15 @@ import router from "@/router"
 const getDefaultSnackbarState = () => {
   return {
     text: null,
-    color: "primary",
-    show: false,
-    timeout: 2000
+    type: "info",
+    show: false
   }
 }
 
 const getDefaulRefreshState = () => {
   return {
     show: false,
-    message: "blah"
+    message: ""
   }
 }
 
--- a/src/dispatch/static/dispatch/src/app/store.js
+++ b/src/dispatch/static/dispatch/src/app/store.js
@@ -65,7 +65,7 @@ const mutations = {
     state.loading = value
   },
   RESET_SNACKBAR(state) {
-    state.snackbar = Object.assign({}, getDefaultSnackbarState())
+    state.snackbar = getDefaultSnackbarState()
   },
   RESET_REFRESH(state) {
     state.refresh = Object.assign(state.refresh, getDefaulRefreshState())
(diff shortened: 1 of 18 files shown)


src/dispatch/static/dispatch/src/app/store.js
```diff
@@ -65,7 +65,7 @@ const mutations = {
     state.loading = value
   },
   RESET_SNACKBAR(state) {
-    state.snackbar = Object.assign(state.snackbar, getDefaultSnackbarState())
+    state.snackbar = Object.assign({}, getDefaultSnackbarState())
```
mvilanova, line 68: @kevgliss once I fixed the close button (see change in line 39), an uninitialized snackbar would appear after closing the snackbar with the error message. This change fixes the issue, but I'm not sure it's the best way.

src/dispatch/static/dispatch/src/app/store.js
```diff
@@ -65,7 +65,7 @@ const mutations = {
...
   },
   RESET_REFRESH(state) {
     state.refresh = Object.assign(state.refresh, getDefaulRefreshState())
```
mvilanova, line 71: We may want to do the same here, replace `state.refresh` with `{}`, but this doesn't seem to be causing any issues that I can tell.

src/dispatch/static/dispatch/src/plugin/store.js
```diff
@@ -83,7 +83,8 @@ const actions = {
             "app/SET_SNACKBAR",
             {
               text: "Plugin not created. Reason: " + err.response.data.detail,
-              color: "red"
+              color: "red",
+              timeout: 0
```
mvilanova, line 87: If we're happy with this, I can go and add it to the rest of stores.

src/dispatch/static/dispatch/src/app/store.js
kevgliss: I don't think we need `Object.assign` here. It could just be `state.snackbar = getDefaultSnackbarState()`

src/dispatch/static/dispatch/src/plugin/store.js
kevgliss: If we want to force users to always close `error` messages (instead of timing out). I wonder if we should instead push that down into the component itself. 

i.e. we would pass a `type` prop to to the Snackbar component. When the component encounters any "error" type messages it will set the timeout to `0` and the ` color to `red`.

We can then also define `success` and `info` type messages. This way, if we want to change how those messages behave/look we only need to deal with the component (instead of modifying all messages like we have to now).

kevgliss (approved):

PR #715, opened 2020-12-11
Improvement/hawkins colors

kevgliss:
This work aligns Dispatch to the hawkins design language.

Screenshot from 2020-12-10 12-44-22
Screenshot from 2020-12-10 12-40-08
Screenshot from 2020-12-10 12-39-35
Screenshot from 2020-12-14 12-15-01


src/dispatch/search/views.py
```diff
@@ -40,11 +41,6 @@ def search(
...
     if current_user.role != UserRoles.admin:
-        filtered_results = []
-        for r in results:
-            if r["type"].lower() == "incident":
-                if r["content"].visibility != Visibility.open:
-                    continue
-            filtered_results.append(r)
-        results = filtered_results
-    return {"query": q, "results": results}
+        results["Incident"] = [i for i in results["Incident"] if i.visibility != Visibility.open]
```
mvilanova, line 44: Looks like we're returning restricted incidents instead of filtering them out here if I'm interpreting it right.

src/dispatch/static/dispatch/src/App.vue
```diff
@@ -1,13 +1,15 @@
...
-        Update
-      </v-btn>
-    </v-snackbar>
-  </div>
+  <v-app>
+    <div class="app-root">
+      <router-view />
+      <v-snackbar bottom right :value="updateExists" :timeout="0" color="info">
+        An update is available
+        <v-btn text @click="refreshApp">
```
mvilanova, line 7: FYI When I made the changes to the snackbar component, I wasn't able to see the button unless I set the `color` property to something different than the one in the snackbar (e.g. `color="primary"`).

PR #727, opened 2020-12-18
Adds cascade behavior to data relationships

mvilanova:
This PR adds cascade behavior to data relationships in order to be able to delete an incident object and all its associated resource objects (e.g. groups, documents, etc.).


src/dispatch/document/models.py
```diff
@@ -38,14 +38,6 @@
     PrimaryKeyConstraint("incident_type_id", "document_id"),
 )
 
-assoc_document_incidents = Table(
```
mvilanova, line 41: This table is not being used.

kevgliss (commented):
This looks good to me so far.

PR #730, opened 2020-12-18
Improvement/commander report participant

kevgliss:
This aligns incident commander and reporter to act on and represent a _participant_ instead of an _individual_. This aligns with how we generally think about those interacting with incidents and aligns us with how tasks are currently handled.


src/dispatch/report/flows.py
```diff
@@ -196,7 +196,7 @@ def create_executive_report(
         overview=overview,
         next_steps=next_steps,
         previous_reports="\n".join(previous_executive_reports),
-        commander_fullname=incident.commander.name,
+        commander_fullname=incident.commander.individual.individual.name,
```
mvilanova, line 199: `individual` twice
```suggestion
        commander_fullname=incident.commander.individual.name,
```

src/dispatch/static/dispatch/src/incident/DetailsTab.vue
```diff
@@ -62,12 +62,12 @@
             hint="The incident's current commander"
```
mvilanova, line 62:
```suggestion
            hint="The participant acting as incident commander."
```

src/dispatch/static/dispatch/src/incident/DetailsTab.vue
```diff
@@ -76,7 +76,7 @@
             hint="The incident's current reporter"
```
mvilanova, line 76:
```suggestion
            hint="The participant who reported the incident."
```

PR #732, opened 2020-12-24
Restricts some commands from participants

kevgliss:


src/dispatch/plugins/dispatch_slack/messaging.py
```diff
@@ -115,13 +115,29 @@
     "\n", " "
 ).strip()
 
+INCIDENT_CONVERSATION_COMMAND_RUN_BY_NON_PRIVILEGED_USER = """
+Looks like you tried to run `{{command}}`.
+This is a sensitive command and cannot be run by regular participants.""".replace(
```
mvilanova, line 120: We could also say: "This is a command reserved for the incident commander."

mvilanova: Never mind. I see you allow everyone except individuals with participant role in the code.

mvilanova (approved):

PR #741, opened 2021-01-04
Fixing current participant roles

kevgliss:


src/dispatch/plugins/dispatch_slack/views.py
```diff
@@ -221,7 +221,7 @@ async def handle_command(
         )
 
         # allow everyone other than participants engage in sensitive commands
-        if participant.role == ParticipantRoleType.participant:
+        if participant.current_role.role == ParticipantRoleType.participant:
```
mvilanova, line 224: Heads up that participants can have more than one current role and you could be comparing a list or roles with a string here.

kevgliss: Good call, I've updated my logic.

PR #763, opened 2021-01-08
Refactors and consolidates email templates

mvilanova:


src/dispatch/messaging/strings.py
```diff
@@ -285,13 +285,13 @@ class MessageType(str, Enum):
 Thanks for participating in the {{name}} ("{{title}}") incident. We would appreciate if you could rate your experience and provide feedback."""
 
 INCIDENT_TYPE_CHANGE_DESCRIPTION = """
-The incident type has been changed from *{{ incident_type_old }}* to *{{ incident_type_new }}*."""
```
mvilanova, line 288: These were not getting rendered on the emails so I took them out.

PR #771, opened 2021-01-13
Fix jira usage per lib docs

sfc-gh-pkommini:
- Fix JIRA config variables
- Fix usage per new API after GDPR changes which only allows `accountId` instead of `username` fields
- Fix docs for configuring JIRA plugin
- Pin `jira` library to `2.0.0` in order to lockdown library usage per library version and allow upgrades after review of version stability


requirements-base.txt
```diff
@@ -14,7 +14,7 @@ google-auth-oauthlib
 h11==0.9.0 # NOTE pinning due to uvicorn
 httpx
 jinja2
-jira
+jira==2.0.0
```
mvilanova, line 17: I'm not sure we should pin the version. I'd rather update the plugin code when new versions are forcing us to do it.

mvilanova:
Changes look good to me. We can merge after deciding if we want to pin the jira Python client version or not.

kevgliss:
I don't like pinning the client either, but if atlassian decides to only update the client when the API version changes (no backward compatibility) it might make sense to pin it.

requirements-base.txt
sfc-gh-pkommini: That makes sense. I can push undo the pinning. The rationale was changing when you decide to change rather than waiting for it to break and then change the version.

PR #777, opened 2021-01-15
Adds a new model, service, and views for notifications

mvilanova:
This PR adds a new data model, service, and views to support configuring notifications. It also refactors the policy service module.


src/dispatch/notification/models.py
```diff
@@ -0,0 +1,82 @@
...
+    Column("incident_priority_id", Integer, ForeignKey("incident_priority.id", ondelete="CASCADE")),
+    Column("notification_id", Integer, ForeignKey("notification.id", ondelete="CASCADE")),
+    PrimaryKeyConstraint("incident_priority_id", "notification_id"),
+)
+
+
+class Notification(Base, TimeStampMixin):
+    # Columns
+    id = Column(Integer, primary_key=True)
+    conversation_name = Column(String)
```
kevgliss, line 40: Should this be more general? I think we could use `service` as a good template: https://github.com/Netflix/dispatch/blob/master/src/dispatch/service/models.py#L53

That way we could use these to define email notifications as well as slack conversations.

kevgliss (changes requested):

src/dispatch/notification/models.py
mvilanova: Yes, I was thinking we could allow users to create notifications and allow them to select from a list of already defined notification (or message types) and plugins that provide communications-related functionality (e.g. email, conversation), and then, depending on the plugin selected, you would provide where that notification has to go. For conversation notifications, we could support both channels and users.

kevgliss: The other portion is to allow a user to define a subset of incidents to be notified for. The service model allows terms, but for notifications, I assume we would also want to allow tags to be associated with notifications such that, when an incident has a given tag that notification will be sent. 

For example: userFooNotification(tag1 OR tag2) -- incidentBar(tag2) -- email -> send email notification

kevgliss (approved):
This lgtm, if you want to start on the UI I can start on the expression analyzer. You might need to revisit the policy API endpoint and UI as that part of the code has been neglected for a bit.

mvilanova:
Sounds like a good plan

PR #784, opened 2021-01-19
Fixing issue with migration and data dump

kevgliss:
@mvilanova I removed some of your drop constraints from the migration file. It looks like you can't rename a constraint if it doesn't exist.  Neither of these constraints existed in the sample database which we mostly keep up-to-date. I'm not sure if that rename was an artifact or something else, but didn't seem to be needed here. 

Let me know if there was some other outlier I'm missing.


mrthankyou:
If this can get the data dump working that will be such a life-saver for me instead of having to recreate data all the time!

mvilanova:
We should re-add them. They could exist in databases from when the data model was called status_report (like ours). We updated the table name to report, but not the pkey, fkeys, and sequence id. We should keep the naming consistent. The IF EXIST should prevent the statement from throwing an error. Was that not the case?

kevgliss:
Your `IF EXISTS` only applies to the report table, not the constraint itself. In this case, the table did exist but the constraint does not. Also, according to the Postgres documentation, you can't apply a RENAME IF EXISTS. You'll notice that alembic always drops and then creates new constraints when renaming. 

Alembic will create new constraints for us, we shouldn't have to manage the rename ourselves. 

Additionally, I don't see `status_report_participant_id_fkey` or `status_report_incident_id_fkey` anywhere in the repo. Which would lead me to believe that they weren't created (or managed) by alembic. 

As an example you can see your new constraint being added by alembic automatically here: 
https://github.com/Netflix/dispatch/blob/69da45c0388b5147ffc6266e398376dfcf427220/src/dispatch/alembic/versions/38f0b01448ba_.py#L80 

I would have expected to see that for the old constraint in an old migration.

mvilanova:
Yeah, you won’t find `status_report_participant_id_fkey` or `status_report_incident_id_fkey` constrains in the code repo, because they would have been part of the initial alembic migration when the data model was called `status_report`: https://github.com/Netflix/dispatch/pull/356/files#diff-95d6b1585d93baa441da5ee419b414f7670186f7449a438202d28a3c2916ef41L32.

You can see how they would get created by relationships https://github.com/Netflix/dispatch/pull/356/files#diff-f863d914618068ec3bdcb4a9ad03accd36d9df84cdd0290f25889847f6bafa71L27 and https://github.com/Netflix/dispatch/pull/356/files#diff-f29c22e331c5c482cff30440ff9f626024d6bfc5335e4b2cfeedfea0ffef1cfeL163.

Again, I added them to ensure migration `38f0b01448ba_.py` would not fail [here](https://github.com/Netflix/dispatch/blob/69da45c0388b5147ffc6266e398376dfcf427220/src/dispatch/alembic/versions/38f0b01448ba_.py#L80) for old databases created between when we open sourced the project and May 28, 2020 when we merged https://github.com/Netflix/dispatch/pull/356. I ran into the migration issue in test, and prod had the same out of date constrains, so I thought other users could run into the same issue and I felt it would look bad if I had to ask them to rename the constrains manually rather than having alembic take care of it. I did not take the sample data file into consideration. Given that it's probably more likely that new users are running into an issue when they're doing a database upgrade than hopefully old ones, I'm ok with leaving them out.

mrthankyou:
Note: I've tested this on my end and the database dump is now working! Thanks guys!

PR #785, opened 2021-01-19
Makes incident conversation creation more robust

mvilanova:
This PR decouples incident channel creation from adding participants to it. Now we wait after the incident conversation has been created via the plugin and in the database to add participants to it.

Additionally, we will also retry when Slack API calls return `fatal_error`, which seems to be when Slack's performance is downgraded.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/service.py
+++ b/src/dispatch/plugins/dispatch_slack/service.py
@@ -99,8 +99,7 @@ SLACK_GET_ENDPOINTS = [
 
 @retry(stop=stop_after_attempt(5), retry=retry_if_exception_type(TryAgain))
 def make_call(client: Any, endpoint: str, **kwargs):
-    """Make an slack client api call."""
-
+    """Make an Slack client api call."""
     try:
         if endpoint in SLACK_GET_ENDPOINTS:
             response = client.api_call(endpoint, http_verb="GET", params=kwargs)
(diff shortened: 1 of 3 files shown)


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -117,6 +116,10 @@ def make_call(client: Any, endpoint: str, **kwargs):
         if e.response["error"] == "user_not_in_channel":
             raise TryAgain
 
+        # NOTE we've experienced a wide range of issues when Slack's performance is degraded
+        if e.response["error"] == "fatal_error":
```
kevgliss, line 120: I think we should add a wait here, if the service is downgraded 5 requests in a row aren't likely to succeed.

kevgliss (approved):
lgtm, minor suggestion

src/dispatch/plugins/dispatch_slack/service.py
mvilanova: Just here under this `if` or should we add `wait=wait_exponential(multiplier=x, min=y, max=z)` to the decorator, so it would also apply to the other cases? The one where we would probably wait for longer than we should would be when we get rate limited.

kevgliss: I'm not sure if there is a clean way to mix and match (while keeping wait for retry), so I was going to just add a simple wait time.

PR #810, opened 2021-01-28
Restricts certain configuration aspects to admins only

mvilanova:
This PR adds logic to restrict the creation and editing of incident types, priorities, and plugins to admins only.

### DIFF
--- a/src/dispatch/incident_type/views.py
+++ b/src/dispatch/incident_type/views.py
@@ -57,7 +57,7 @@ def create_incident_type(
     # We restrict the creation of incident types to admins only
     if current_user.role != UserRoles.admin:
         raise HTTPException(
-            status_code=401, detail="You do not have permission to create incident types."
+            status_code=403, detail="You do not have permission to create incident types."
         )
     incident_type = create(db_session=db_session, incident_type_in=incident_type_in)
     return incident_type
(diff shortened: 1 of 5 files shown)


src/dispatch/incident_type/views.py
```diff
@@ -42,11 +46,19 @@ def get_incident_types(
...
+    current_user: DispatchUser = Depends(get_current_user),
 ):
     """
-    Create a new incident_type.
+    Create a new incident type.
     """
+    # We restrict the creation of incident types to admins only
+    if current_user.role != UserRoles.admin:
+        raise HTTPException(
+            status_code=403, detail="You do not have permission to create incident types."
```
mvilanova, line 60: @kevgliss FYI I used `403` instead of `401` because the latter was triggering re-auth when using basic auth provider.

kevgliss: Should we change the other 401s we have to 403s?

mrthankyou: As a complete outsider I would consider it would good practice to expect the same status codes for the same exception raised by an application. But that's just my two cents ¯\_(ツ)_/¯

mvilanova: Yeah, I'll take care of it.

PR #811, opened 2021-01-28
Centralizes code and creates a more uniform permission system.

kevgliss:


src/dispatch/auth/permissons.py
```diff
@@ -0,0 +1,158 @@
...
+        class TeapotUserAgentPermission(BasePermission):
+
+            def has_required_permissions(self, request: Request) -> bool:
+                return request.headers.get('User-Agent') == "Teapot v1.0"
+
+    """
+
+    error_msg = "Forbidden."
+    status_code = HTTP_403_FORBIDDEN
+    error_code = HTTP_403_FORBIDDEN
```
mvilanova, line 41: Do we need it? It's not used anywhere.

src/dispatch/auth/permissons.py
```diff
@@ -0,0 +1,158 @@
...
+    .. code-block:: python
+
+        class TeapotUserAgentPermission(BasePermission):
+
+            def has_required_permissions(self, request: Request) -> bool:
+                return request.headers.get('User-Agent') == "Teapot v1.0"
+
+    """
+
+    error_msg = "Forbidden."
```
mvilanova, line 39: Should we customize this error message based on what the user tries to do?

mvilanova (approved):
Looks good. Just some nits.

src/dispatch/auth/permissons.py
kevgliss: We can override this when we construct the permission. But this is the default.

PR #828, opened 2021-02-03
Allowing notification filters to be optional.

kevgliss:


src/dispatch/notification/service.py
```diff
@@ -83,6 +83,23 @@ def delete(*, db_session, notification_id: int):
...
+            notification_params["type"],
+            **notification_params["kwargs"],
+        )
+    else:
+        log.warning(
+            f"Notification {notification.name} not sent. No {notification.type} plugin is active."
+        )
+
+
 def send(*, db_session, class_instance: Type[Base], notification_params: dict):
```
mvilanova, line 103: I'd rename this function to `filter_and_send` and then `send_matched` to `send`.

mvilanova: Calling `send_matched` [here](https://github.com/Netflix/dispatch/pull/828/files#diff-f884284e6f0f4bcc03313f8793e215f633c992cd1f6a3cfb92aa5ef1564956c5R121) when there was no match feels a little bit weird to me.

mvilanova (approved):

PR #865, opened 2021-02-11
Adds support for incident daily reports based on notification filters

mvilanova:


src/dispatch/incident/scheduled.py
```diff
@@ -97,55 +101,83 @@ def daily_report(db_session=None):
...
+                    }
+
+                    if incident.status != IncidentStatus.closed.value:
+                        item.update(
+                            {
+                                "button_text": "Join Incident",
+                                "button_value": str(incident.id),
+                                "button_action": f"{ConversationButtonActions.invite_user.value}-{incident.status}-{idx}",
+                            }
+                        )
```
mvilanova, line 152: We could probably move this to incidents/messaging.py

PR #880, opened 2021-02-16
Allowing the sentry configuration for the UI to be more flexible.

kevgliss:


src/dispatch/static/dispatch/src/main.js
```diff
@@ -13,11 +13,22 @@ import * as Integrations from "@sentry/integrations"
 
 import VueClipboard from "vue-clipboard2"
 
-if (process.env.VUE_APP_SENTRY_DSN) {
+// Configure sentry
+if (process.env.VUE_APP_SENTRY_ENABLED) {
+  const APP_HOSTNAME = document.location.host
+
+  let DSN = `"https://1:1@${APP_HOSTNAME}/0`
```
mvilanova, line 20: Aren't we missing some `"` here?

PR #901, opened 2021-02-19
Jira plugin now supports cloud and server hosting type

mvilanova:
This PR refactors the Jira plugin to add support for cloud and server hosting types and to respect any provided incident type plugin metadata.


mvilanova:
@sfc-gh-pkommini would you mind checking that everything works on your side? You don't need to add the `JIRA_HOSTING_TYPE` envvar in your `.env` file as it defaults to `Cloud`.

sfc-gh-pkommini:
@mvilanova Tested and found that it works.

mvilanova:
Great. Thanks for testing it!

PR #904, opened 2021-02-22
Adds CLI command to uninstall plugins

mvilanova:


src/dispatch/cli.py
```diff
@@ -137,6 +137,27 @@ def install_plugins(force):
         db_session.commit()
 
 
+@plugins_group.command("uninstall")
+@click.argument("plugins")
```
kevgliss, line 141: nit: specify `nargs=-1` and allow for multiple arguments (instead of doing comma-separated)

https://click.palletsprojects.com/en/7.x/arguments/#variadic-arguments

kevgliss (approved):

PR #908, opened 2021-02-22
Adds description and reporter to the welcome message and title, description, and reporter to list resources command

mvilanova:


kevgliss:
Should we run this again on incident/description update?

mvilanova:
We could append the description to the template [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident/messaging.py#L293) if it has changed. We're already including the updated title, and not sure including the reported is necessary.

PR #910, opened 2021-02-23
Create services and documents from incident type

kevgliss:
Adding the ability to create services and documents when creating/editing incident types.

### DIFF
--- a/src/dispatch/static/dispatch/src/service/ServiceSelect.vue
+++ b/src/dispatch/static/dispatch/src/service/ServiceSelect.vue
@@ -1,23 +1,34 @@
 <template>
-  <v-autocomplete
-    v-model="service"
-    :items="items"
-    :search-input.sync="search"
-    :menu-props="{ maxHeight: '400' }"
-    item-text="name"
-    :label="label"
-    placeholder="Start typing to Search"
-    return-object
-    :hint="hint"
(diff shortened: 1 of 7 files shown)


src/dispatch/static/dispatch/src/service/ServiceSelect.vue
```diff
@@ -1,23 +1,34 @@
...
-  </v-autocomplete>
+  <ValidationProvider name="service" immediate>
+    <v-autocomplete
+      v-model="service"
+      :items="items"
+      :search-input.sync="search"
+      :menu-props="{ maxHeight: '400' }"
+      item-text="name"
+      :label="label"
+      placeholder="Start typing to Search"
```
mvilanova, line 10:
```suggestion
      placeholder="Start typing to search"
```

src/dispatch/static/dispatch/src/service/ServiceSelect.vue
```diff
@@ -1,23 +1,34 @@
...
+      return-object
+      :hint="hint"
```
mvilanova, line 12: We should remove the hint or update its string. If we update it, we may want to add one to the ValidationProvider document as well.

mvilanova (approved):

PR #911, opened 2021-02-23
Allowing for '-' in comment ids

kevgliss:


src/dispatch/plugins/dispatch_slack/actions.py
```diff
@@ -28,6 +29,11 @@
 from .decorators import slack_background_task
 
 
+def base64_decode(input: str):
```
mvilanova, line 32: Should we move this and the other b64 function to a `utils.py` module in this plugin or outside of it where other modules can import them?

mvilanova (approved):

src/dispatch/plugins/dispatch_slack/actions.py
kevgliss: We can once we have a need, I'm not sure where else we might need such functionality.

PR #922, opened 2021-02-25
Fixes API query param names

mvilanova:


src/dispatch/static/dispatch/src/search/SearchFilterCreateDialog.vue
```diff
@@ -247,12 +247,12 @@ export default {
           if (has(value, "id")) {
             subFilter.push({
               model: toPascalCase(key),
-              field: "id",
-              op: "==",
-              value: value.id
+              fields: "id",
```
kevgliss, line 250: We don't want to modify these as they are using a POST request to generate the filter and not a GET request

kevgliss (changes requested):

src/dispatch/static/dispatch/src/search/SearchFilterCreateDialog.vue
mvilanova: Ah, good catch.

PR #924, opened 2021-02-26
Ensures re-resolving services doesn't result in multiple enagements

kevgliss:


src/dispatch/incident/flows.py
```diff
@@ -976,18 +970,22 @@ def resolve_incident_participants(incident: Incident, db_session: SessionLocal):
...
+                individual.email, incident.id, service=service, db_session=db_session
             )
 
-        # we add the team distributions lists to the notifications group
-        group_plugin = plugin_service.get_active(
-            db_session=db_session, plugin_type="participant-group"
-        )
-        if group_plugin:
-            team_participant_emails = [x.email for x in team_participants]
-            group_plugin.instance.add(incident.notifications_group.email, team_participant_emails)
```
mvilanova, line 990: My preference here would be to resolve and add both individual and team participants to the incident or only resolve them. Resolving and adding individuals but only resolving teams is confusing and will be harder to maintain.

kevgliss: > My preference here would be to resolve and add both individual and team participants to the incident or only resolve them. Resolving and adding individuals but only resolving teams is confusing and will be harder to maintain.

I don't understand, could you elaborate?

mvilanova: Before your changes, we're resolving and adding individuals to the incident and team participants to the notifications group here. After your changes, we're resolving and adding individuals to the incident, but not adding the team participants to the notifications group. The logic for the latter has been moved outside of this function and to the `incident_update_flow()`. My preference would be to keep it inside `resolve_incident_participants()` to make it more easier to maintain.

kevgliss: I moved it outside of the `resolve_incident_participants` function because adding folks to notification groups does not have anything to do with resolving them. It also meant that I couldn't re-use the resolve function for both the create and update flows (which we now do).

I believe this change specifically improves the maintainability and pureness of the `resolve_incident_participants` function.

mvilanova: Hmm... shouldn't we do the same for individuals then? Only resolving them?

kevgliss: I've pushed up a change, that moves the checking logic `add_or_reactive_participant` function, and removes the `resolve_incident_participants` logic, in favor of putting this logic directly in the create/update flows.

mvilanova (approved):
Looks good!

PR #928, opened 2021-02-26
Refactors incident cost and adds support for non-op costs

mvilanova:
This PR adds support for defining different types of incidents costs and adding incident costs of those types to any given incident using the new Costs tab in the incident details view.

The incident response cost gets added during the database migration and can't be edited or deleted as it's used by the scheduled job that calculates the response cost on a regular basis.

Screen Shot 2021-03-15 at 7 27 46 PM

Screen Shot 2021-03-15 at 7 27 23 PM


src/dispatch/cli.py
```diff
@@ -161,21 +161,24 @@ def sync_triggers():
...
-    sync_trigger(engine, "search_filter", "search_vector", ["name", "description"])
     sync_trigger(engine, "report", "search_vector", ["details_raw"])
+    sync_trigger(engine, "search_filter", "search_vector", ["name", "description"])
     sync_trigger(engine, "service", "search_vector", ["name"])
     sync_trigger(engine, "tag", "search_vector", ["name"])
     sync_trigger(engine, "task", "search_vector", ["description"])
     sync_trigger(engine, "team_contact", "search_vector", ["name", "company", "notes"])
     sync_trigger(engine, "term", "search_vector", ["text"])
-    sync_trigger(engine, "dispatch_user", "search_vector", ["email"])
     sync_trigger(engine, "workflow", "search_vector", ["name", "description"])
```
mvilanova, line 182: @kevgliss I added sync_trigger() calls for some that were missing. I didn't check all models, so there could be more.

src/dispatch/cli.py
```diff
@@ -520,13 +523,23 @@ def dispatch_scheduler():
...
+        sync_active_stable_tasks,
+        sync_tasks,
+    )  # noqa
     from .term.scheduled import sync_terms  # noqa
-    from .workflow.scheduled import sync_workflows, sync_active_stable_workflows  # noqa
+    from .workflow.scheduled import (
+        sync_workflows,
+        daily_sync_workflow,
+        sync_active_stable_workflows,
+    )  # noqa
```
mvilanova, line 542: @kevgliss I added some imports that were missing (e.g. `close_incident_reminder`). Some of the functions we're importing (e.g. `sync_workflows`, `sync_tasks`) do not have the `scheduler` and `background_task` decorators. Mind checking that we're importing only the ones we need?

src/dispatch/incident_cost/scheduled.py
```diff
@@ -0,0 +1,55 @@
...
+log = logging.getLogger(__name__)
+
+
+@scheduler.add(every(5).minutes, name="calculate-incidents-opportunity-cost")
+@background_task
+def calculate_incidents_opportunity_cost(db_session=None):
+    """
+    Calculates and saves the opportunity cost for all incidents.
+    """
+    opportunity_cost_type = incident_cost_type_service.get_by_name("Opportunity Cost", db_session)
```
mvilanova, line 25: We can ship Dispatch with an incident cost type called `Opportunity Cost` or perhaps allow to set a default cost type and use that here.

src/dispatch/incident_cost_type/models.py
```diff
@@ -0,0 +1,44 @@
...
+from dispatch.models import DispatchBase, TimeStampMixin
+
+
+# SQLAlchemy Model
+class IncidentCostType(Base, TimeStampMixin):
+    # columns
+    id = Column(Integer, primary_key=True)
+    name = Column(String)
+    description = Column(String, nullable=False)
+    details = Column(JSONType, nullable=True)
```
mvilanova, line 16: Added this field in case we want to store links to internal/ external docs describing this type of cost, or other things.

src/dispatch/incident_cost/scheduled.py
kevgliss: Yeah, I would prefer we follow setting defaults as we do with incident types.

src/dispatch/cli.py
kevgliss: By the looks of it, the functions you mentioned are called by other functions that have those decorators.

src/dispatch/incident_cost/models.py
```diff
@@ -0,0 +1,41 @@
...
+    IncidentCostTypeRead,
+)
+from dispatch.models import DispatchBase, TimeStampMixin
+
+
+# SQLAlchemy Model
+class IncidentCost(Base, TimeStampMixin):
+    # columns
+    id = Column(Integer, primary_key=True)
+    amount = Column(Numeric(precision=10, scale=2), nullable=False)
```
kevgliss, line 17: We could consider something like a currency type here:

https://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/types/currency.html

In case we want to support other types of currency.

src/dispatch/incident_cost/service.py
```diff
@@ -0,0 +1,197 @@
...
+        ParticipantRoleType.scribe: 0.75,
+        ParticipantRoleType.liaison: 0.75,
+        ParticipantRoleType.participant: 0.5,
+        ParticipantRoleType.reporter: 0.5,
+    }
+
+    return engagement_mappings.get(participant_role)
+
+
+def calculate_incident_opportunity_cost(
```
kevgliss, line 118: Do we think these can be generalized into expressions? e.g. could we define some set of incident components (participant_active_hours, incident_type, priority) and allow the user to define their own algorithm? 

my_oppertunity_cost = participant_active x incident_type x priority ?

kevgliss (commented):
Looks good so far.

src/dispatch/cli.py
mvilanova: Right, so we should only import the ones that have the decorators correct?

kevgliss: Yes, only the ones we decorators need to be imported.

src/dispatch/incident_cost/models.py
mvilanova: My idea was to keep it generic here and then add a setting under /configuration/general, or something like that, where we'd allow users to specify the currency. Then display the amount according to that setting. We could also move other settings in the `.env` to /configuration/general.

src/dispatch/incident_cost/service.py
mvilanova: Yeah, I think so. I can tackle it in a separate PR.

Issue #912, opened 2021-02-23
MJML error with v20210212


AlexaKelley:
**Describe the bug**
When attempting to send emails using the email plugin I get the following in my error log. As with #976 I was not defining the mjml path inside my environmental vars. The incident _does_ create successfully, but it does not send out emails. This build is using v20210212 as base. How would you recommend resolving? I have not worked with mjml before. 

```
2021-02-23 13:55:02.637 EST- /usr/local/lib/python3.8/site-packages/dispatch/static/dispatch/node_modules/.bin/mjml

2021-02-23 13:55:02.637 EST at Function.Module._resolveFilename (internal/modules/cjs/loader.js:980:15)

2021-02-23 13:55:02.637 EST at Function.Module._load (internal/modules/cjs/loader.js:862:27)

2021-02-23 13:55:02.637 EST at Module.require (internal/modules/cjs/loader.js:1042:19)

2021-02-23 13:55:02.637 EST at require (internal/modules/cjs/helpers.js:77:18)

2021-02-23 13:55:02.637 EST at Object.<anonymous> (/usr/local/lib/python3.8/site-packages/dispatch/static/dispatch/node_modules/.bin/mjml:3:1)

[...]

PR #929, opened 2021-03-01
Updating configuration documentation RE: MJML_PATH

AlexaKelley:
The default MJML_PATH is no longer accurate per #912. Updating documented path to reflect what it should be set to. I've also put in some brief wording for those using stock docker image as a stopgap until the coded default can be changed. 

I recommend updating [here](https://github.com/Netflix/dispatch/blob/78ab6484471c2ea79e745141b15c45b23dbba559/src/dispatch/config.py#L101) in the future to reflect new path as well. I would update it as well as part of this PR but I don't have the time to build and test right now.


AlexaKelley:
requesting review from @kevgliss

mvilanova:
Thanks for the PR, @AlexaKelley !

PR #932, opened 2021-03-05
reducing filtered months from forecast

snkilmartin:


src/dispatch/incident/metrics.py
```diff
@@ -56,7 +56,7 @@ def make_forecast(
 
     # exclude last two months
```
mvilanova, line 57: nit: we should update this comment too

src/dispatch/incident/metrics.py
```diff
@@ -56,7 +56,7 @@ def make_forecast(
...
     query = query.filter(
-        Incident.reported_at < date.today().replace(day=1) + relativedelta(months=-2)
+        Incident.reported_at < date.today().replace(day=1) + relativedelta(months=-1)
```
mvilanova, line 59: Curious about what prompted this change. Was filtering out 2 months messing with the forecast results?

kevgliss:
Excluding two months is expected behavior. We exclude two months so that we can present the data in the UI showing the predicted value against the actual value for the previously entire month.

snkilmartin:
Excluding the last two months will have a negative impact on the accuracy of the forecast; the prior month was very important when I evaluated the data internally. If it's just a problem with visualizing the results, can we instead store the historical forecast data in a table and use that to visualize forecasted and actual counts?

kevgliss:
Will the accuracy not just lag by a month?  Is there any concern for a recency bias in the forecast (if the most recent months affect the forecast more so than other months)?

snkilmartin:
I think the forecast is good for:

1. Yearly planning
2. Having an idea of what to expect for the current month

Removing two months won't have as big of an impact on the first, but will be bad for the second.

RE recency, since this is a simple exponential smoothing model, the most recent data is the most important. You don't have to believe me though :) I'll quote Rob Hyndman since he's who I like to reference about forecasting:

>  Forecasts produced using exponential smoothing methods are weighted averages of past observations, with the weights decaying exponentially as the observations get older. In other words, the more recent the observation the higher the associated weight.

mvilanova:
Sounds like we should merge the change then?

kevgliss:
> Sounds like we should merge the change then?

No, I have another PR that will fix this issue. Closing this one.

PR #941, opened 2021-03-10
Allows for more flexible forecasting

kevgliss:
This PR improves our ability to forecast incidents and show historical accuracy. It attempts to get the best of both worlds by doing a forward-looking forecast to see how many incidents will be handled in the future, as well as "rolling historical forecasts". By restricting the data sets in the past we are able to re-forecast a historical month as if it were the last day of the previous month. 

We are then able to use that forecast to compare against the actual number of incidents in a given month. Combining these two strategies allows us to both observe the historical accuracy of our forecasting while using all currently available data for future forecasting.

Additionally, this PR adds the ability for us to forecast based on different subsets of data, e.g. how many incidents of type `foo` do we expect to see? how many incidents of priority `bar`, how many incidents of type `foo` AND priority `bar` and so on. 

See also #923 

@snkilmartin I had to remove the `box cot` fit to the model due to it having issues with the smaller filtered datasets, in my tests I didn't see that this had any effect. Please let me know if this will greatly affect forecasting.

### DIFF
--- a/src/dispatch/incident/metrics.py
+++ b/src/dispatch/incident/metrics.py
@@ -84,7 +84,6 @@ def make_forecast(incidents: List[Incident]):
             dataframe, seasonal_periods=12, trend="add", seasonal="add"
         ).fit()
     except Exception as e:
-        raise e
         log.warning(f"Issue forecasting incidents: {e}")
         return [], []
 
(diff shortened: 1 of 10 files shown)


mvilanova (approved):
LGTM

snkilmartin:
I reviewed our internal data with box_cox and without, and the forecast is a bit more stable without the transform so I'm fine with this change. Thanks for making all these updates; this was a much bigger change that I realized!

kevgliss:
Thanks for the review @snkilmartin and @mvilanova !

PR #944, opened 2021-03-15
Feature/organizations projects

kevgliss:
This work would allow us to create multiple projects each with their own unique dispatch world view. Meaning that incident_types, incidents, and all other configuration options would be unique on a per-project basis. 

This is a multi-tenant approach that gives each "project" its own schema and set of tables that is unique to that project. We will then route the users to the correct schema based on their request. e.g. `/api/v1/netflix/projectFoo` would route to a project in the Netflix organization, with its own unique dispatch setup. 

For details about the approach:

https://alembic.sqlalchemy.org/en/latest/cookbook.html#rudimental-schema-level-multi-tenancy-for-postgresql-databases 
https://github.com/sqlalchemy/alembic/issues/658
https://gist.github.com/nickretallack/bb8ca0e37829b4722dd1


mrthankyou:
Wow, this is a really interesting feature. Best of luck on this!

src/dispatch/alembic/versions/cdca3fcfc278_.py
```diff
@@ -0,0 +1,105 @@
+"""empty message
```
mvilanova, line 1: nit: add description

src/dispatch/organization/service.py
```diff
@@ -0,0 +1,44 @@
...
+    return db_session.query(Organization)
+
+
+def create(*, db_session, organization_in: OrganizationCreate) -> Organization:
+    team = Organization(
+        **organization_in.dict(),
+    )
+    db_session.add(team)
+    db_session.commit()
+    return team
```
mvilanova, line 22: I think team is supposed to be organization

src/dispatch/organization/service.py
```diff
@@ -0,0 +1,44 @@
...
+    for field in organization_data:
+        if field in update_data:
+            setattr(organization, field, update_data[field])
+
+    db_session.add(organization)
+    db_session.commit()
+    return organization
+
+
+def delete(*, db_session, organization_id: int):
```
mvilanova, line 41: Are we going to keep or delete users linked to the organization? Keep them and allow the admin to assign them to a new organization?

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@router.post("/", response_model=OrganizationRead)
+def create_organization(
+    *, db_session: Session = Depends(get_db), organization_in: OrganizationCreate
+):
+    """
+    Create a new organization.
+    """
+    organization = get_by_email(db_session=db_session, email=organization_in.name)
```
mvilanova, line 55: Should this be get_by_name instead? I don't see a column for email in the organization data model. Also get_by_email is not defined in the service module.

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+            status_code=400, detail="The organization with this name already exists."
+        )
+    organization = create(db_session=db_session, organization_in=organization_in)
+    return organization
+
+
+@router.get("/{organization_id}", response_model=OrganizationRead)
+def get_organization(*, db_session: Session = Depends(get_db), organization_id: int):
+    """
+    Get a organization.
```
mvilanova, line 67:
```suggestion
    Get an organization.
```

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    organization = get(db_session=db_session, organization_id=organization_id)
+    if not organization:
+        raise HTTPException(status_code=404, detail="The organization with this id does not exist.")
```
mvilanova, line 71:
```suggestion
        raise HTTPException(status_code=404, detail="An organization with this id does not exist.")
```

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+
+@router.put("/{organization_id}", response_model=OrganizationRead)
+def update_organization(
+    *,
+    db_session: Session = Depends(get_db),
+    organization_id: int,
+    organization_in: OrganizationUpdate,
+):
+    """
+    Update a organization contact.
```
mvilanova, line 83:
```suggestion
    Update an organization.
```

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    organization = get(db_session=db_session, organization_id=organization_id)
+    if not organization:
+        raise HTTPException(status_code=404, detail="The organization with this id does not exist.")
```
mvilanova, line 87:
```suggestion
        raise HTTPException(status_code=404, detail="An organization with this id does not exist.")
```

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+    organization = update(
+        db_session=db_session, organization=organization, organization_in=organization_in
+    )
+    return organization
+
+
+@router.delete("/{organization_id}", response_model=OrganizationRead)
+def delete_organization(*, db_session: Session = Depends(get_db), organization_id: int):
+    """
+    Delete a organization.
```
mvilanova, line 97:
```suggestion
    Delete an organization.
```

src/dispatch/organization/views.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    organization = get(db_session=db_session, organization_id=organization_id)
+    if not organization:
+        raise HTTPException(status_code=404, detail="The organization with this id does not exist.")
```
mvilanova, line 101:
```suggestion
        raise HTTPException(status_code=404, detail="An organization with this id does not exist.")
```

src/dispatch/organization/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+from sqlalchemy import Column, Integer, String
+from sqlalchemy.orm import relationship
+from sqlalchemy_utils import TSVectorType
+
+from dispatch.database import Base
+from dispatch.models import DispatchBase
+
+
+class Organization(Base):
```
mvilanova, line 11: Is the `ContactMixin` model missing here?

mvilanova (commented):
Some early feedback. Skipped the JS stuff because it feels it's still in early stages of dev.

src/dispatch/organization/models.py
kevgliss: I don't think so.

src/dispatch/organization/service.py
kevgliss: We could go either way. I don't think it matters too much. But I would guess if we were to delete an organization it would delete all of the user information as well.

src/dispatch/organization/models.py
mvilanova: Ok, the `get_by_email` in the views module led me to believe that it was perhaps missing.

src/dispatch/organization/service.py
mvilanova: That would be my preference too, but I also don't have a strong opinion.

kevgliss:
I'm going to try a different approach. Will open a separate PR.

PR #964, opened 2021-03-17
Replaces total cost with cost card

mvilanova:
Screen Shot 2021-03-18 at 12 09 00 PM


src/dispatch/static/dispatch/src/incident/Table.vue
```diff
@@ -124,8 +132,9 @@ export default {
         { text: "Status", value: "status" },
         { text: "Type", value: "incident_type.name" },
         { text: "Priority", value: "incident_priority.name", width: "10%" },
-        { text: "Total Cost", value: "total_cost" },
         { text: "Commander", value: "commander", sortable: false },
+        { text: "Reporter", value: "reporter", sortable: false },
```
mvilanova, line 136: Added the reporter back as looks like we may have removed it at some point.

kevgliss: I intentionally removed the reporter. It's not a common access pattern to go look for who reported or at least not on the same level as attempting to figure out the commander. We also don't promote the other (sometimes more important, than reporter) roles of scribe and liaison.

mvilanova: I like having the reporter, because it allows me to see the source of the report, like we do when we announce new incidents, etc.

kevgliss: Yes, but is that a common need from a historical perspective? The incident table/list is geared to inspection after the fact. Also, you can view that information by looking that the details, are you often scanning over all incidents for a reporter or just want to know the reporter for a particular incident?

mvilanova: Yes, I know it's in the details tab. When I search for an incident in Dispatch I like to see them both without having to click two times and wait for Dispatch to load the details tab. I think it's a reasonable expectation and it's consistent with other IR/ IM systems out there. I don't think we have data to determine whether it's a `common need from a historical perspective` at Netflix? We should poll our users if we really want to make a decision based on that.

kevgliss: I was basing 'common' on that the change was made several (https://github.com/Netflix/dispatch/pull/715/files#diff-6b587a3ecd8e4c7767110d428a841a4b25a9f3d8d15aa37dfb04e68d93b34f1dL120) months ago and has gone unnoticed until now.

mvilanova: Ok, l'll skip re-adding it until we have consensus.

PR #966, opened 2021-03-18
doc improvements

klarrio-dlamb:
Some improvements from my following of the docs so far to set-up.
Tried converting the Slack commands & their vars to a table for delineation and easier cut & paste into Slack, but perhaps not the ideal format.


kevgliss:
This looks reasonable to me. Unfortunately, slack doesn't provide an API to allow us to have these added automatically :-(. Thanks for the PR!

PR #979, opened 2021-03-19
Adds ability to select the FAIR Form of Loss category for incident cost types

mvilanova:
More information about FAIR Form of Losses here: https://www.fairinstitute.org/blog/a-crash-course-on-capturing-loss-magnitude-with-the-fair-model

Screen Shot 2021-03-19 at 2 04 09 PM


src/dispatch/alembic/versions/b6da2dad0396_.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "b6da2dad0396"
+down_revision = "a769ea6bad64"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
kevgliss, line 23: Would it be easier to just set a default for this column?

kevgliss (approved):

src/dispatch/alembic/versions/b6da2dad0396_.py
mvilanova: I thought about adding a default, but decided not to to motivate users to think under which FAIR FoL phase they need to bucket the costs types they add.

PR #995, opened 2021-03-26
Consolidates restricted visibility-based filtering

mvilanova:


src/dispatch/database/base.py
```diff
@@ -0,0 +1,66 @@
+import re
```
kevgliss, line 1: nit: we seem to be using "base" a lot e.g. database.py import Base class now it's database.base import Base, this module might better named "core" or similar.

src/dispatch/database/service.py
```diff
@@ -222,10 +153,26 @@ def search_filter_sort_paginate(
     else:
         query = db_session.query(model_cls)
 
+    if user_role != UserRoles.admin.value:
+        if model.lower() == "incident":
```
kevgliss, line 157: Do we need to do something similar for tasks for restricted incidents?

kevgliss (approved):

src/dispatch/database/base.py
mvilanova: Yeah, sure. Honestly, I did not spend a lot of time thinking about it.

src/dispatch/database/service.py
mvilanova: I was going to add logic for it, but when I started testing tasks I realized they were already being filtered. I did some research to try to figure out the _why_ and the conclusion that I reached was that when we load `/tasks` we perform a GET against the `/api/v1/incidents/` endpoint, which returns a list of filtered incidents, and then sqlalchemy filters out the tasks based on that list.

PR #1023, opened 2021-04-12
Improvements to organizations and projects

mvilanova:


src/dispatch/alembic/versions/7ea14a81dc59_.py
```diff
@@ -115,14 +114,6 @@ def upgrade():
     # Create organization table
     Organization.__table__.create(bind)
 
-    # op.create_index(
```
kevgliss, line 118: I think we need this, it was just causing an error locally for me.

kevgliss (changes requested):

PR #1054, opened 2021-04-20
Allows the reporter/commander to be changed 

kevgliss:
Allows the reporter/commander to be changed even when no conversation is associated.

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -396,6 +396,10 @@ def set_conversation_topic(incident: Incident, db_session: SessionLocal):
         db_session=db_session, project_id=incident.project.id, plugin_type="conversation"
     )
 
+    if not incident.conversation:
+        log.warning("Conversation topic not sent because incident has no conversation.")
+        return
+
     try:
         plugin.instance.set_topic(incident.conversation.channel_id, conversation_topic)
     except Exception as e:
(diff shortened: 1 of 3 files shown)


src/dispatch/incident/flows.py
```diff
@@ -396,6 +396,10 @@ def set_conversation_topic(incident: Incident, db_session: SessionLocal):
         db_session=db_session, project_id=incident.project.id, plugin_type="conversation"
     )
 
+    if not incident.conversation:
+        log.warning("Conversation topic not sent because incident has no conversation.")
```
mvilanova, line 400:
```suggestion
        log.warning("Conversation topic not set because incident has no conversation.")
```

src/dispatch/incident/flows.py
```diff
@@ -396,6 +396,10 @@ def set_conversation_topic(incident: Incident, db_session: SessionLocal):
         db_session=db_session, project_id=incident.project.id, plugin_type="conversation"
     )
 
+    if not incident.conversation:
```
mvilanova, line 399: I'd move this check to the beginning to speed up the execution in case we don't have a conversation.

src/dispatch/incident/flows.py
mvilanova: Looks like you forgot this one before merging.

PR #1085, opened 2021-04-26
Fixing issue with incident filter

kevgliss:


src/dispatch/database/service.py
```diff
@@ -29,18 +29,18 @@
...
         .join(IndividualContact)
         .filter(
-            not_(
-                and_(
-                    Incident.visibility == Visibility.restricted.value,
-                    IndividualContact.email != current_user.email,
-                )
+            or_(
+                Incident.visibility == Visibility.open.value,
+                IndividualContact.email == current_user.email,
```
mvilanova, line 38: They two expressions are equivalent, but the OR one is easier to read.

mvilanova (approved):

PR #1093, opened 2021-04-27
Allows the creation of oncall services with same external id across projects

mvilanova:


kevgliss:
Do we want to add a constraint on the external_id and project_id? Like we do here:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident_type/models.py#L20

mvilanova:
Yes, I haven't gotten to it yet.

PR #1100, opened 2021-04-28
Bugfixes and improvements

mvilanova:


src/dispatch/static/dispatch/src/plugin/PluginInstanceCombobox.vue
```diff
@@ -2,7 +2,7 @@
   <v-combobox
     v-model="plugin"
     :items="items"
-    item-text="plugin.slug"
+    item-text="slug"
```
kevgliss, line 5: This is a bigger issue; essentially sometimes the model has "plugin.slug" but in other contexts it only has slug. (like when loading existing data vs selecting a new one). 

I've moved between both of these but we need to fix the way the underlying metadata is stored to have it unified.

mvilanova: Hmmm ok. I'll look into it.

PR #1070, opened 2021-04-21
Use incident search filters for engagement

kevgliss:

PR #1114, opened 2021-04-29
Bugfix/adding participants to closed incidents

kevgliss:


src/dispatch/incident/flows.py
```diff
@@ -1057,20 +1057,24 @@ def incident_update_flow(
...
-    # we add the team distributions lists to the notifications group
-    group_plugin = plugin_service.get_active_instance(
-        db_session=db_session, project_id=incident.project.id, plugin_type="participant-group"
-    )
-    if group_plugin:
-        team_participant_emails = [x.email for x in team_participants]
-        group_plugin.instance.add(incident.notifications_group.email, team_participant_emails)
+    # TODO determine how to deduplicate service participants.
+    # don't try to add new participants on close
+    # if incident.status != IncidentStatus.closed:
```
mvilanova, line 1063: We were missing this check. The incident was closed and channel archived when we were trying to add participants to it.

kevgliss: Yes we will need that change regardless. But I don't know that we've figured out why a service was getting re-resolved. #1070 may help as well.

mvilanova: We're either resolving it and not saving it in participant or not resolving it correctly and saving null in the database (the column service_id is null for all participants). If #1070 is going to change the way we recommend, then I'll stop debugging until we're done with it.

PR #1132, opened 2021-05-04
Adding responses and default error message for incident modals 

kevgliss:
Ensures we send confirmation and error messages when a user submits a modal.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/modals.py
+++ b/src/dispatch/plugins/dispatch_slack/modals.py
@@ -1120,9 +1120,8 @@ def rating_feedback_from_submitted_form(
     db_session.add(incident)
     db_session.commit()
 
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/dispatch_slack/modals.py
```diff
@@ -1082,3 +1119,10 @@ def rating_feedback_from_submitted_form(action: dict, db_session=None):
 
     db_session.add(incident)
     db_session.commit()
+
+    dispatch_slack_service.send_ephemeral_message(
+        client=slack_client,
+        conversation_id=channel_id,
+        user_id=user_id,
+        text="Thank you for your feedback!",
+    )
```
mvilanova, line 1128: This needs to be a DM not an ephemeral message.

mvilanova (changes requested):

PR #1179, opened 2021-05-12
Improves participant management

mvilanova:
This PR attempts to improve the management of incident participants in the following scenarios:
* Participants are added to the incident
* Participants leave the incident and if they decide to join again later
* The incident is closed and later re-opened as active or stable


src/dispatch/plugins/dispatch_slack/actions.py
```diff
@@ -150,17 +150,24 @@ def handle_engage_oncall_action(
...
 
-    if not oncall_service or not oncall_individual:
+    if not oncall_individual and not oncall_service:
         message = "Could not engage oncall. Oncall service plugin not enabled."
-    else:
-        message = f"You have successfully engaged {oncall_individual.name} from oncall rotation {oncall_service.name}."
+
+    if not oncall_individual and oncall_service:
+        message = (
+            f"The oncall for the {oncall_service.name} rotation is already in the conversation."
```
kevgliss, line 165:
```suggestion
            f"A member of {oncall_service.name} is already in the conversation."
```

src/dispatch/participant_role/service.py
```diff
@@ -21,9 +20,20 @@ def get(*, db_session, participant_role_id: int) -> Optional[ParticipantRole]:
...
-def get_all(*, db_session):
-    """Returns all participant roles."""
-    return db_session.query(ParticipantRole)
+def get_last_active_role(
+    *,
+    db_session,
+    participant_id: int,
+) -> Optional[ParticipantRole]:
+    """
+    Returns the last active participant role.
```
kevgliss, line 29:
```suggestion
    Returns the participant's last active role.
```

src/dispatch/participant/flows.py
```diff
@@ -111,44 +117,34 @@ def remove_participant(user_email: str, incident: Incident, db_session: SessionL
...
         return False
 
-    # We mark the participant as active
-    participant.is_active = True
+    log.debug(f"Reactivating {participant.individual.name} on {incident.name} incident...")
 
-    # We create a role for the participant
-    participant_role_in = ParticipantRoleCreate(role=ParticipantRoleType.participant)
+    # we get the last active role
+    participant_role = participant_role_service.get_last_active_role(
```
kevgliss, line 131: I'm not sure this is what we want, if their last role was as incident commander and they leave and are re-added, I wouldn't want to give them that role again. I would imagine it would be simpler to just give them a participant role whenever they are reactivated.

src/dispatch/incident/flows.py
```diff
@@ -97,6 +97,34 @@ def get_incident_participants(incident: Incident, db_session: SessionLocal):
...
+    for participant in incident.participants:
+        participant_flows.inactivate_participant(participant.individual.email, incident, db_session)
+
+    event_service.log(
+        db_session=db_session,
+        source="Dispatch Core App",
+        description="Incident participants inactivated",
+        incident_id=incident.id,
+    )
+
```
kevgliss, line 126: It seems like these are pretty redundant without much actual logic, can they be combined and have a flag that states whether we should activate or inactivate participants?

kevgliss (approved):
This looks good, stylistically you have several small functions that contain very little reusable logic that could be combined to make things easier to follow.

src/dispatch/incident/flows.py
mvilanova: Yes, I could combine them. I felt it would be easier to maintain if I kept them separate and small.

src/dispatch/participant/flows.py
mvilanova: We don't allow incident commanders to leave the channel without first doing a handoff.

PR #1221, opened 2021-05-24
chore: improved GitHub actions

### LINKED ISSUE #1207: Add black as a separate github action instead of running it under build workflow
**Is your feature request related to a problem? Please describe.**
Currently, the python.yml runs black under the `build job`, instead we should run black as a separate action and build dependent on it, so if black is successful then only build will be trigged and if it is not then build should be skipped.

vj-codes:
Fixes #1207 
1) Added black as a separate job
2) Made build dependent on lint so if lint fails, build job will be skipped
Thus, improved the use of GitHub actions in the repository

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -3,7 +3,15 @@ name: Python Lint and Test
 on: pull_request
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: psf/black@stable
+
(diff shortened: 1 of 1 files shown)


vj-codes:
@kevgliss hey can you review this, please
and the workflow run should be approved by a maintainer as I'm contributing for the first time here

vj-codes:
Oops seems like black failed, will reformat the code and update. 
But this is a good example of showing that if black fails build is skipped

kevgliss:
This looks good to me, we will fixup the black stuff in other PRs. Thanks!

PR #1256, opened 2021-06-01
Bugfix/incident visibility

kevgliss:


src/dispatch/plugins/dispatch_slack/modals/incident/handlers.py
```diff
@@ -181,6 +181,10 @@ def update_incident_from_submitted_form(
 
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
     existing_incident = IncidentRead.from_orm(incident)
+
+    # we don't allow visibility to be set in slack so we copy it over
+    incident_in.visibility = incident.visibility
```
mvilanova, line 186:
```suggestion
    incident_in.visibility = incident.visibility.value
```

kevgliss: It's not an enum it's just a regular object attr.

PR #1289, opened 2021-06-07
Fixes add participant timeline event description not showing role

mvilanova:


src/dispatch/incident/flows.py
```diff
@@ -1267,7 +1267,6 @@ def incident_add_or_reactivate_participant_flow(
     user_email: str,
     incident_id: int,
     service_id: int = 0,
-    role: ParticipantRoleType = None,
```
mvilanova, line 1270: @kevgliss I removed `role` since we're not passing it in any of the calls to this function

PR #1296, opened 2021-06-08
Setting the access token correctly for pcke login flow.

kevgliss:


bender92:
Thank you, it is working, but we use idToken instead accessToken

PR #1298, opened 2021-06-09
Fixed a command issue and added clarification

dilbwagsingh:
Without adding the project name it was giving a Traceback error - 
project name -> 
   none is not an allowed value (type=type_error.none.not_allowed)


kevgliss:
Thanks for the PR!

PR #1378, opened 2021-06-30
Silence events from non-incident channels

kevgliss:


src/dispatch/plugins/dispatch_slack/events.py
```diff
@@ -99,6 +102,11 @@ async def handle_slack_event(*, client, event, background_tasks):
 
     if user_id and channel_id:
         db_session = get_organization_from_channel_id(channel_id=channel_id)
```
mvilanova, line 104: Should we rename this function to `get_organization_scoped_db_session()` or something similar to make it more obvious what it is returning? Also, should it live in the service module under organization?

kevgliss: Only slack needs to convert channel_id to an organization.

mvilanova: Ok, sure. How about a new module called database under the slack plugin directory then? I find it confusing that it leaves in the decorators one.

kevgliss: We could move it... I'd rather not have a module with one function in it though. I think originally it was in the decorators function because that was the only module that used it but I needed it in events as well.

PR #1387, opened 2021-07-01
Only get users that have permission

kevgliss:
This addresses an issue with Jira server where we may retrieve a user that doesn't have access to the given project (e.g. multiple users for a given email).

### DIFF
--- a/src/dispatch/plugins/dispatch_jira/plugin.py
+++ b/src/dispatch/plugins/dispatch_jira/plugin.py
@@ -50,10 +50,12 @@ def get_email_username(email: str) -> str:
     return email
 
 
-def get_user_field(client: JIRA, user_email) -> dict:
+def get_user_field(client: JIRA, user_email: str, project_key: str) -> dict:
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -173,13 +175,15 @@ def create(
...
         client = JIRA(str(JIRA_API_URL), basic_auth=(JIRA_USERNAME, str(JIRA_PASSWORD)))
 
-        assignee = get_user_field(client, commander_email)
-        reporter = get_user_field(client, reporter_email)
-
         project_id, issue_type_name = process_incident_type_plugin_metadata(
             incident_type_plugin_metadata
         )
 
+        # should we move to project keys?
```
mvilanova, line 182: IIRC, we moved to project id because Jira cloud did not support project key.
```suggestion
```

mvilanova (approved):

PR #1405, opened 2021-07-07
Improves participant management related to tasks

mvilanova:
We only add the assignees of a task to the incident if the status of the task is open and we ask incident participants to resolve or re-assign their incident tasks before leaving the incident channel.

Screen Shot 2021-07-09 at 2 49 42 PM


src/dispatch/task/flows.py
```diff
@@ -112,8 +113,13 @@ def create_or_update_task(
     )
 
     if existing_task:
-        # save the status before we attempt to update the record
+        # we don't attempt to update a task if both the existing and current tasks are resolved
+        if existing_task.status == TaskStatus.resolved and task["status"] == TaskStatus.resolved:
```
mvilanova, line 117: Task could be re-opened, updated, and re-resolved in between syncs.

PR #1482, opened 2021-07-26
Workflow and WorkflowInstance CRUD tests

mvilanova:


src/dispatch/workflow/enums.py
```diff
@@ -0,0 +1,9 @@
+from dispatch.enums import DispatchEnum
+
+
+class WorkflowInstanceStatus(DispatchEnum):
```
mvilanova, line 4: @kevgliss should the values be in capital letters?

kevgliss: They could be... I believe they get rendered as "Workflow foo is {status}".

mvilanova: Ok. I'll convert them to capital, so we're consistent with other enums and messaging.

PR #1497, opened 2021-07-29
Adds service update tests

mvilanova:


kevgliss:
This looks good. I have a plan to fix the feedback tests here as it's a bit more universal. e.g. we need a standardized way of checking for missing/invalid related objects. We do or don't do this on a one-off basis today.

kevgliss (approved):

PR #1502, opened 2021-07-29
Improves sub object input validation

kevgliss:


src/dispatch/auth/views.py
```diff
@@ -134,32 +122,21 @@ def login_user(
...
+        [
+            ErrorWrapper(
+                InvalidValueError(msg="Invalid username."),
+                loc="username",
+            ),
+            ErrorWrapper(
+                InvalidValueError(msg="Invalid password."),
+                loc="password",
+            ),
+        ]
```
mvilanova, line 139: My preference would be to say something like "You have entered an invalid username or password" instead of letting a potential malicious user know which one they got wrong.

src/dispatch/incident_priority/service.py
```diff
@@ -26,6 +33,22 @@ def get_default(*, db_session, project_id: int):
...
 
+def get_default_or_raise(*, db_session, project_id: int) -> IncidentPriority:
+    """Returns the default incident_priority or raise a ValidationError if one doesn't exist."""
+    incident_priority = get_default(db_session=db_session, project_id=project_id)
+
+    if not incident_priority:
+        raise ValidationError(
+            [
+                ErrorWrapper(
+                    NotFoundError(msg="No default incident_priority defined."),
```
mvilanova, line 44: We may want to make these messages a little bit more human readable. In this case, replacing `_` with a space.

src/dispatch/individual/views.py
```diff
@@ -30,29 +32,14 @@ def create_individual(
...
-                {
-                    "msg": str(e),
-                    "loc": ["Unknown"],
-                    "type": "Unknown",
-                }
-            ],
+        raise ValidationError(
+            [
+                ErrorWrapper(
+                    ExistsError(msg="A individual with this email already exists."), loc="email"
```
mvilanova, line 38:
```suggestion
                    ExistsError(msg="An individual with this email already exists."), loc="email"
```

src/dispatch/organization/views.py
```diff
@@ -48,15 +50,12 @@ def create_organization(
...
-                {
-                    "msg": "An organization with this name already exists.",
-                    "loc": ["name"],
-                    "type": "Exists",
-                }
-            ],
+        raise ValidationError(
+            [
+                ErrorWrapper(
+                    ExistsError(msg="A organization with this name already exists."), loc="name"
```
mvilanova, line 56:
```suggestion
                    ExistsError(msg="An organization with this name already exists."), loc="name"
```

src/dispatch/organization/views.py
```diff
@@ -102,15 +101,12 @@ def update_organization(
...
```
mvilanova, line 107:
```suggestion
                    ExistsError(msg="An organization with this name already exists."), loc="name"
```

src/dispatch/search_filter/views.py
```diff
@@ -62,15 +61,12 @@ def update_search_filter(
...
-                {
-                    "msg": "An search filter with this name already exists.",
-                    "loc": ["name"],
-                    "type": "Exists",
-                }
-            ],
+        raise ValidationError(
+            [
+                ErrorWrapper(
+                    ExistsError(msg="A sarch filter with this name already exists."), loc="name"
```
mvilanova, line 67:
```suggestion
                    ExistsError(msg="A search filter with this name already exists."), loc="name"
```

src/dispatch/auth/views.py
kevgliss: The error returns both, regardless of which was actually the issue.

mvilanova: So, for example, if login auth fails we show "Invalid username. Invalid password" in the UI?

kevgliss: The API will return both, Yes. The error response is a list. We haven't traditionally displayed those very well (or at all in some cases). My hope is that with this change the error response will be more uniform across the API endpoints and we can display them better in the UI.

PR #1509, opened 2021-08-01
Unprocessable entity on any API request due to insufficient parameters

stefanm8:
Unprocessable entity on any API request due to insufficient parameters supplied. Thus raising a pydantic exception which being in middleware will be handled as a 422 response


kevgliss:
lgtm, thanks!

PR #1533, opened 2021-08-05
Allow build time VUE_APP variables

ymatsiuk:
As an alternative to `.env` file when building releases using `docker-compose` this will be useful


kevgliss:
lgtm, I'm not sure why those checks are failing but I don't think it matters for this PR.

PR #1535, opened 2021-08-05
Addresses some sticky circular imports with our pydantic models

kevgliss:


src/dispatch/search/fulltext/__init__.py
```diff
@@ -401,6 +401,7 @@ def hstore_vectorizer(column):
     ]
     for class_ in classes:
         sql = class_(**params)
+        print(sql)
```
kevgliss, line 404:
```suggestion
```

PR #1541, opened 2021-08-10
Queries Project API if project parameter is included in URL

mvilanova:
This PR ensures we will pass the IncidentCreate -> ProjectRead Pydantic validation when a new incident is reported using the reporting form by querying the Project API and setting the id required by the ProjectRead model.


mvilanova:
I thought this would be a better approach than making `id` in the [ProjectRead](https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident/models.py#L259)  optional.

mvilanova:
@kevgliss and I decided that it's best to make `IncidentCreate -> ProjectRead -> id` optional, so I'm going to revert these changes and make it optional.

PR #1552, opened 2021-08-13
Fixes issue where 'stable_at' wasn't being set from the UI.

kevgliss:


mvilanova (approved):
And us thinking that making it more explicit would be better...

PR #1560, opened 2021-08-18
Update "First User" instructions in documentation

themullinator:
Following the existing instructions, I get the error `error: no such option: --project`


mvilanova (approved):

Issue #1566, opened 2021-08-19
Not able to modify default organization or create new one


vikassinha2019:
**Describe the bug**
Admin user of Dispatch is not able to modify the name of the Organization which is set to 'default' in the default Dispatch installation. Failing with 403 forbidden error.

Creating a new Org also fails with following error:

```
INFO:     172.26.0.1:51314 - "POST /api/v1/organizations HTTP/1.0" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 373, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 26, in __call__
    await response(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/responses.py", line 224, in __call__
    await run_until_first_complete(
  File "/usr/local/lib/python3.8/site-packages/starlette/concurrency.py", line 24, in run_until_first_complete
    [task.result() for task in done]
  File "/usr/local/lib/python3.8/site-packages/starlette/concurrency.py", line 24, in <listcomp>
    [task.result() for task in done]
  File "/usr/local/lib/python3.8/site-packages/starlette/responses.py", line 216, in stream_response
    async for chunk in self.body_iterator:
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 56, in body_stream
    task.result()
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 38, in coro
    await self.app(scope, receive, send)
[...]

PR #1567, opened 2021-08-20
Fixes issue with trigger function on organization creation

mvilanova:
Fixes https://github.com/Netflix/dispatch/issues/1566

```
sqlalchemy.exc.InternalError: (psycopg2.errors.DependentObjectsStillExist) cannot drop function project_search_vector_update() because other objects depend on it
DETAIL:  trigger project_search_vector_trigger on table dispatch_organization_default.project depends on function project_search_vector_update()
HINT:  Use DROP ... CASCADE to drop the dependent objects too.

[SQL: DROP FUNCTION IF EXISTS project_search_vector_update()]
```


mvilanova:
Function `project_search_vector_update()` is defined in the `public` schema and triggers under org schemas are pointing to it, hence the error when we're creating a new org and trying to drop the function.

I don't think the suggestion of adding `CASCADE` at the end of the SQL statement [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/search/fulltext/__init__.py#L193) is a good idea, because it mess with the other triggers pointing to it.

I think we have two options 1) we copy the trigger functions that live in the `public` schema that triggers point to to every org schema, update the triggers and delete the trigger functions from the public schema or 2) we disable dropping and recreating trigger functions on schema initialization and keep making triggers point to the trigger functions in the `public` schema.

kevgliss:
> Trigger function project_search_vector_update() is defined in the public schema and triggers under org schemas are pointing to it, hence the error when we're creating a new org and trying to drop the function.

I don't think this is true for all cases (maybe just in your local instance?). By public did you mean `default`? 

I don't think this is the best way to solve this particular issue. Ideally, we don't want functions to cross the schema boundary. 

Instead, I think we should make the function on a per-schema basis it looks like the create function allows us to prefix the function name with a prefix:

https://www.postgresql.org/docs/9.5/sql-createfunction.html

So we should probably modify the function here to accept a schema:
https://github.com/Netflix/dispatch/blob/master/src/dispatch/search/fulltext/__init__.py#L149

mvilanova:
> I don't think this is true for all cases (maybe just in your local instance?). By public did you mean default?

No, I meant `public`. It doesn't seem to happen on our `prod` env, but it's on the `test` one.

kevgliss (approved):

PR #1633, opened 2021-09-21
Enable Slack Public Conversation Channels

jeffreynn:
Add a field to the config which would allow users to create public conversation channels as opposed to only private ones.


mvilanova:
Thanks for the PR. @kevgliss is working on moving plugin configuration to the Web UI in #1619, so I think it would make more sense to include your suggested changes in his PR.

jeffreynn:
Ok will do. I'll close this PR. Thanks

kevgliss:
Yes, I will add this to my PR.

PR #1638, opened 2021-09-22
Support windows file path & env vars for CLI serve command

axellaurelut:
Hello,

Currently, running `dispatch server develop` fails on windows because of the static file paths.
Also the front-end will never be served  because the env var `STATIC_DIR` has a default.

### DIFF
--- a/src/dispatch/cli.py
+++ b/src/dispatch/cli.py
@@ -704,11 +704,13 @@ def run_server(log_level):
         # take our frontend vars and export them for the frontend to consume
         envvars = os.environ.copy()
         envvars.update({x: getattr(config, x) for x in dir(config) if x.startswith("VUE_APP_")})
-
+        is_windows = os.name == "nt"
+        windows_cmds = ["cmd", "/c"]
+        default_cmds = ["npm", "run", "serve"]
+        cmds = windows_cmds + default_cmds if is_windows else default_cmds
         p = Popen(
-            ["npm", "run", "serve"],
+            cmds,
             cwd=os.path.join("src", "dispatch", "static", "dispatch"),
-            shell=True,
             env=envvars,
         )
         atexit.register(p.terminate)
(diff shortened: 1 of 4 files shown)


mvilanova:
Thanks for the PR. I'll test the changes locally and merge them if everything is working as expected.

src/dispatch/cli.py
```diff
@@ -697,15 +697,20 @@ def run_server(log_level):
...
 
         # take our frontend vars and export them for the frontend to consume
         envvars = os.environ.copy()
         envvars.update({x: getattr(config, x) for x in dir(config) if x.startswith("VUE_APP_")})
 
-        p = Popen(["npm", "run", "serve"], cwd="src/dispatch/static/dispatch", env=envvars)
+        p = Popen(
+            ["npm", "run", "serve"],
+            cwd=os.path.join("src", "dispatch", "static", "dispatch"),
+            shell=True,
```
mvilanova, line 711: Adding `shell=True` breaks my local dev env. It should be working without it in Windows according to the [Popen docs](https://docs.python.org/3/library/subprocess.html#popen-constructor):
> On Windows with shell=True, the COMSPEC environment variable specifies the default shell. The only time you need to specify shell=True on Windows is when the command you wish to execute is built into the shell (e.g. dir or copy). You do not need shell=True to run a batch file or console-based executable.

Also, there are [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations) that need to be taken into account when using `shell=True`.

Can you try without it and let me know if it works for you?
```suggestion
```

mvilanova (changes requested):

axellaurelut:
Removing `shell=True` results on a `FileNotFoundError: [WinError 2] The system cannot find the file specified` error. Looks like a known issue (https://bugs.python.org/issue20927).

There is a work around though (https://stackoverflow.com/questions/49018413/filenotfounderror-subprocess-popendir-windows-7/49018557), I updated the PR, let me know if that is OK for you.

mvilanova (approved):
Changes look good.

PR #1639, opened 2021-09-22
Fix table name for incident_role_incident_priority in db migration

mnuck-fp:
The name of the table is singular `incident_role_incident_priority`, not plural `priorities`. The migration fails without this fix. Check the names of the variables and table for the `incident_role_incident_type` table for a working example.


mvilanova:
Thanks for the fix!

PR #1643, opened 2021-09-23
Fixes incident type plugin metadata UI

kevgliss:


mvilanova:
Pulled the PR locally. Getting `none is not an allowed value` when trying to save. Also, the `+` button to add an item does not work.

Screen Shot 2021-09-23 at 3 24 02 PM

mvilanova:
Looks like the `PluginMetadata` Pydantic model expects `slug` but we're not providing it:
```
{"detail":[{"loc":["body","plugin_metadata",0,"slug"],"msg":"none is not an allowed value","type":"type_error.none.not_allowed"}]}
```

Issue #1729, opened 2021-10-14
Slack channel creation fails


jtorvald:
**Describe the bug**
When creating a new incident I get the following error in the logs:

The channel creation fails and that response does not seems to be checked because the following actions also fail because of a missing channel_id
```

web_1        | INFO:     172.20.0.1:40684 - "POST /api/v1/default/incidents HTTP/1.1" 200 OK
web_1        | INFO:Dispatch Core App: Creation of incident conversation failed. Reason: 'NoneType' object has no attribute 'lower':/usr/local/lib/python3.9/site-packages/dispatch/event/service.py:log:100
web_1        | ERROR:'NoneType' object has no attribute 'lower':/usr/local/lib/python3.9/site-packages/dispatch/incident/flows.py:incident_create_flow:716
web_1        | Traceback (most recent call last):
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/incident/flows.py", line 688, in incident_create_flow
web_1        |     conversation = create_conversation(incident, db_session)
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/incident/flows.py", line 379, in create_conversation
web_1        |     conversation = plugin.instance.create(incident.name)
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/decorators.py", line 122, in wrapper
web_1        |     return func(*args, **kwargs)
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/decorators.py", line 106, in wrapper
web_1        |     result = func(*args, **kwargs)
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/plugins/dispatch_slack/plugin.py", line 69, in create
web_1        |     return create_conversation(client, name, self.configuration.private_channels)
web_1        |   File "/usr/local/lib/python3.9/site-packages/dispatch/plugins/dispatch_slack/service.py", line 270, in create_conversation
web_1        |     name=name.lower(),  # slack disallows upperCase
web_1        | AttributeError: 'NoneType' object has no attribute 'lower'
web_1        | WARNING:External ticket not updated. No ticket plugin enabled.:/usr/local/lib/python3.9/site-packages/dispatch/incident/flows.py:update_external_incident_ticket:177
[...]

PR #1747, opened 2021-10-20
Removed whitespace in default UI URL

jtorvald:
This one was causing issues with TicketCreate Weblink validation see #1729


mvilanova:

PR #1793, opened 2021-11-08
Adds a subscribe button to incident notifications

kevgliss:


vikassinha2019:
@kevgliss can you please provide a screenshot where this Subscribe button will show up. 
Also will it work for newly created incidents after upgrading Dispatch to v20211116, or will it work on open incidents from old version as well ?

kevgliss:
@vikassinha2019 this button will only show up for _new_ messages (created after that push is deployed). It should look something like:
Screen Shot 2021-12-03 at 9 40 48 AM

PR #1868, opened 2021-12-03
Enhances the performance of the incident data model

kevgliss:


src/dispatch/incident/models.py
```diff
@@ -209,18 +209,17 @@ def participant_observer(self, participants):
         self.participants_location = Counter(p.location for p in participants).most_common(1)[0][0]
 
         commanders_locations = []
+        reporters_locations = []
         for p in participants:
             for pr in p.participant_roles:
                 if pr.role == ParticipantRoleType.incident_commander:
                     commanders_locations.append(p.location)
+                if pr.role == ParticipantRoleType.reporter:
+                    reporters_locations.append(p.location)
```
kevgliss, line 218: Shouldn't this only ever be one location? Why is it a list, the same could be said for commander location?

PR #1899, opened 2022-01-13
Allows to define database engine pool size and max overflow values

mvilanova:


kevgliss (approved):

src/dispatch/config.py
```diff
@@ -175,6 +173,8 @@ def __str__(self) -> str:
 _QUOTED_DATABASE_PASSWORD = parse.quote(str(_DATABASE_CREDENTIAL_PASSWORD))
 DATABASE_NAME = config("DATABASE_NAME", default="dispatch")
 DATABASE_PORT = config("DATABASE_PORT", default="5432")
+DATABASE_ENGINE_POOL_SIZE = config("DATABASE_ENGINE_POOL_SIZE", cast=int, default=1000)
```
kevgliss, line 176: This might be a bit high for folks with smaller databases. It should probably be 20 and we can override it for our database setup.

mvilanova: Will update.

Issue #1922, opened 2022-01-25
Change database schema name after renaming an organization on the UI 


olam-slack:
- Currently, changing an organization will not lead to schema change in the database - be it the default organization or a newly created one. 
- This would lead to "organization not found error", when one tries to change the project name afterwards. 
- A current alternative is to create a new organization, instead of changing the name of an existing organization.

PR #1929, opened 2022-01-26
Adds support for renaming the org db schema on org name update

mvilanova:
Closes https://github.com/Netflix/dispatch/issues/1922


mvilanova:
Abandoning this one out. I'll cut a new PR to disallow org name changes, as changing the org db schema name is causing all sorts of issues with existing db sessions, as expected.

PR #1935, opened 2022-01-26
Disables the name field when editing an organization

### LINKED ISSUE #1922: Change database schema name after renaming an organization on the UI

mvilanova:
Closes #1922
Screen Shot 2022-01-26 at 2 12 59 PM
Screen Shot 2022-01-26 at 2 13 15 PM

### DIFF
--- a/src/dispatch/organization/service.py
+++ b/src/dispatch/organization/service.py
@@ -131,7 +131,8 @@ def update(
     """Updates an organization."""
     organization_data = organization.dict()
 
-    update_data = organization_in.dict(skip_defaults=True, exclude={"banner_color"})
+    # we exclude the organization's name because the organization db schema is named after it
+    update_data = organization_in.dict(skip_defaults=True, exclude={"name", "banner_color"})
 
     for field in organization_data:
         if field in update_data:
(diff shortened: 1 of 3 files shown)


kevgliss (changes requested):
Should we disable this in the API as well?

mvilanova:
Yes.

src/dispatch/organization/service.py
```diff
@@ -131,7 +131,8 @@ def update(
     """Updates an organization."""
     organization_data = organization.dict()
 
-    update_data = organization_in.dict(skip_defaults=True, exclude={"banner_color"})
+    # we exclude the organization's name because the organization db schema is named after it
+    update_data = organization_in.dict(skip_defaults=True, exclude={"name", "banner_color"})
```
kevgliss, line 135: I think we should remove organization "name" from the update model here: https://github.com/Netflix/dispatch/blob/master/src/dispatch/organization/models.py#L57

Otherwise, the swagger documentation will say that the field is updateable but we will simply ignore it we would be confusing.

mvilanova: Ah, good point. Didn't think about that. Will revert the changes on the last commit and update the Pydantic model.

PR #1962, opened 2022-02-08
Update our estimate of yearly business hours.

forestmonster:
We now assume each employee will work 1640 effective hours per year.


jschroth:
1. I think we should explain more about our logic around this. (we expect the the average person to work 6.5 hrs/day = 32.5/wk = 1690/yr)
2. I think this should have been `1690` but if there is a reason we are saying 50hrs less, that would be good to explain.

forestmonster:
> 1. I think we should explain more about our logic around this. (we expect the the average person to work 6.5 hrs/day = 32.5/wk = 1690/yr)

We have an open loop on this. I'll relay the information back here when closed.

> 2. I think this should have been `1690` but if there is a reason we are saying 50hrs less, that would be good to explain.

I believe our current understanding is 8 hours/day * 205 work days/year  = 1640 hours/year.

src/dispatch/project/models.py
```diff
@@ -22,7 +22,7 @@ class Project(Base):
     color = Column(String)
 
     annual_employee_cost = Column(Integer, default=650000)
```
mvilanova, line 24:
```suggestion
    annual_employee_cost = Column(Integer, default=50000)
```

src/dispatch/project/models.py
```diff
@@ -22,7 +22,7 @@ class Project(Base):
...
-    business_year_hours = Column(Integer, default=2080)
+    business_year_hours = Column(Integer, default=1640)
```
mvilanova, line 25: I'd suggest we leave the default to 2080, as this is the standard, and set 1640 to all our projects' settings in the Web UI. Dispatch gets these values from the project (see [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident_cost/service.py#L187))

src/dispatch/static/dispatch/src/project/NewEditSheet.vue
```diff
@@ -93,7 +93,7 @@
                     type="number"
                     min="1"
                     pattern="\d+"
-                    placeholder="2080"
+                    placeholder="1640"
```
mvilanova, line 96:
```suggestion
                    placeholder="2080"
```

mvilanova: We should also update the hint `hint="Number of hours to average employee host."`. We could say something along the lines of "Number of working hours in a year to calculate hourly rate."

src/dispatch/static/dispatch/src/project/store.js
```diff
@@ -12,7 +12,7 @@ const getDefaultSelectedState = () => {
     color: null,
     loading: false,
     annual_employee_cost: 50000,
-    business_year_hours: 2080,
+    business_year_hours: 1640,
```
mvilanova, line 15:
```suggestion
    business_year_hours: 2080,
```

mvilanova (changes requested):

src/dispatch/static/dispatch/src/project/NewEditSheet.vue
forestmonster: Ah, thank you.

src/dispatch/project/models.py
mvilanova: Permissions? Only organization owners and managers can create projects, and organization owners and managers and project admins can update projects

forestmonster: Ah, by this, I just meant that we should have some way to retain the "wisdom" around `1640` versus `2080`. Otherwise it's just in our heads. So in the future, if we're no longer here for some reason, and if it's not enforced by the software. we should have documentation saying "when you create a new project, you shouldn't use the default of 2080 hours per year. We've decided here that 2080 hours is incorrect for us; you should use 1640." That seems like internal Netflix documentation, so I'll add it.

mvilanova (approved):

PR #1980, opened 2022-02-11
Addresses a race condition on incident update

mvilanova:


src/dispatch/incident/flows.py
```diff
@@ -1215,9 +1237,6 @@ def incident_assign_role_flow(
             # we send a message to the incident commander with tips on how to manage the incident
             send_incident_management_help_tips_message(incident, db_session)
 
-        # we update the external ticket
-        update_external_incident_ticket(incident, db_session)
```
kevgliss, line 1219: Are there not areas where assign_role_flow gets called that's not from the update flow? If so, we wouldn't update the ticket in those instances.

kevgliss (approved):

src/dispatch/incident/flows.py
mvilanova: Yes, there was only one place and I added a call to update ticket: https://github.com/Netflix/dispatch/pull/1980/files#diff-07b5794644cf7adfbcc5321f07e7fc9445c1e80380ba927aeb9664dcbbf37229R531

forestmonster (commented):
Yep, let's see how it does in test.

PR #2048, opened 2022-03-08
Improvement/incident closed at

kevgliss:


src/dispatch/static/dispatch/src/incident/Table.vue
```diff
@@ -151,6 +159,7 @@ export default {
         { text: "Commander", value: "commander", sortable: false },
         { text: "Cost", value: "incident_costs", sortable: false },
         { text: "Reported At", value: "reported_at" },
+        { text: "Closed At", value: "closed_at" },
```
mvilanova, line 162: I believe the agreement was to add both stable at and closed at times.

PR #2052, opened 2022-03-09
Improves data filtering reliability

mvilanova:


kevgliss:
At this point, how much of that library have we vendored in? I think at this point it might make sense to vendor in the rest and create it's own module (e.g. `filtering`) or some such.

mvilanova:
Roughly between 50-60% based on line count. We can look into vendoring the rest of the code in in a separate PR.

src/dispatch/incident/views.py
```diff
@@ -304,7 +304,7 @@ def get_incident_forecast(
         if i == 1:
             incidents = create_incident_metric_query(
                 db_session=db_session,
-                filter_spec=common['filter_spec'],
+                filter_spec=common["filter_spec"],
```
mvilanova, line 307: @kevgliss you may want to double check that your black plugin is running correctly.

src/dispatch/database/service.py
```diff
@@ -45,28 +50,216 @@
...
+        if arity == 2:
+            return function(sqlalchemy_field, value)
+
+
+class BooleanFilter(object):
+    def __init__(self, function, *filters):
+        self.function = function
+        self.filters = filters
+
+    def get_named_models(self):
```
mvilanova, line 147: This func was refactored to use `SortedSet()`

src/dispatch/database/service.py
```diff
@@ -45,28 +50,216 @@
...
+
+
+def get_model_class_by_name(registry, name):
+    """Return the model class matching `name` in the given `registry`."""
+    for cls in registry.values():
+        if getattr(cls, "__name__", None) == name:
+            return cls
+
+
+def get_named_models(filters):
```
mvilanova, line 229: This func was refactored to use `[]`

PR #2083, opened 2022-03-16
Adds support for user settings

mvilanova:
This PR adds support for allowing organization users to set their default projects. This setting will later be used for filtering across data tables and dashboard.

Screen Shot 2022-03-17 at 1 38 12 PM


kevgliss (approved):

kevgliss:
Will the UI using these settings be in a separate PR?

PR #2093, opened 2022-03-18
Uses the simple search dictionary for fulltext search

kevgliss:
This change allows for more accurate searching for tag names. Since most tags are single words, we use the simple dictionary to capture the entire name (instead of breaking up lexemes like 'hack' & 'ing'. The search also enables better control when exact matching is required by wrapping your query in quotes. for example:

names: hacking, thehacking, thebiggesthack
lexemes: "hack"

query: hacking -> matches all
query: hack -> matches all
query: "hacking" -> matches hacking


forestmonster (approved):
Got it, thank you. 👍

PR #2123, opened 2022-03-29
Update Admin Documentation on Environment variables

Keetmalin:
I had to browse through the code to understand the possible values of Log levels. Apparently the log levels are in Upper case, and the default value is shown in lowercase, which misled me into believing that all log levels should be in lower case.

Hopefully, this change will be meaningful for other readers.


mvilanova:
Thanks for the PR!

PR #2149, opened 2022-04-12
Updates executive report template text

mvilanova:
Updates the text in the executive report template to try to address situations where the recipient of the report is not directly added to the notifications distribution list but indirectly through a nested distribution list.

### DIFF
--- a/src/dispatch/messaging/email/templates/executive_report.mjml
+++ b/src/dispatch/messaging/email/templates/executive_report.mjml
@@ -6,7 +6,7 @@
       {{ name }} - Executive Report
     </mj-text>
     <mj-text>
-      You're receiving this executive report, because you, or a distribution list you're a member of,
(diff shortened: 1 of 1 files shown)


src/dispatch/messaging/email/templates/executive_report.mjml
```diff
@@ -6,8 +6,8 @@
       {{ name }} - Executive Report
     </mj-text>
     <mj-text>
-      You're receiving this executive report, because you're a member of this incident's
-      notifications Google group ({{ notifications_group }}).
+      You're receiving this executive report, because you, or a distribution list you're a member of,
```
forestmonster, line 9:
```suggestion
      You're receiving this executive report, because you, or an e-mail distribution list you're on,
```

forestmonster (changes requested):
Just a language suggestion.

PR #2394, opened 2022-08-04
Attribute action_id missing in task actions button.

3c-un-str:
Prior to this, Slack was assinging a random 4-digit action_id that was causing the resolve and re-open buttons for a task to be a no-op.

Adding this routes the slack action to the right block_action_function (update_task_status).

Tested on my end.


mvilanova:
Nice fix. Thank you!

PR #2410, opened 2022-08-13
Copy over incident_costs, terms & duplicates before updating incident from Slack

3c-un-str:
We don't (yet) present incident_costs attribute/field in Slack's update_incident modal, causing the update function to incorrectly (re)set it to zero.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/modals/incident/handlers.py
+++ b/src/dispatch/plugins/dispatch_slack/modals/incident/handlers.py
@@ -204,6 +204,9 @@ def update_incident_from_submitted_form(
     # we don't allow visibility to be set in slack so we copy it over
     incident_in.visibility = incident.visibility
 
+    # We don't (yet) present incident_costs field in Slack's modal; copy it over.
+    incident_in.incident_costs = incident.incident_costs
+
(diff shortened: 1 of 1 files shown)


mvilanova:
Nice catch. I suspect the same is happening with `duplicates` and `terms`, which we also don't allow to be set via the Slack modal. We could address them in this PR or I can cut a separate one. Either way is fine with me.

3c-un-str:
Updated my PR.

src/dispatch/plugins/dispatch_slack/modals/incident/handlers.py
```diff
@@ -201,8 +201,11 @@ def update_incident_from_submitted_form(
 
     previous_incident = IncidentRead.from_orm(incident)
 
-    # we don't allow visibility to be set in slack so we copy it over
+    # we don't allow visibility, incident_costs, terms & duplicates to be set in slack, so we copy it over
```
mvilanova, line 204:
```suggestion
    # we currently don't allow users to update the incident's visibility,
    # costs, terms, or duplicates via Slack, so we copy them over
```

mvilanova (approved):

PR #2416, opened 2022-08-16
Fixes slack_menu returning empty tag list from non-incident channels

3c-un-str:
When reporting incidents from non-incident channels, db_session was being set to none as we were trying to get the org scope from channel_id; not the org slug. This was resulting in "no results found" for the tag modal field. 

Edited a function's comment to better reflect its functionality. 

Tested the change.


mvilanova:
Hey @3c-un-str, what did it happen here to make Github call this PR as merged to this repo even though there were 0 commits?

PR #2419, opened 2022-08-17
slack_menu empty tag list fix; uvloop support.

3c-un-str:
Add uvloop support to run the async event loop; uvicorn will pick this over asyncio, if available. Not sure about others, but In my setup (ubuntu VM), not having uvloop was resulting in 8/10 calls failing (some like /dispatch-update-incident failed 10/10) to Slack's API with the 'expired_trigger_id' error. I still see that error (w/uvloop), but it is more like 1/10. Also, this is an configurable option: "--loop [auto|asyncio|uvloop]    Event loop implementation.  [default: auto]"

Also, including the commit from earlier today which fixes slack_menu returning empty tag list from non-incident channels. Not sure how that was missing in the master branch although. 

Some minor str changes.

### DIFF
--- a/requirements-base.in
+++ b/requirements-base.in
@@ -42,4 +42,5 @@ statsmodels
 tabulate
 tenacity
 uvicorn
+uvloop
 validators==0.18.2
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -410,6 +410,8 @@ urllib3==1.26.11
     #   sentry-sdk
 uvicorn==0.18.2
     # via -r requirements-base.in
+uvloop==0.16.0
+    # via -r requirements-base.in
 validators==0.18.2
     # via -r requirements-base.in
 wasabi==0.10.0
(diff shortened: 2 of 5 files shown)


mvilanova:
Thanks for the PR. I'll take a look at it as soon as I can.

src/dispatch/evergreen/scheduled.py
```diff
@@ -54,15 +54,24 @@ def create_evergreen_reminder(
...
-        owner_email,
-        notification_text,
-        notification_template,
-        notification_type,
-        name=name,
-        subject=subject,
-        items=items,  # plugin expect dicts
-    )
+
+    # Gmail API seems to constantly throw "tenacity.RetryError: RetryError", even though the email(s)
```
mvilanova, line 58: Interesting... We are not experiencing the same behavior, but I'll take a look at our server logs to double check. Any useful errors in your GCP project that could shed light at what is happening?

mvilanova:
Can we split this PR into two: one for the Gmail protections and another one for the Slack stuff?

mvilanova:
The Slack and str changes look good. I can merge them once the Gmail ones are moved to a separate PR.

3c-un-str:
dropped the gmail changes from this PR. Will submit a new one.

PR #2430, opened 2022-08-18
Wrap gmail.send in try/except due to tenacity.RetryError

3c-un-str:
Gmail is consistency throwing below exception during .send(). GCP "golden signals" for the API look fine. Also, no similar issues with other gsuite calls like drive and groups. 

`ERROR:Error/Exception in sending Tactical Report email to ...... RetryError[<Future at 0x7f613d72ce80 state=finished raised KeyError>]  `

Weird thing is, the emails are still being sent. Actually, sent three times (tenacity retry_count of 3) to a domain account; one email to a regular gmail account (despite making three calls and same end error). 

Without the wrap, incident creation fails at the send_inc_notifs stage. Not sure if anyone else is experiencing this, but if they do,  this should help in the interim. 

I will continue to dig deeper.

### DIFF
--- a/src/dispatch/incident/messaging.py
+++ b/src/dispatch/incident/messaging.py
@@ -185,7 +185,7 @@ def send_welcome_email_to_participant(
             **message_kwargs,
         )
     except Exception as e:
-        log.error(f"Error in sending welcome email to {participant_email}: {e}.")
+        log.error(f"Error in sending welcome email to {participant_email}: {e}")
 
     log.debug(f"Welcome email sent to {participant_email}.")
 
--- a/src/dispatch/incident/messaging.py
+++ b/src/dispatch/incident/messaging.py
@@ -174,13 +174,18 @@ def send_welcome_email_to_participant(
         )
 
     notification_text = "Incident Notification"
-    plugin.instance.send(
-        participant_email,
-        notification_text,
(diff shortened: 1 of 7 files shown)


src/dispatch/incident/messaging.py
```diff
@@ -174,13 +174,18 @@ def send_welcome_email_to_participant(
...
+    try:
+        plugin.instance.send(
+            participant_email,
+            notification_text,
+            INCIDENT_PARTICIPANT_WELCOME_MESSAGE,
+            MessageType.incident_participant_welcome,
+            **message_kwargs,
+        )
+    except Exception as e:
+        log.error(f"Error in sending welcome email to {participant_email}: {e}.")
```
mvilanova, line 188:
```suggestion
        log.error(f"Error in sending welcome email to {participant_email}: {e}")
```

mvilanova:
We're not experiencing this issue and haven't heard about others. Changes look good. I think we should eventually move all these `plugin.instance.send` calls to an `email/flows.py` module to reduce copy/pasta and maintenance. Same with calls to send Slack messages.

3c-un-str:
Yep, I agree; googling the error doesn't show much. It's not a HttpError (tried catching that); it's just throwing an exception. 

I guess the gods have been dropping some bits en route ;-).

Thanks,

PR #2440, opened 2022-08-19
Always upper log_level before logging config

filippog:
Hello,
thank you for dispatch, looks great and super useful!

While testing/deploying I came across an inconsistency between cli's `--log-level` and `LOG_LEVEL` in env/config. Namely the latter can be only uppercase while the former is lower case. What do you think of the attached commit ?

thank you!


mvilanova (approved):
Changes look good. Thanks for the PR!

PR #2442, opened 2022-08-20
Fixes document creation when template file does not exist

3c-un-str:
During create_flow, if the reporting template doesn't exist, we create a new one, but immediately try to access the non-existent description attr during update(), which throws an exception. This fixes it.

Added code to create a new tracking sheet if template is not in config, with update() moved like above.

Added if/else flow based on the mimeType.

Tested my changes.


kjalleda:
Will submit a fresh one.

3c-un-str:
will submit a new one

PR #2455, opened 2022-08-23
Fixes issue during the creation of incident collaboration documents

mvilanova:
Fixes an issue introduced in https://github.com/Netflix/dispatch/pull/2443. Ensures we always have a description for each collaboration document.


mvilanova:
cc @3c-un-str

3c-un-str:
Even better. Thanks.

PR #2501, opened 2022-09-09
Bugfix/table layout

kevgliss:
At larger screen sizes this makes sure the table fills the entire container (doesn't have extra padding). Additionally, we don't specify the number of columns for the buttons and instead let them expand as necessary and simply ensure they are to the right of the column.


mvilanova (approved):
LGTM

PR #2510, opened 2022-09-12
Runs every scheduled job in it's own thread.

kevgliss:
This ensures that each scheduled task is not run serially. This should increase the reliability of the scheduled tasks, guaranteeing that long-running tasks do not cause other tasks to miss their execution windows.


mvilanova (approved):
Looks good.

Issue #2276, opened 2022-06-22
Missing image in the docs


rlevit:
**Describe the bug**
The image is missing on [this](url) page. There is an image placeholder with the message that reads "Could not load image".

**Screenshots**
<img width="831" alt="Screen Shot 2022-06-22 at 10 59 59 AM" src="https://user-images.githubusercontent.com/15822524/175062977-39c1b6aa-3b10-461a-93fc-c6b5e4747af1.png">

mvilanova:
Not sure what @kevgliss had in mind here. I'll let him add it when he's back.

PR #2525, opened 2022-09-15
Updates documentation

mvilanova:
* Re-organizes structure to be more inline with configuration process
* Adds docs for case types, severities, and priorities
* Updates several docs


mvilanova:
@kevgliss not sure what you had in mind for https://github.com/Netflix/dispatch/issues/2276, but I can take care of it on this one if you tell me where should I take the screenshot.

PR #2528, opened 2022-09-16
Fixes task reminders not being sent; fixes bug in in switching organizations

3c-un-str:


src/dispatch/static/dispatch/src/router/index.js
```diff
@@ -50,10 +50,10 @@ router.beforeEach((to, from, next) => {
           .then(function () {
             return userSettings.load()
           })
-          .then(next)
+          .then(next())
```
mvilanova, line 53: I tried this in https://github.com/Netflix/dispatch/pull/2335, but had to roll the changes back in https://github.com/Netflix/dispatch/pull/2346, because it caused Dispatch to not wait until the user settings would be loaded to continue loading the rest of components. Curious if you observed the same behavior during testing?

mvilanova:
Thanks for the fix for the task reminders. It was in my todo list :)

src/dispatch/static/dispatch/src/router/index.js
3c-un-str: Based on my testing so far, most switchOrganizations calls work except for the project settings paths (if that's what you were referring to), like /default/settings/projects/caseTypes?project=default. If you are on of these paths (/default/settings/projects/*), the switch just reloads w/the current context.  Rest of the paths switch ok. 

Either way, I think this is probably a better UX, than not having any switches work at all :), until of course, we figure out what is causing that particular path (at least in my case) to not work. I will of course continue to look into this, and report back findings.

mvilanova: Ok, sounds good. Would you mind moving these changes to another PR, so we can merge the rest of them? Thanks!

PR #2531, opened 2022-09-17
Fixes task reminder not being sent

3c-un-str:


mvilanova:
Looks good. Thank you!

PR #2535, opened 2022-09-19
Enhancement/case report

kevgliss:
This PR duplicates the functionality we have for incident reports and allows for cases to be reported through a streamlined form instead of deep within the UI. 

These screens are located at the /cases/report and the /cases/status endpoints. 

Screen Shot 2022-09-19 at 1 21 12 PM
Screen Shot 2022-09-19 at 1 21 40 PM


mvilanova:
This PR includes the code from the signals PR.

kevgliss:
Yes, I am removing those commits.

PR #2536, opened 2022-09-19
Adds case reporting support

kevgliss:
This PR duplicates functionality that we have with reporting incidents, allowing for the direct reporting of cases via a form. 

Screen Shot 2022-09-19 at 1 21 12 PM
Screen Shot 2022-09-19 at 1 21 40 PM

### DIFF
--- a/src/dispatch/static/dispatch/src/case/CaseSummaryTable.vue
+++ b/src/dispatch/static/dispatch/src/case/CaseSummaryTable.vue
@@ -1,0 +1,99 @@
+<template>
+  <div>
+    <v-data-table
+      :headers="headers"
+      :items="items"
+      :loading="loading"
+      :items-per-page="-1"
+      disabled-pagination
+      hide-default-footer
+    >
+      <template v-slot:item.case_priority.name="{ item }">
+        <case-priority :priority="item.case_priority.name" />
+      </template>
+      <template v-slot:item.status="{ item }">
+        <case-status :status="item.status" :id="item.id" />
+      </template>
+      <template v-slot:item.reporter="{ item }">
+        <case-participant :participant="item.reporter" />
+      </template>
+      <template v-slot:item.project.name="{ item }">
+        <v-chip small :color="item.project.color" text-color="white">
+          {{ item.project.name }}
+        </v-chip>
+      </template>
+      <template v-slot:item.reported_at="{ item }">
+        <v-tooltip bottom>
+          <template v-slot:activator="{ on, attrs }">
+            <span v-bind="attrs" v-on="on">{{ item.reported_at | formatRelativeDate }}</span>
+          </template>
+          <span>{{ item.reported_at | formatDate }}</span>
+        </v-tooltip>
+      </template>
+      <template v-slot:item.data-table-actions="{ item }">
+        <v-menu bottom left>
+          <template v-slot:activator="{ on }">
+            <v-btn icon v-on="on">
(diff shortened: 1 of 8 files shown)


src/dispatch/static/dispatch/src/case/ReportSubmissionCard.vue
```diff
@@ -0,0 +1,205 @@
...
+      isSubmitted: false,
+      project_faq: null,
+    }
+  },
+
+  computed: {
+    ...mapFields("case_management", [
+      "selected.case_priority",
+      "selected.case_type",
+      "selected.commander",
```
mvilanova, line 125: Remove.

src/dispatch/static/dispatch/src/case/ReportSubmissionCard.vue
```diff
@@ -0,0 +1,205 @@
...
+      "selected.title",
+      "selected.tags",
+      "selected.description",
+      "selected.conversation",
+      "selected.conference",
```
mvilanova, line 130: Remove or comment out.

src/dispatch/static/dispatch/src/case/CaseSummaryTable.vue
```diff
@@ -0,0 +1,99 @@
...
+      hide-default-footer
+    >
+      <template v-slot:item.case_priority.name="{ item }">
+        <case-priority :priority="item.case_priority.name" />
+      </template>
+      <template v-slot:item.status="{ item }">
+        <case-status :status="item.status" :id="item.id" />
+      </template>
+      <template v-slot:item.reporter="{ item }">
+        <case-participant :participant="item.reporter" />
```
mvilanova, line 18: There's no component for case participant. Did you forget to add it and import it?

mvilanova (changes requested):

src/dispatch/static/dispatch/src/case/CaseSummaryTable.vue
kevgliss: I will remove it.

kevgliss: Afaict it looks like cases don't track the reporter, I think that's probably something that should be added.

src/dispatch/static/dispatch/src/case/ReportReceiptCard.vue
```diff
@@ -0,0 +1,260 @@
...
+  components: {},
+  data() {
+    return {
+      isSubmitted: false,
+      activeResourcePlugins: {
+        document: null,
+        ticket: null,
+        storage: null,
+        conversation: null,
+        conference: null,
```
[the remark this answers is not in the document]
kevgliss: Is the question around whether a case would ever have an associated conference resource? I can't think of a reason, I will remove both and we can add them back when support is available.

src/dispatch/static/dispatch/src/case/CaseSummaryTable.vue
mvilanova: Yes, I agree we should add it.

src/dispatch/static/dispatch/src/case/ReportReceiptCard.vue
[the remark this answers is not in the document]
mvilanova: Sounds good.

mvilanova (approved):

PR #2541, opened 2022-09-20
Enhancement/case escalation modal

kevgliss:
Adds the ability to modify incident details before a case is escalated via a dedicated dialog. 

Screen Shot 2022-09-20 at 1 21 56 PM
Screen Shot 2022-09-26 at 1 52 57 PM
Screen Shot 2022-09-20 at 4 21 45 PM
Screen Shot 2022-09-20 at 4 22 03 PM

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -412,7 +412,9 @@ def case_to_incident_escalate_flow(
     )
 
 
-def case_escalate_flow(case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal):
+def case_to_incident_endpoint_escalate_flow(
+    case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal
+):
     """Allows for a case to be escalated to an incident while modifying its properties."""
     case = get(case_id=case_id, db_session=db_session)
     incident = incident_flows.incident_create_flow(incident_id=incident_id, db_session=db_session)
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -13,7 +13,7 @@ from dispatch.group.enums import GroupType, GroupAction
 from dispatch.incident import flows as incident_flows
 from dispatch.incident import service as incident_service
 from dispatch.incident.enums import IncidentStatus
-from dispatch.incident.models import IncidentCreate, Incident
+from dispatch.incident.models import IncidentCreate
 from dispatch.individual.models import IndividualContactRead
 from dispatch.models import OrganizationSlug, PrimaryKey
 from dispatch.participant.models import ParticipantUpdate
@@ -239,9 +239,35 @@ def case_triage_status_flow(case: Case, db_session=None):
 def case_escalated_status_flow(case: Case, organization_slug: OrganizationSlug, db_session=None):
     """Runs the case escalated transition flow."""
     # we set the escalated_at time
-    case.escalated_at = datetime.utcnow()
-    db_session.add(case)
-    db_session.commit()
+    if case.incidents:
+        # we don't escalate the case if the case is already linked to incidents
+        return
+
+    if not case.case_type.incident_type:
+        # we don't escalate the case if its type is not mapped to an incident type
+        return
+
+    # we make the assignee of the case the reporter of the incident
+    reporter = ParticipantUpdate(individual=IndividualContactRead(email=case.assignee.email))
+
+    # we add information about the case in the incident's description
+    description = (
+        f"{case.description}\n\n"
+        f"This incident was the result of escalating case {case.name} "
+        f"in the {case.project.name} project. Check out the case in the Dispatch Web UI for additional context."
+    )
+
+    # we create the incident
+    incident_in = IncidentCreate(
+        title=case.title,
+        description=description,
+        status=IncidentStatus.active,
+        incident_type=case.case_type.incident_type,
+        incident_priority=case.case_priority,
+        project=case.case_type.incident_type.project,
+        reporter=reporter,
+    )
+    incident_service.create(db_session=db_session, incident_in=incident_in)
 
     case_to_incident_escalate_flow(
         case=case, organization_slug=organization_slug, db_session=db_session
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -15,7 +15,7 @@ from dispatch.incident import service as incident_service
 from dispatch.incident.enums import IncidentStatus
 from dispatch.incident.models import IncidentCreate, Incident
 from dispatch.individual.models import IndividualContactRead
-from dispatch.models import OrganizationSlug
+from dispatch.models import OrganizationSlug, PrimaryKey
 from dispatch.participant.models import ParticipantUpdate
 from dispatch.storage import flows as storage_flows
 from dispatch.storage.enums import StorageAction
@@ -386,10 +386,10 @@ def case_to_incident_escalate_flow(
     )
 
 
-def case_escalate_flow(case: Case, incident: Incident, db_session: SessionLocal):
+def case_escalate_flow(case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal):
     """Allows for a case to be escalated to an incident while modifying it's properties."""
-
-    incident_flows.incident_create_flow(incident_id=incident.id, db_session=db_session)
+    case = get(case_id=case_id, db_session=db_session)
+    incident = incident_flows.incident_create_flow(incident_id=incident_id, db_session=db_session)
 
     case.escalated_at = datetime.utcnow()
     case.status = CaseStatus.escalated
@@ -421,5 +421,4 @@ def case_escalate_flow(case: Case, incident: Incident, db_session: SessionLocal)
         description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
         case_id=case.id,
     )
-
     return incident
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -239,34 +239,9 @@ def case_triage_status_flow(case: Case, db_session=None):
 def case_escalated_status_flow(case: Case, organization_slug: OrganizationSlug, db_session=None):
     """Runs the case escalated transition flow."""
     # we set the escalated_at time
-    if case.incidents:
-        # we don't escalate the case if the case is already linked to incidents
-        return
-
-    if not case.case_type.incident_type:
-        # we don't escalate the case if its type is not mapped to an incident type
-        return
-
-    # we make the assignee of the case the reporter of the incident
-    reporter = ParticipantUpdate(individual=IndividualContactRead(email=case.assignee.email))
-
-    # we add information about the case in the incident's description
-    description = (
-        f"{case.description}\n\n"
-        f"This incident was the result of escalating case {case.name} "
-        f"in the {case.project.name} project. Check out the case in the Dispatch Web UI for additional context."
-    )
-
-    # we create the incident
-    incident_in = IncidentCreate(
-        title=case.title,
-        description=description,
-        status=IncidentStatus.active,
-        incident_type=case.case_type.incident_type,
-        incident_priority=case.case_priority,
-        project=case.case_type.incident_type.project,
-        reporter=reporter,
-    )
+    case.escalated_at = datetime.utcnow()
+    db_session.add(case)
+    db_session.commit()
     incident_service.create(db_session=db_session, incident_in=incident_in)
 
     case_to_incident_escalate_flow(
@@ -412,8 +387,12 @@ def case_to_incident_escalate_flow(
     )
 
 
+@background_task
 def case_to_incident_endpoint_escalate_flow(
-    case_id: PrimaryKey, incident_id: PrimaryKey, organization_slug: OrganizationSlug, db_session=None
+    case_id: PrimaryKey,
+    incident_id: PrimaryKey,
+    organization_slug: OrganizationSlug,
+    db_session=None,
 ):
     """Allows for a case to be escalated to an incident while modifying its properties."""
     case = get(case_id=case_id, db_session=db_session)
@@ -449,4 +428,3 @@ def case_to_incident_endpoint_escalate_flow(
         description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
         case_id=case.id,
     )
-    return incident
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -242,7 +242,6 @@ def case_escalated_status_flow(case: Case, organization_slug: OrganizationSlug, 
     case.escalated_at = datetime.utcnow()
     db_session.add(case)
     db_session.commit()
-    incident_service.create(db_session=db_session, incident_in=incident_in)
 
     case_to_incident_escalate_flow(
         case=case, organization_slug=organization_slug, db_session=db_session
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -395,7 +395,9 @@ def case_to_incident_endpoint_escalate_flow(
 ):
     """Allows for a case to be escalated to an incident while modifying its properties."""
     case = get(case_id=case_id, db_session=db_session)
-    incident = incident_flows.incident_create_flow(incident_id=incident_id, db_session=db_session)
+    incident = incident_flows.incident_create_flow(
+        incident_id=incident_id, organization_slug=organization_slug, db_session=db_session
+    )
 
     case.escalated_at = datetime.utcnow()
     case.status = CaseStatus.escalated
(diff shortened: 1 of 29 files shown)


mvilanova:
I'll run the code locally and circle back with feedback.

src/dispatch/case/flows.py
```diff
@@ -384,3 +384,41 @@ def case_to_incident_escalate_flow(
         description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
         case_id=case.id,
     )
+
+
+def case_escalate_flow(case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal):
+    """Allows for a case to be escalated to an incident while modifying it's properties."""
```
mvilanova, line 390:
```suggestion
    """Allows for a case to be escalated to an incident while modifying its properties."""
```

src/dispatch/case/views.py
```diff
@@ -167,6 +172,46 @@ def update_case(
...
+    """Escalates an existing case."""
+    # current user is better than assignee (although likely the same)
+    if not incident_in.reporter:
+        incident_in.reporter = ParticipantUpdate(
+            individual=IndividualContactRead(email=current_user.email)
+        )
+
+    # allow for default values
+    if not incident_in.incident_type:
+        incident_in.incident_type = {"name": current_case.case_type.incident_type.name}
```
mvilanova, line 199: We're assuming that the case type is mapped to an incident type here, which may not always be the case. We should check if the mapping exists, and if not, default to the default incident type on that project?

src/dispatch/case/views.py
```diff
@@ -167,6 +172,46 @@ def update_case(
...
+
+    if not incident_in.incident_priority:
+        incident_in.incident_priority = {"name": current_case.case_priority.name}
```
mvilanova, line 202: We're assuming that cases and incident priorities have the same name here. Perhaps we default to the project default one?

src/dispatch/case/views.py
```diff
@@ -167,6 +172,46 @@ def update_case(
...
+
+    if not incident_in.project:
+        incident_in.project = {"name": current_case.case_type.incident_type.project.name}
```
mvilanova, line 205: Same assumption here.

src/dispatch/case/flows.py
```diff
@@ -384,3 +384,41 @@ def case_to_incident_escalate_flow(
         description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
         case_id=case.id,
     )
+
+
+def case_escalate_flow(case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal):
```
mvilanova, line 389: Should we rename it to `case_to_incident_escalated_custom_flow` to denote that it's a custom version of `case_to_incident_escalate_flow`?

src/dispatch/case/flows.py
```diff
@@ -384,3 +384,41 @@ def case_to_incident_escalate_flow(
...
+    """Allows for a case to be escalated to an incident while modifying it's properties."""
+    case = get(case_id=case_id, db_session=db_session)
+    incident = incident_flows.incident_create_flow(incident_id=incident_id, db_session=db_session)
+
+    case.escalated_at = datetime.utcnow()
+    case.status = CaseStatus.escalated
+    case.incidents.append(incident)
```
mvilanova, line 396:
```suggestion
    incident = incident_flows.incident_create_flow(incident_id=incident_id, db_session=db_session)
    
    case = get(case_id=case_id, db_session=db_session)
    case.escalated_at = datetime.utcnow()
    case.status = CaseStatus.escalated
    case.incidents.append(incident)
```

src/dispatch/case/flows.py
```diff
@@ -384,3 +384,41 @@ def case_to_incident_escalate_flow(
...
+        storage_members=storage_members,
+        db_session=db_session,
+    )
+
+    event_service.log_case_event(
+        db_session=db_session,
+        source="Dispatch Core App",
+        description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
+        case_id=case.id,
+    )
```
mvilanova, line 423: This is copypasta from the `case_to_incident_escalate_flow` function. Should we create a ` case_to_incident_escalate_flow_common` function and move this logic there and call the function from both `case_to_incident_escalate_flow` and `case_to_incident_escalate_custom_flow`?

src/dispatch/incident/flows.py
```diff
@@ -635,7 +635,7 @@ def incident_create_closed_flow(
 
 
 @background_task
-def incident_create_flow(*, organization_slug: str, incident_id: int, db_session=None):
+def incident_create_flow(*, incident_id: int, db_session=None):
```
mvilanova, line 638: Hmm did we test all the calls to this function after removing `organization_slug`?

src/dispatch/static/dispatch/src/case/EditSheet.vue
```diff
@@ -11,6 +11,10 @@
               Reported - {{ reported_at | formatRelativeDate }}
             </v-list-item-subtitle>
           </v-list-item-content>
+          <v-btn @click="showEscalateDialog(selected)" color="error">
+            <v-icon left> error_outline </v-icon>
+            Escalate Case as Incident
```
mvilanova, line 16: Incident is the only thing that a case can be escalated to, so I'd suggest we remove it
```suggestion
            Escalate Case
```

mvilanova: It also helps with having a smaller button.

src/dispatch/static/dispatch/src/case/EscalateDialog.vue
```diff
@@ -0,0 +1,93 @@
...
+        <v-spacer />
+        <v-btn color="blue en-1" text @click="closeEscalateDialog()"> Close </v-btn>
+      </v-card-actions>
+    </v-card>
+    <v-card v-else>
+      <v-card-title>
+        <span class="headline">Escalate Case?</span>
+      </v-card-title>
+      <v-card-text>
+        Lets do it. Update information as needed or accept the pre-filled defaults.
```
mvilanova, line 20:
```suggestion
        Update the fields or accept the pre-filled defaults.
```

src/dispatch/static/dispatch/src/case/Table.vue
```diff
@@ -79,6 +80,9 @@
                   </v-btn>
                 </template>
                 <v-list>
+                  <v-list-item @click="showEscalateDialog(item)">
+                    <v-list-item-title>Escalate as Incident</v-list-item-title>
```
mvilanova, line 84:
```suggestion
                    <v-list-item-title>Escalate</v-list-item-title>
```

mvilanova: Same thinking as above.

src/dispatch/static/dispatch/src/case/store.js
```diff
@@ -173,12 +174,46 @@ const actions = {
...
+    commit("incident/RESET_SELECTED", null, { root: true })
+  },
+  // showReportDialog({ commit }, value) {
+  //   commit("SET_DIALOG_REPORT", true)
+  //   commit("SET_SELECTED", value)
+  // },
+  // closeReportDialog({ commit }) {
+  //   commit("SET_DIALOG_REPORT", false)
+  //   commit("RESET_SELECTED")
+  // },
```
mvilanova, line 193: Let's remove them if they're not being used.

src/dispatch/static/dispatch/src/case/store.js
```diff
@@ -304,6 +339,9 @@ const mutations = {
   SET_DIALOG_DELETE(state, value) {
     state.dialogs.showDeleteDialog = value
   },
+  SET_DIALOG_ESCALATE(state, value) {
+    state.dialogs.showEscalateDialog = value
+  },
   SET_DIALOG_REPORT(state, value) {
     state.dialogs.showReportDialog = value
   },
```
mvilanova, line 347: I think this can be removed as well.

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
```diff
@@ -0,0 +1,218 @@
...
+              </v-list-item-subtitle>
+            </v-list-item>
+            <v-list-item>
+              <v-list-item-title>Priority</v-list-item-title>
+              <v-list-item-subtitle class="text-right">
+                <incident-priority :priority="incident_priority.name" />
+              </v-list-item-subtitle>
+            </v-list-item>
+            <v-list-item>
+              <v-list-item-title>Visability</v-list-item-title>
```
mvilanova, line 32:
```suggestion
              <v-list-item-title>Visibility</v-list-item-title>
```

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
```diff
@@ -0,0 +1,218 @@
...
+                v-if="resourceData(pluginInstance.plugin.type)"
+                target="_blank"
+                :href="resourceData(pluginInstance.plugin.type).weblink"
+              >
+                <v-list-item-content>
+                  <v-list-item-title>{{
+                    pluginInstance.plugin.type | capitalize
+                  }}</v-list-item-title>
+                  <v-list-item-subtitle>{{
+                    pluginInstance.plugin.description
```
mvilanova, line 114: I think showing the plugin type name and description instead of the resource name and description like before might be confusing for users.

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
```diff
@@ -0,0 +1,218 @@
+<template>
+  <v-container>
+    <v-row dense>
+      <v-col cols="12">
+        <v-card outlined elevation="0">
+          <v-list-item two-line>
+            <v-list-item-content>
+              <v-list-item-title class="text-h5"> Incident Details </v-list-item-title>
+              <v-list-item-subtitle>{{ reported_at | formatRelativeDate }}</v-list-item-subtitle>
```
mvilanova, line 9: Isn't this time always going to render as `Today`? Should we remove it if so?

kevgliss: Good point, will remove.

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
kevgliss: This is technically new functionality and won't affect the current submissions, but I agree they should be the same, afaik before we had one off strings that weren't derived from anywhere. This was my attempt as generalizing them. We can either use these strings, add new ones or take the hit and make it so each reasource is rendered individually.

src/dispatch/case/views.py
mvilanova: What was the resolution here?

src/dispatch/incident/flows.py
kevgliss: hmm, I think you're right let me add it back.

src/dispatch/case/views.py
mvilanova: What was the resolution here?

kevgliss: The project default is handled during case creation, I have a change that does not include that value if no `case_type` exists.

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
mvilanova: I think we should be consistent across the app and showing plugin info instead of resource info is confusing. When I look at the screenshot and I see the plugin title and description under incident resources it makes me think I created a plugin not a resource.

kevgliss: I understand, but we have to choose how we accomplish this, our options are 1) Use these strings 2) Create new strings 3) Render each resource individually. It sounds like you prefer option 3?

src/dispatch/case/flows.py
```diff
@@ -239,9 +239,35 @@ def case_triage_status_flow(case: Case, db_session=None):
 def case_escalated_status_flow(case: Case, organization_slug: OrganizationSlug, db_session=None):
     """Runs the case escalated transition flow."""
     # we set the escalated_at time
```
mvilanova, line 241: Remove

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
mvilanova: Yeah, I prefer # 3 as it's more clear for the user and consistent with what we already have.

src/dispatch/case/flows.py
mvilanova: Oh wait, this and the other code below that was removed needs to stay and what was added to be removed as it already lives in the `case_to_incident_escalate_flow` function.

src/dispatch/case/flows.py
```diff
@@ -384,3 +410,43 @@ def case_to_incident_escalate_flow(
...
+        db_session=db_session,
+    )
+
+    event_service.log_case_event(
+        db_session=db_session,
+        source="Dispatch Core App",
+        description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
+        case_id=case.id,
+    )
+    return incident
```
mvilanova, line 452: I don't think we need this return, as we don't have the ability to capture it in the background task that calls this function.

src/dispatch/case/flows.py
```diff
@@ -384,3 +410,43 @@ def case_to_incident_escalate_flow(
         description=f"The members of the incident's tactical group {incident.tactical_group.email} have been given permission to access the case's storage folder",
         case_id=case.id,
     )
+
+
```
mvilanova, line 414: Aren't we missing the `@background_task` decorator here?

src/dispatch/case/flows.py
```diff
@@ -384,3 +410,43 @@ def case_to_incident_escalate_flow(
...
+def case_to_incident_endpoint_escalate_flow(
+    case_id: PrimaryKey, incident_id: PrimaryKey, db_session: SessionLocal
```
mvilanova, line 416:
```suggestion
    case_id: PrimaryKey, incident_id: PrimaryKey, organization_slug: OrganizationSlug, db_session=None
```

src/dispatch/case/flows.py
```diff
@@ -239,9 +239,35 @@ def case_triage_status_flow(case: Case, db_session=None):
 def case_escalated_status_flow(case: Case, organization_slug: OrganizationSlug, db_session=None):
     """Runs the case escalated transition flow."""
     # we set the escalated_at time
-    case.escalated_at = datetime.utcnow()
-    db_session.add(case)
-    db_session.commit()
```
mvilanova, line 244: Keep

src/dispatch/case/flows.py
```diff
@@ -239,9 +239,35 @@ def case_triage_status_flow(case: Case, db_session=None):
...
+    incident_in = IncidentCreate(
+        title=case.title,
+        description=description,
+        status=IncidentStatus.active,
+        incident_type=case.case_type.incident_type,
+        incident_priority=case.case_priority,
+        project=case.case_type.incident_type.project,
+        reporter=reporter,
+    )
+    incident_service.create(db_session=db_session, incident_in=incident_in)
```
mvilanova, line 270: Remove. This lives in `case_to_incident_escalate_flow`.

src/dispatch/case/views.py
```diff
@@ -167,6 +172,48 @@ def update_case(
...
+    if not incident_in.project:
+        if current_case.case_type.incident_type:
+            incident_in.project = {"name": current_case.case_type.incident_type.project.name}
+
+    incident = incident_service.create(db_session=db_session, incident_in=incident_in)
+    background_tasks.add_task(
+        case_to_incident_endpoint_escalate_flow,
+        case_id=current_case.id,
+        incident_id=incident.id,
+        db_session=db_session,
```
mvilanova, line 211:
```suggestion
        organization_slug=organization,
```

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
```diff
@@ -0,0 +1,214 @@
...
+        }
+      })
+    })
+  },
+  computed: {
+    ...mapFields("incident", [
+      "selected",
+      "selected.incident_priority",
+      "selected.incident_type",
+      "selected.reported_at",
```
mvilanova, line 189:
```suggestion
```

src/dispatch/static/dispatch/src/incident/ReportReceiptResources.vue
```diff
@@ -0,0 +1,214 @@
...
+      "selected.commander",
+      "selected.participants",
```
mvilanova, line 191:
```suggestion
```

mvilanova (approved):

PR #2544, opened 2022-09-21
Improves logging configuration

mvilanova:
Defines and uses default log level if no level or unknown level is provided in the `.env` file.


kevgliss:
lgtm, the only thing I might change is having that logging enum in the enum.py if it will only ever be used by logging.py, I don't see a lot of value of keeping them seperate.

kevgliss (approved):

mvilanova:
Good point. I don't think it will be used outside the logging.py module. I'll move it to the logging.py module.

PR #2546, opened 2022-09-21
Show related cases in the incident details tab

mvilanova:
This PR adds support for displaying related cases in the incident details tab.

Screen Shot 2022-09-30 at 10 17 24 AM


src/dispatch/static/dispatch/src/case/Table.vue
```diff
@@ -80,9 +80,6 @@
                   </v-btn>
                 </template>
                 <v-list>
-                  <v-list-item @click="showEscalateDialog(item)">
-                    <v-list-item-title>Escalate</v-list-item-title>
-                  </v-list-item>
```
mvilanova, line 85: @kevgliss FYI I moved the escalate option after view/ edit.

PR #2550, opened 2022-09-23
Enhancement/case management workflows

kevgliss:
This change allows for workflows to be associated with cases. Additionally, it provides UI to allow workflows to be executed from the web UI (previously it was exclusive to Slack) for both cases and incidents.

As part of the consolidation, I've simplified the syncing of workflows to find _any_ outstanding or uncompleted workflow (as opposed for just open incidents).

It also adds a new endpoint `/run` to the workflows endpoint. 

Screen Shot 2022-09-23 at 3 44 26 PM
Screen Shot 2022-10-03 at 3 01 34 PM
Screen Shot 2022-10-03 at 3 01 57 PM


src/dispatch/cli.py
```diff
@@ -607,8 +607,7 @@ def dispatch_scheduler():
     )  # noqa
     from .term.scheduled import sync_terms  # noqa
     from .workflow.scheduled import (
-        daily_sync_workflow,  # noqa
-        sync_active_stable_workflows,  # noqa
+        sync_workflow,  # noqa
     )
```
mvilanova, line 611: Nit: this could be a one-liner

src/dispatch/workflow/scheduled.py
```diff
@@ -25,110 +23,84 @@
 WORKFLOW_SYNC_INTERVAL = 30  # seconds
 
 
-def sync_workflows(db_session, project, workflow_plugin, incidents, notify: bool = False):
+def sync_workflow(db_session, project, workflow_plugin, instance, notify: bool = False):
```
mvilanova, line 26: Nit: add typing

src/dispatch/workflow/service.py
```diff
@@ -83,22 +86,48 @@ def get_instance(*, db_session, instance_id: int) -> WorkflowInstance:
...
-    creator = participant_service.get_by_incident_id_and_email(
-        db_session=db_session, incident_id=incident.id, email=instance_in.creator.individual.email
-    )
-    instance.creator = creator
+    if instance_in.creator:
+        creator = participant_service.get_by_incident_id_and_email(
+            db_session=db_session,
+            incident_id=incident.id,
+            email=instance_in.creator.individual.email,
+        )
```
mvilanova, line 129: How's this going to work for cases where we don't have participants (yet)?

mvilanova (approved):

src/dispatch/workflow/service.py
kevgliss: We don't pass a creator for cases. This is only here to support the existing incident use case (where we do have participants). When/if we add participants to cases we can pass a creator when creating workflows from cases.

PR #2569, opened 2022-10-03
Add hint to using CLI to install plugins

caioaao:
I've spent a considerable amount of time trying to debug why the plugins wouldn't show up in the `Add Plugin` form in the dashboard. Maybe it's missing because most people start with the prepopulated DB? If so, wouldn't it be better to add some of this to the init db script?

### DIFF
--- a/docs/admin-guide/administration/cli.md
+++ b/docs/admin-guide/administration/cli.md
@@ -229,3 +229,19 @@ Google Docs - Document            google-docs-document            0.1.0       do
 Google Group - Participant Group  google-group-participant-group  0.1.0       participant_group  Kevin Glisson  Uses Google Groups to help manage participant membership.
 ...
 ```
+
+### Install
+
+The `install` command will try installing all available plugins.
+
+```bash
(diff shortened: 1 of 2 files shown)


mvilanova:
Thanks for the PR. I think it would be better if we add a new section called `Install` [here](https://hawkins.gitbook.io/dispatch/administration-guide/cli#plugins) and we reference it [here](https://hawkins.gitbook.io/dispatch/administration-guide/plugins).

mvilanova:
We may want to take the opportunity and fix the dispatch command in the [List](https://hawkins.gitbook.io/dispatch/administration-guide/cli#list-1) section where we say `dispatch database list` instead of `dispatch plugins list`.

mvilanova:
@caioaao thanks for fixing the command. Not sure if you saw my suggestion in this comment: https://github.com/Netflix/dispatch/pull/2569#issuecomment-1266089197

mvilanova (approved):
LGTM

PR #2608, opened 2022-10-17
Adds support for incident severity

mvilanova:
This PR adds support for defining a list of incident severities that then can be used to set the level of severity of an incident via the Web UI or the `/dispatch-update-incident` Slack command.  An incident severity called "Undetermined" will be added to all existing projects and set to all incidents in all projects. A new Severities chart has been added to the incidents dashboard.

Screen Shot 2022-10-18 at 11 05 30 AM


mvilanova:
This is a test.

PR #2625, opened 2022-10-25
Adds support for a new incident observer participant role

mvilanova:
This PR adds support for a new incident participant role called observer that will enable us to more accurately calculate incident cost. Incident participants that get resolved via engagement filters will be added with this role, as those are usually added for informational purposes. Existing incident participants will also be able to add participants with this role to the incident using the `/dispatch-assign-role` Slack slash command. Participants with the observer role will get their role converted to participant once and if they send 10 messages to the incident channel. Until then, time spent for them won't be included in the incident cost. The same will apply to the reporter role moving forward.


src/dispatch/plugins/dispatch_slack/events.py
```diff
@@ -203,12 +203,12 @@ def increment_activity(
...
     slack_client=None,
 ):
-    # increment activity for user
+    """Increments the participant's activity counter."""
     participant = participant_service.get_by_incident_id_and_email(
         db_session=db_session, incident_id=incident_id, email=user_email
     )
 
-    # member join also creates a message but they aren't yet a participant
+    # the member_joint event also creates a message, but they're not a participant yet
```
forestmonster, line 211: nit: `member_join`

src/dispatch/plugins/dispatch_slack/events.py
```diff
@@ -193,7 +193,7 @@ def handle_reaction_added_event(
 
 
 @slack_background_task
-def increment_activity(
+def increment_participant_activity(
```
forestmonster, line 196: Should we examine the activities that increment [Participant.activity](https://github.com/Netflix/dispatch/blame/master/src/dispatch/participant/models.py#L48)?

forestmonster (approved):
question about what qualifies incrementing Participant.activity.

src/dispatch/plugins/dispatch_slack/events.py
mvilanova: Will fix.

src/dispatch/plugins/dispatch_slack/events.py
mvilanova: Activities in and outside the incident channel other than messages? I haven't given it a lot of thought, but if you have ideas on other activities we could programmatically track let me know.

src/dispatch/database/revisions/tenant/versions/2022-10-26_4b65941d065a.py
```diff
@@ -0,0 +1,32 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "4b65941d065a"
+down_revision = "3b0f5b81376f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
mvilanova, line 20: Need to add migration code.

src/dispatch/plugins/dispatch_slack/events.py
forestmonster: Document accesses (opens) or changes might be a useful signal. If you're interacting with an incident, I think you're very likely to hit the documents.

mvilanova: Yeah, we could look into it, but history tells us that the drive activity api is not very reliable.

PR #2631, opened 2022-10-27
Enhancement/signal deduplication

kevgliss:
This PR refactors the current signal model to make it more flexible. Breaks the current model into two (now `signal_instance` and `signal`). The `signal` model contains metadata about the signal and how we might process it (e.g. what case should be opened, when the signal should be marked as duplicate or suppressed). 

Additionally, I've condensed the UI for signal `rules`, opting for them to be managed directly beside the the other signal metadata. 

Screen Shot 2022-10-27 at 10 40 44 AM
Screen Shot 2022-10-27 at 10 40 26 AM
Screen Shot 2022-10-27 at 10 40 00 AM


sfault (approved):
lgtm

PR #2635, opened 2022-10-28
Sorted PD Schedule names

jneelamekam:
Sorted PagerDuty schedules that show while using `dispatch-engage-oncall` command

### DIFF
--- a/src/dispatch/service/service.py
+++ b/src/dispatch/service/service.py
@@ -143,6 +143,7 @@ def get_all_by_project_id_and_status(
         db_session.query(Service)
         .filter(Service.project_id == project_id)
         .filter(Service.is_active.is_(is_active))
+        .order_by(Service.name)
     )
 
 


src/dispatch/service/service.py
```diff
@@ -143,6 +143,7 @@ def get_all_by_project_id_and_status(
         db_session.query(Service)
         .filter(Service.project_id == project_id)
         .filter(Service.is_active.is_(is_active))
+        .order_by(Service.name)
```
mvilanova, line 146: Is it ordering them by name ASC or DESC? We may want to change it to `.order_by(asc(Service.name))` and add import `desc` by adding `from sqlalchemy import asc` if DESC (see https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.asc).

mvilanova: If it's already in ASC order then we don't need to do anything.

mvilanova (approved):

src/dispatch/service/service.py
jneelamekam: It is in ASC order.

PR #2650, opened 2022-11-02
[WIP] Bugfix/idle transactions

kevgliss:
Observed quite a lot of open idle sessions, there are some cases (especially within the slack code) that we haven't be careful enough about closed db_sessions when we are done with them.


src/dispatch/plugins/dispatch_slack/menus.py
```diff
@@ -39,7 +42,9 @@ async def handle_slack_menu(*, config, client: WebClient, request: Request, orga
         db_session = get_organization_scope_from_slug(organization)
 
     f = menu_functions(action_id)
-    return f(db_session, user_id, user_email, channel_id, incident_id, query_str, request)
+    menu = f(db_session, user_id, user_email, channel_id, incident_id, query_str, request)
+    db_session.close()
```
mvilanova, line 46: Should we close the session if `f` needs it?

kevgliss: It's only closed after f returns, so it would be able to use it.

mvilanova: Ah, never mind. Read the code too fast.

PR #2653, opened 2022-11-03
Moves signal definitions to its own settings section

mvilanova:


kevgliss:
Should we remove the general section altogether? I'm not sure why `workflows` would fit in the general section but `signals` would not.

mvilanova:
If we remove the General section, where do we put plugins since they apply to incidents and cases? Same with workflows. I also kept notifications there for when we have support for them in cases. I don't see signals applying to both cases and incidents.

kevgliss:
Should signals be under the case settings then? They map directly to cases, there is no reason they couldn't map directly to incidents in the future, but that is not the current use case.

mvilanova:
Sure, unless you're planning to add more signals-related configuration settings?

kevgliss:
> Sure, unless you're planning to add more signals-related configuration settings?

I can't think of any right now that wouldn't be directly in the signal definition itself.

PR #2655, opened 2022-11-03
Disable Run Workflow option if incident is closed or case is escalated or closed

mvilanova:


src/dispatch/static/dispatch/src/case/Table.vue
```diff
@@ -89,7 +89,10 @@
                   >
                     <v-list-item-title>View / Edit</v-list-item-title>
                   </v-list-item>
-                  <v-list-item @click="showRun({ type: 'case', data: item })">
+                  <v-list-item
+                    @click="showRun({ type: 'case', data: item })"
+                    :disabled="item.status == 'Escalated' || item.status == 'Closed'"
```
mvilanova, line 94: @kevgliss I think that makes sense for cases, but let me know if you only want to disable it when they're closed.

kevgliss: lgtm

Issue #2566, opened 2022-09-30
Proposal: turn vuejs build-time env variables into runtime variables


caioaao:
**Is your feature request related to a problem? Please describe.**
This would make deploying and maintaining it a lot easier. It's unusual for a project like this to need to be built by the users, and it's a cumbersome process for the uninitiated. 
Another key improvement is it would be a lot easier to maintain a "canonical" docker image for the project, which I think it's lacking.

**Describe the solution you'd like**
Instead of injecting the configuration into the vue app during build, store this configuration in a path within the HTTP server. This way users wouldn't need to build the project to inject their config into the dashboard. A simple JSON map would suffice.
This could also be done in a backwards compatible way by still supporting the build-time env variables with a higher priority when constructing the final app config.

**Describe alternatives you've considered**
Separating the project into the dashboard and the headless back-end. While I think this would be a cleaner approach, I get the appeal of having a single docker image to serve everything.

If this is something the maintainers would approve, I'd be happy to implement.

kevgliss:
Hi @caioaao!

> It's unusual for a project like this to need to be built by the users, and it's a cumbersome process for the uninitiated.

I agree this is not ideal and is likely an artifact of our CI/CD being set up when we first started this project and largely going untouched. 

> Another key improvement is it would be a lot easier to maintain a "canonical" docker image for the project, which I think it's lacking.

I also agree with this point; for better or worse our docker images have been "best effort" in nature. 

> Separating the project into the dashboard and the headless back-end. While I think this would be a cleaner approach, I get the appeal of having a single docker image to serve everything.

Can you elaborate on what you are thinking here? Would this mean two separate code repos or simply two docker images? We don't use these images internally, so I think we are generally open to making changes that allow for easier consumption for others. 

cc @mvilanova

caioaao:
Hi @kevgliss , thanks for the blazing fast response!
> Can you elaborate on what you are thinking here? Would this mean two separate code repos or simply two docker images? We don't use these images internally, so I think we are generally open to making changes that allow for easier consumption for others.

I don't think code organization is that important, but the dashboard could be an independent and optional deployable - I'm just experimenting with dispatch, so correct me if I'm wrong, but it's possible to use it without ever touching the dashboard, correct? It also seem to be what makes the docker image so heavy, so splitting that would also be beneficial
That being said, the split wouldn't solve the "build source to configure" issue (although it should reduce confusion around this?) so I'd consider it as a separate enhancement

PR #2664, opened 2022-11-07
Adds the way to configure client settings via an API call

kevgliss:
This change allows us to configure client settings (e.g. auth provider) via a call to the /api/v1/settings endpoint, before the frontend is initialized. 

This has several advantages, removing build time envvars. 

See: #2566


mvilanova:
Is there anything in particular you want me to review? Perhaps we can sync and you can walk me through the changes?

PR #2659, opened 2022-11-04
Adding a settings filter for project settings

kevgliss:
Screenshot 2022-11-04 at 3 29 06 PM
Screenshot 2022-11-04 at 3 28 56 PM

PR #2665, opened 2022-11-07
Bugfix/filter setting only

kevgliss:
This PR fixes #2659, ensuring that filtering is only enabled for project settings subMenu's and not others (e.g. dashboard).


mvilanova (approved):
Lgtm

PR #2666, opened 2022-11-07
Don't attempt to sync triggers automatically on database upgrade

kevgliss:
When fulltext search triggers already exist, syncing new triggers can be problematic as the functions and expressions are tightly coupled with the columns they monitor. This change removes the automatic syncing functionality to make migrations more reliable. From now on if you add a new column that requires fulltext search functionality, you will have to manually sync the trigger in the migration file itself. See:

https://sqlalchemy-searchable.readthedocs.io/en/latest/alembic_migrations.html#sqlalchemy_searchable.sync_trigger


mvilanova (approved):
We should document how to manually sync the trigger in the db revision file in our internal docs.

PR #2670, opened 2022-11-07
Ensures we only show enabled incident types

kevgliss:
Also fixes case types as well.

### DIFF
--- a/src/dispatch/static/dispatch/src/incident/type/IncidentTypeSelect.vue
+++ b/src/dispatch/static/dispatch/src/incident/type/IncidentTypeSelect.vue
@@ -91,7 +91,7 @@ export default {
       this.loading = "error"
 
       let filterOptions = {
-        sortBy: ["project.name"],
+        sortBy: ["name"],
         descending: [false],
         itemsPerPage: this.numItems,
       }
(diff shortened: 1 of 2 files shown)


src/dispatch/static/dispatch/src/incident/type/IncidentTypeSelect.vue
```diff
@@ -91,7 +91,7 @@ export default {
       this.loading = "error"
 
       let filterOptions = {
-        sortBy: ["project.name"],
```
mvilanova, line 94: Can we revert this change? The idea is to sort the results by the project name if a project hasn't been provided for filtering purposes, otherwise by incident type name (line 106).

mvilanova: This is useful when a case needs to be mapped to an incident type.

kevgliss: > This is useful when a case needs to be mapped to an incident type, as we don't pass the prop project.

Is this required to support cross-project escalation? Should we be more explicit about this and make the user select the project before selecting the type?

kevgliss: The common case (e.g. the default) would be to escalate within the current project, but we could also allow them to override which project they want to escalate to.

mvilanova: That would work too. We could add a selector for project in the form.

kevgliss: Added via: #2673

PR #2681, opened 2022-11-09
Fixes validation that allowed multiple FAQs to be defined

kevgliss:


src/dispatch/document/service.py
```diff
@@ -64,10 +66,32 @@ def get_all(*, db_session) -> List[Optional[Document]]:
...
+                        ExistsError(
+                            msg="FAQ document already defined for this project.",
+                            document=faq_doc.name,
+                        ),
+                        loc="document",
+                    )
+                ],
+                model=DocumentCreate,
+            )
+
```
mvilanova, line 94: We should move this to https://github.com/Netflix/dispatch/blob/master/src/dispatch/document/flows.py#L18 when the incident flows code is refactored to use `create_document` in the document/flows.py module.

mvilanova (approved):

PR #2778, opened 2022-12-17
enhancement/page_case_assignee

wssheldon:
"Page Assignee" may be misleading as the link is tied to the `CaseType` and not the `Assignee` of the case.

[Table.vue](https://github.com/Netflix/dispatch/compare/enhancement/page_case_assignee?expand=1#diff-630d2f8105efc2b6515d652ab8e08216166ef393aab681122580e962b1255e75) auto-formats for me like seen in the PR upon saving the file.


kevgliss:
This looks good. One thing that's missing is the ability to set the checkbox via the new/edit modal here:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/static/dispatch/src/case/priority/NewEditSheet.vue 

Once we get that in place, this should be good to merge.

kevgliss (approved):

PR #2767, opened 2022-12-13
Enhancement/slack bolt

kevgliss:
This PR refactors the current slack functionality to use the [bolt](https://github.com/slackapi/bolt-python) and [blockkit](https://github.com/imryche/blockkit) libraries. 

@mvilanova, this isn't perfect and still needs some more testing. But I would appreciate if you took a look and maybe did some testing if you had some cycles available. 

I've been mostly using socket mode as it's a bit easier to add commands in my dev environment.

PR #2784, opened 2022-12-20
Enhancement/bolt bookmarks

wssheldon:
Migrates Slack API call from `make_call()` to `slack_bolt` SDK.

Requires `bookmarks:write` scope.

### Testing

Screenshot 2022-12-20 at 10 55 57 AM

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -542,6 +542,58 @@ def set_conversation_topic(incident: Incident, db_session: SessionLocal):
         log.exception(e)
 
 
+def set_conversation_bookmarks(incident: Incident, db_session: SessionLocal):
+    """Sets the conversation bookmarks."""
+    if not incident.conversation:
+        log.warning("Conversation bookmark not set. No conversation available for this incident.")
+        return
+
+    plugin = plugin_service.get_active_instance(
+        db_session=db_session, project_id=incident.project.id, plugin_type="conversation"
+    )
+    try:
+        plugin.instance.set_bookmark(
+            incident.conversation.channel_id,
+            resolve_attr(incident, "incident_document.weblink"),
+            title="Incident Document",
+        ) if incident.documents else log.warning(
+            "Document bookmark not set. No document available for this incident."
+        )
+
+        plugin.instance.set_bookmark(
+            incident.conversation.channel_id,
+            resolve_attr(incident, "conference.weblink"),
+            title="Video Conference",
+        ) if incident.conference else log.warning(
+            "Conference bookmark not set. No conference available for this incident."
+        )
+
+        plugin.instance.set_bookmark(
+            incident.conversation.channel_id,
+            resolve_attr(incident, "storage.weblink"),
+            title="Storage",
+        ) if incident.storage else log.warning(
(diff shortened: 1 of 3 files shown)


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -100,6 +101,11 @@ def create_slack_client(config: SlackConversationConfiguration, run_async: bool
     return AsyncWebClient(token=config.api_bot_token.get_secret_value())
 
 
+def create_bolt_slack_app(config: SlackConversationConfiguration):
```
kevgliss, line 104: The bolt app uses the `AsyncWebClient` I don't think we should create a whole bolt app just to use it's client. We have a function that creates a client for us to use in non "slack app" contexts (e.g. we just need to send a message based on the app itself, not based on some slack interaction). 

https://github.com/Netflix/dispatch/blob/e279cf6a49714d2d9b38eec9e21cb3d81dfbd815/src/dispatch/plugins/dispatch_slack/service.py#L17

kevgliss (approved):

src/dispatch/incident/flows.py
```diff
@@ -823,6 +875,42 @@ def incident_create_flow(*, organization_slug: str, incident_id: int, db_session
...
+                )
+            except Exception as e:
+                event_service.log_incident_event(
+                    db_session=db_session,
+                    source="Dispatch Core App",
+                    description=f"Incident documents rendering failed. Reason: {e}",
+                    incident_id=incident.id,
+                )
+                log.exception(e)
+
```
mvilanova, line 913: @wssheldon the conversation links in the external ticket and incident document won't render correctly as the conversation has not been created at this point. We should move this code back to where it was and then call `set_conversation_topic(incident, db_session)` and `set_conversation_bookmarks(incident, db_session)` after it instead of inside the `try` where we create the conversation.

src/dispatch/incident/flows.py
```diff
@@ -542,6 +542,58 @@ def set_conversation_topic(incident: Incident, db_session: SessionLocal):
...
+def set_conversation_bookmarks(incident: Incident, db_session: SessionLocal):
+    """Sets the conversation bookmarks."""
+    if not incident.conversation:
+        log.warning("Conversation bookmark not set. No conversation available for this incident.")
+        return
+
+    plugin = plugin_service.get_active_instance(
+        db_session=db_session, project_id=incident.project.id, plugin_type="conversation"
+    )
+    try:
```
mvilanova, line 554: We may want to check if `plugin` exists here, and if not, log a warning and return (see https://github.com/Netflix/dispatch/pull/2784/files#diff-9f1e8b67dd91333ce78c91db2036e284006cf54fa05595c9e43b0b66451d628fR122 as an example).

src/dispatch/incident/flows.py
wssheldon: Ah yup, good catch, thank you. I've re-arranged things in a new PR (https://github.com/Netflix/dispatch/pull/2799/) and will merge that into https://github.com/Netflix/dispatch/pull/2767, since this one was already merged.

PR #2788, opened 2022-12-21
Bugfix/incident-endpoint-performance

kevgliss:
They query performance for /incidents is quite poor. Via nested pedantic relationships we are loading a cascading set of attributes we don't need. This PR introduced the idea of "minimal" models that try to not load any unnecessary sub-models. This along with modifying some relationships' lazy loading to use subqueries provided a 10x reduction in API response time (loading 500 incidents). 

There are likely similar improvements to be back for cases and other endpoints. Should the need arise. I've removed several "nested" models that were similar to "minimal" but never actually used.

### DIFF
--- a/src/dispatch/incident/models.py
+++ b/src/dispatch/incident/models.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from collections import Counter, defaultdict
-from typing import List, Optional, Any
+from typing import List, Optional, ForwardRef
 
 from pydantic import validator
 from sqlalchemy import (
@@ -27,18 +27,15 @@ from dispatch.group.models import Group
 from dispatch.incident.priority.models import (
     IncidentPriorityBase,
     IncidentPriorityCreate,
-    IncidentPriorityRead,
     IncidentPriorityReadMinimal,
 )
 from dispatch.incident.severity.models import (
     IncidentSeverityCreate,
-    IncidentSeverityRead,
     IncidentSeverityReadMinimal,
     IncidentSeverityBase,
 )
 from dispatch.incident.type.models import (
     IncidentTypeCreate,
-    IncidentTypeRead,
     IncidentTypeReadMinimal,
     IncidentTypeBase,
 )
@@ -273,6 +270,9 @@ class IncidentCreate(IncidentBase):
     tags: Optional[List[TagRead]] = []
 
 
+IncidentReadMinimal = ForwardRef("IncidentReadMinimal")
+
+
 class IncidentReadMinimal(IncidentBase):
     id: PrimaryKey
     closed_at: Optional[datetime] = None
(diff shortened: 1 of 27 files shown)


src/dispatch/incident/models.py
```diff
@@ -273,10 +273,33 @@ class IncidentCreate(IncidentBase):
     tags: Optional[List[TagRead]] = []
 
 
+class IncidentReadMinimal(IncidentBase):
```
mvilanova, line 276: Looks like `cases` was dropped in this version? I think that could impact our ability to display related cases in the details tab in the UI and potentially in other places.

src/dispatch/incident/models.py
```diff
@@ -273,10 +273,33 @@ class IncidentCreate(IncidentBase):
...
+    participants_location: Optional[str]
+    participants_team: Optional[str]
+    project: ProjectRead
+    reported_at: Optional[datetime] = None
+    reporter: Optional[ParticipantReadMinimal]
+    reporters_location: Optional[str]
+    stable_at: Optional[datetime] = None
+    tags: Optional[List[TagReadMinimal]] = []
+    total_cost: Optional[float]
+    duplicates: Optional[List[Any]] = []
```
mvilanova, line 295: `IncidentReadMinimal` instead of `Any`?

src/dispatch/incident/models.py
kevgliss: It shouldn't; this should only be used for bulk listing incidents. afik we only display case data when we load incidents individually. If there is a place where we display case data alongside bulk incident information, we should be wary of if we need to present it as the incident <-> case SQL join is quite expensive.

mvilanova: I've tested the changes and they're causing related cases and other data needed in the different tabs in the details view to _not_ load, because we don't use `IncidentRead` when we view incident details. We use `IncidentReadMinimal`. Same is happening with cases.

Screenshot 2023-01-09 at 10 11 47 AM
Screenshot 2023-01-09 at 10 12 22 AM
Screenshot 2023-01-09 at 10 12 29 AM

mvilanova: Looks like we already tried this in https://github.com/Netflix/dispatch/commit/0dacd7dec2c0bc4c33f55495cbf47086d62b0528 about a year ago and I had to swap `IncidentReadMinimal` with `IncidentRead` in https://github.com/Netflix/dispatch/commit/20b41b4cd7eb0e9afceb51109304a5d529a74f78 because we ran into the same issue.

mvilanova: I think one way to fix this would be to issue an API query to get the single incident or case when we click view/ edit in the UI, which would then use the `IncidentRead` Pydantic model (https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident/views.py#L86). We would still need to add `cases` to the `IncidentRead` model, and `incidents` in the `CaseRead` (missing) model?

kevgliss: Okay, I'm not sure why we are not fetching an individual item on edit. Let me see if I can make that change.

kevgliss: I have updated the PR to ensure that we do a complete fetch instead of relying on data from the bulk endpoint.

mvilanova: Did the same for cases and changed the typing and added missing fields in the case pydantic models.

PR #2793, opened 2022-12-30
Checks for auth token on app refresh

vindex10:
Hi, All!

I noticed that, at least, in case of basic auth, the token stored in the storage is not checked for. Therefore, upon refresh user needs to login again.

I suggest some potential change that fixes this. I'm a newcomer to dispatch, so I expect I will learn something new soon :)

Look forward to your feedback!

Cheers,
Victor

### DIFF
--- a/src/dispatch/static/dispatch/src/auth/store.js
+++ b/src/dispatch/static/dispatch/src/auth/store.js
@@ -15,14 +15,37 @@ const getDefaultSelectedState = () => {
   }
 }
 
+function authCurrentUser (base, token, bad_token_ok=false) {
+  let decodedToken
+  let loggedIn
+  try {
+    decodedToken = jwt_decode(token)
+    loggedIn = true
+  } catch (e) {
+    if (!bad_token_ok) {
+      throw e
+    }
+    token = null
+    decodedToken = {}
+    loggedIn = false
+  }
+
+  return {
+    ...base,
+    loggedIn: loggedIn,
+    token: token,
+    ...decodedToken
+  }
+}
+
 const state = {
-  currentUser: {
+  currentUser: authCurrentUser({
     loggedIn: false,
     token: null,
     email: "",
     projects: [],
     role: null,
-  },
+  }, localStorage.getItem("token"), true),
   selected: {
     ...getDefaultSelectedState(),
   },
@@ -178,12 +201,7 @@ const mutations = {
     state.loading = value
   },
   SET_USER_LOGIN(state, token) {
-    state.currentUser = {
-      ...state.currentUser,
-      ...jwt_decode(token),
-      token: token,
-      loggedIn: true,
-    }
+    state.currentUser = authCurrentUser(state.currentUser, token)
     localStorage.setItem("token", token)
   },
   SET_USER_LOGOUT(state) {
--- a/src/dispatch/static/dispatch/src/auth/store.js
+++ b/src/dispatch/static/dispatch/src/auth/store.js
@@ -15,7 +15,7 @@ const getDefaultSelectedState = () => {
   }
 }
 
-function authCurrentUser (base, token, bad_token_ok=false) {
+function authCurrentUser(base, token, bad_token_ok = false) {
   let decodedToken
   let loggedIn
   try {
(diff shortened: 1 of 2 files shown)


mvilanova:
Thanks for the PR, @vindex10 . I'll take a look and circle back soon.

mvilanova:
Changes look good to me. I'll let @kevgliss take a look at them before merging them.

kevgliss:
@vindex10 Could you explain a little more about the behavior you are seeing? Locally, I can't replicate what you describe (a page refresh does not force login if the user has a valid token).

If the app state gets wiped (e.g. by a refresh), we explicitly test for an existing token here and use if it's available:
https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/auth/basicAuthProvider.js#L4

Local storage should not be wiped out by a refresh (a user needs to do a hard refresh or otherwise delete this data).

vindex10:
OK, thanks! indeed it works. For dev, I had to:

```
export VITE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG=dispatch-auth-provider-basic
```
For prod I had to change docker-compose.yaml:

```
web:
    build:
        context: ../dispatch
        args:
            VITE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG: "dispatch-auth-provider-basic"
```

And basic auth is enabled by default in the backend? So should this become a part of:
https://github.com/Netflix/dispatch-docker

vindex10:
Hmm. It actually didn't work with docker compose. And I have a suspect.

In the setup.py, it looks like the `env` is not propagated to the npm command:
https://github.com/Netflix/dispatch/blob/e52af851a6f8487ceb5ffefb7e59c57e9cb11145/setup.py#L147

and build_static that used to build the frontend before, doesn't do it anymore:
https://github.com/Netflix/dispatch/blob/e52af851a6f8487ceb5ffefb7e59c57e9cb11145/setup.py#L306

but it actually handled env vars before.

vindex10:
sorry for the noise :) in summary:

1. I was wrong. env in setup py is fine.
2. I had to change build-args for `core` not for `web`
3. I corrected my messages above so that they look less confusing

I'm trying to understand now, which PR to open :) I see the options:

1. Set default arg in the Netflix/dispatch dockerfile (moderate intrusive option)
2. Set build arg in the Netflix/dispatch-docker, in the docker-compose.yaml (the least intrusive option)
3. Set default in router (we are consistent with backend, and api.js)
4. just mention in the docs that empty auth slug is a feature and how to use it + comment that for the basic auth you need to specify the variable VITE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG explicitly during build
5. do nothing :) 

What confuses me is the slight inconsistency between:

api.js: https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/api.js#L12
where default auth is assumed to be "basic auth"

router/index.js: https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/router/index.js#L39
here empty authentication slug falls back to custom auth (which seems to be a feature, not a bug)

backend config.py: https://github.com/Netflix/dispatch/blob/17e1a64a498de53eb84b3d4ac27ec32a045636c8/src/dispatch/config.py#L100


@kevgliss look forward to your feedback )

kevgliss:
I think your right when you point out the inconsistency here:
https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/router/index.js#L39

I think the change should align the default to basic-auth similar to what we do here:
https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/api.js#L12

So, in summary I think we should define:
```
const authProviderSlug =
  import.meta.env.VITE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG || "dispatch-auth-provider-basic"
 ```
here: 
https://github.com/Netflix/dispatch/blob/138c983c2de5fb42e92712efca9b90c04cf645fc/src/dispatch/static/dispatch/src/router/index.js#L20

vindex10:
@kevgliss done. I checked, at least it worked when I ran locally :)

kevgliss:
LGTM, thanks for the contribution!

PR #2800, opened 2023-01-04
Add participants to the Case model and add Case permissions

ghost:
Initial skeleton for case participants. This PR adds participants to the case model. Additionally, this adds the participants tab to the case edit sheet to remain consistent between incident and case front-end, and finally adds permissions to cases.

Screenshot 2023-01-30 at 11 42 28 AM

Screenshot 2023-02-01 at 4 38 13 PM


src/dispatch/case/models.py
```diff
@@ -106,6 +110,16 @@ class Case(Base, TimeStampMixin, ProjectMixin):
...
 
+    participants = relationship(
+        Participant,
+        backref="case",
+        cascade="all, delete-orphan",
+        foreign_keys=[Participant.case_id],
+    )
+
+    observer_id = Column(Integer, ForeignKey("participant.id"))
+    observer = relationship(Participant, foreign_keys=[observer_id], post_update=True)
```
kevgliss, line 121: Do we need an observer for cases? I'm not sure how you determine if someone was an observer. 

For incidents, we define an observer as someone in the channel who doesn't say anything. Since cases will be threaded messages, you could view the thread and not say anything... but I don't think Dispatch would be able to tell that is the case.

src/dispatch/participant/flows.py
```diff
@@ -13,58 +15,75 @@
 
 log = logging.getLogger(__name__)
 
+Obj = TypeVar("Obj", Case, Incident)
+
 
 def add_participant(
     user_email: str,
-    incident: Incident,
+    obj: Obj,
```
kevgliss, line 23: When a function could be either an `Incident` or a `Case`. I have used `subject` as the variable name as it's a bit more descriptive.

src/dispatch/case/service.py
```diff
@@ -187,6 +189,24 @@ def create(*, db_session, case_in: CaseCreate, current_user: DispatchUser = None
         case_id=case.id,
     )
 
+    # add reporter
```
kevgliss, line 192: We also need to ensure that `case.assignee` is a proper participant e.g. https://github.com/Netflix/dispatch/pull/2800/files#diff-85562e526dc8615124aead4d1b1dc02d9fbfdbc211396e4c880f3e63f354fb00R156

We will likely need some data migration for that as well (e.g. moving from email to participant_role_id).

kevgliss:
Looks good so far!

src/dispatch/case/models.py
```diff
@@ -180,6 +199,8 @@ class CaseCreate(CaseBase):
     case_severity: Optional[CaseSeverityRead]
     case_type: Optional[CaseTypeRead]
     project: Optional[ProjectRead]
+    reporter: Optional[ParticipantUpdate]
```
mvilanova, line 202: The Pydantic models have changed. You may want to merge the changes in `master` and add reporter, observer, and participants to the CaseReadMinimal model.

src/dispatch/case/models.py
mvilanova: If we determine that we need case participants with an observer (aka subscriber or follower) role we should leverage the observer participant role we already have available. Perhaps observers can be added via engagement filters and managed using a Slack command? There might be people who are interested in keeping an eye on specific case types.

src/dispatch/case/service.py
```diff
@@ -187,6 +189,24 @@ def create(*, db_session, case_in: CaseCreate, current_user: DispatchUser = None
...
+
+    # add observer
+    observer_email = case_in.observer.individual.email
+    if observer_email:
+        participant_flows.add_participant(
+            observer_email,
+            case,
+            db_session,
+            role=ParticipantRoleType.observer,
+        )
```
mvilanova, line 208: I think observer(s) would make more sense to be resolved via engagement filters and added to the case after the case is created in the same way we do it for incidents rather than having one participant with observer role being added here. Also, who would that be on case creation?

src/dispatch/static/dispatch/src/case/DetailsTab.vue
```diff
@@ -46,6 +46,15 @@
...
       </v-flex>
+      <v-flex xs6>
+        <participant-select
+          v-model="observer"
+          label="Observer"
+          hint="A participant who can observe the case."
+          clearable
+          :project="project"
+        />
+      </v-flex>
```
mvilanova, line 57: We may want to add more than one.

src/dispatch/participant/flows.py
mvilanova: I used `obj` but I don't have strong opinions between one vs other as long as we're consistent across the codebase.

src/dispatch/case/models.py
wssheldon: I'm thinking about scrapping the observer role for this PR and just getting participants into the model. Can add observer or similar in a follow up PR. It's also unclear to me how observer would work from a Slack perspective when these are in threads.

src/dispatch/participant/flows.py
wssheldon: I'll standardize on `subject` to keep it consistent.

src/dispatch/case/service.py
wssheldon: Addressed here https://github.com/Netflix/dispatch/pull/2800/commits/f5d778bcb5b91abae8b87ea64af7ab08c67e89e0 will need to a do a data migration for existing cases though.

src/dispatch/case/service.py
wssheldon: Removed observer role for now.

src/dispatch/auth/permissions.py
```diff
@@ -291,3 +293,84 @@ def has_required_permissions(
...
+            db_session=request.state.db, incident_id=request.path_params["case_id"]
+        )
+        if not current_case:
+            return False
+
+        if current_case.assignee:
+            if current_case.assignee.individual.email == current_user.email:
+                return True
+
+
```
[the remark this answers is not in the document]
wssheldon: Updated to return `False` for some of these cases here: https://github.com/Netflix/dispatch/pull/2800/commits/a1329f35ce81aeec00f9e7fd375617357a9bb14f

What did you mean by "Idem"?

src/dispatch/auth/permissions.py
```diff
@@ -291,3 +293,71 @@ def has_required_permissions(
...
+        request: Request,
+    ) -> bool:
+        current_case = case_service.get(
+            db_session=request.state.db, case_id=request.path_params["case_id"]
+        )
+
+        if not current_case:
+            return False
+
+        if current_case.visibility == Visibility.restricted:
```
kevgliss, line 313: Shouldn't any case participant have access to view the case?

src/dispatch/auth/permissions.py
```diff
@@ -291,3 +293,71 @@ def has_required_permissions(
...
+    ) -> bool:
+        current_user = get_current_user(request=request)
+        current_case: Case = case_service.get(
+            db_session=request.state.db, incident_id=request.path_params["case_id"]
+        )
+        if not current_case:
+            return False
+
+        if current_case.assignee:
+            if current_case.assignee.individual.email != current_user.email:
```
kevgliss, line 338: These could be written more clearly but defining a `has_permission = False` at the beginning and then only having conditionals that would change modify `has_permission = True` and returning the value of `has_permission`.

src/dispatch/group/flows.py
```diff
@@ -52,41 +52,45 @@ def create_group(
...
-        obj.notifications_group_id = group.id
+        subject.notifications_group_id = group.id
 
-    db_session.add(obj)
+    db_session.add(subject)
     db_session.commit()
 
-    obj_type = get_table_name_by_class_instance(obj)
-    if obj_type == "case":
+    subject_type = get_table_name_by_class_instance(subject)
```
kevgliss, line 65: I think event logging should take place in the calling function e.g. `create_case_flow` and `create_incident_flow`

src/dispatch/participant/flows.py
```diff
@@ -13,57 +15,75 @@
 
 log = logging.getLogger(__name__)
 
+Subject = TypeVar("Subject", Case, Incident)
+
 
 def add_participant(
```
kevgliss, line 21: This feels like we may want to eventually, break up into `add_incident_participant` and `add_case_participant`. Since the overlap is high. I am okay leaving them combined for now.

src/dispatch/participant/service.py
```diff
@@ -79,34 +79,43 @@ def get_all_by_incident_id(*, db_session, incident_id: int) -> List[Optional[Par
...
 ) -> Participant:
     """Gets an existing participant object or creates a new one."""
-    from dispatch.incident import service as incident_service
-
-    participant = (
+    participant: Participant = (
         db_session.query(Participant)
-        .filter(Participant.incident_id == incident_id)
+        .filter(
+            Participant.incident_id == subject_id
```
kevgliss, line 92: Not a huge of fan of doing this inline due to readable concerns. Especially as the query building allows you to do this pretty easily. 

```
query = db_session.query(Participant)

if subject_type == 'incident':
     query = query.filter(Participant.incident_id == subject_id)
 else:
     query = query.filter(Participant.case_id == subject_id)
```

src/dispatch/participant/service.py
```diff
@@ -79,34 +79,43 @@ def get_all_by_incident_id(*, db_session, incident_id: int) -> List[Optional[Par
...
+        if subject_type == "incident":
+            from dispatch.incident import service as incident_service
 
-        # We get information about the individual
-        individual_contact = individual_service.get(
-            db_session=db_session, individual_contact_id=individual_id
-        )
+            subject = incident_service.get(db_session=db_session, incident_id=subject_id)
+        if subject_type == "case":
+            from dispatch.case import service as case_service
```
kevgliss, line 106: Is this causing you circular import issues?

kevgliss (commented):
Have you thought about how we would organically add participants? I'm thinking that anyone that is either mentioned or messages the conversation thread should be added as a case participant.

wssheldon:
> Have you thought about how we would organically add participants? I'm thinking that anyone that is either mentioned or messages the conversation thread should be added as a case participant.

@kevgliss and I chatted about this on VOIP. We will start with mentions within a Case thread being used to add people as a participant. Kevin to add this logic to https://github.com/Netflix/dispatch/pull/2913 and rebase off this PR (https://github.com/Netflix/dispatch/pull/2800).

src/dispatch/auth/permissions.py
wssheldon: Much cleaner, thanks!

src/dispatch/auth/permissions.py
wssheldon: I think so, I'll add another permission that handles this for all participants (and keep the existing reporter/assignee ones in case we want to keep those to have different permission models.)

wssheldon: Scratch that, I just switched it to be "if you're a participant, you can view the case, and edit it."

src/dispatch/participant/service.py
wssheldon: No, I moved the imports (and there's still no circular import) to the top of the file since that pattern isn't necessary.

src/dispatch/participant/flows.py
wssheldon: I'd be okay with breaking it out. The function can be hard to follow/read (for me) because of it handling two types.

src/dispatch/group/flows.py
wssheldon: I moved the logging out to the calling function. I could see value in having logs in the method that was called (or both places), this case `update_group`, so people get some logging for free across the board. I didn't see incident flows using the group flow here, so I just removed it as well. 

cc @mvilanova in case you have any thoughts, since I think you added these log statements.

wssheldon:
@kevgliss Are we at a stable point w/ the front-end fixes to merge this?

PR #2805, opened 2023-01-05
gmail plugin. dont fail when there are no threads with mailer daemon

vindex10:


mvilanova:
Looks good. Thanks for the PR!

PR #2807, opened 2023-01-07
individual name is wrongly resolved

vindex10:
when an incident is created, reporter's name is reset to their "email".

reasons:
1. project_id is not provided to `get` method of the contact plugin
2. `individual.service` checks for a `fullname` field while `name` is actual field of the IndividualContact table

### DIFF
--- a/src/dispatch/individual/service.py
+++ b/src/dispatch/individual/service.py
@@ -89,12 +89,10 @@ def get_or_create(
 
     individual_info = {}
     if contact_plugin:
-        individual_info = contact_plugin.instance.get(email,
-                                                      project_id=contact_plugin.project_id,
-                                                      db_session=db_session)
+        individual_info = contact_plugin.instance.get(email, db_session=db_session)
 
     kwargs["email"] = individual_info.get("email", email)
-    kwargs["name"] = individual_info.get("name", "Unknown")
+    kwargs["name"] = individual_info.get("fullname", "Unknown")
     kwargs["weblink"] = individual_info.get("weblink", "")
 
     if not individual_contact:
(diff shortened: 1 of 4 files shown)


kevgliss:
Which plugin are you using for contact information? I think this has to do with the default contact plugin:

https://github.com/Netflix/dispatch/blob/3e51e4705d67b8101869ca20a491f6e234600d68/src/dispatch/plugins/dispatch_core/plugin.py#L281

We could modify the default plugin to return the actual full name of the contact, but I wouldn't modify the `individual_service` as that would break our slack contact plugin here:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/plugin.py#L210

vindex10:
Thank you, @kevgliss !

I indeed use the default contact plugin:

maybe I could change the plugin itself, and instead of returning "IndividualContact" `__dict__`, we could rename "name" (used in IndividualContact schema) -> "fullname" (used in Individual service).
https://github.com/Netflix/dispatch/blob/e52af851a6f8487ceb5ffefb7e59c57e9cb11145/src/dispatch/plugins/dispatch_core/plugin.py#L277

it, however, may be more reasonable to fix the slack plugin. because it seems better to be consistent between individual_contact (that has `name` in the schema) and individual.service (that expects `fullname`)?

kevgliss:
Aligning on `name` sounds good to me.

vindex10:
I started replacing `fullname` with `name` and noticed that there are many places where `fullname` is assumed.

So to give this PR hope to be merged, I just fixed it in the dispatch contact plugin :)

Moreover, contact plugin requires `project_id`, and, while Plugin itself doesn't know the project_id, PluginInstance through which plugins are used, know the id but don't provide it to the plugin :) I tried to fixed this, I hope I understood the abstraction correctly.

With this change all plugin instances have access to project id if they need it. I thought it was a valid change because PluginInstance model already included ProjectMixin.

@kevgliss @mvilanova thank you for your attention!

kevgliss:
lgtm

PR #2829, opened 2023-01-11
organization switch reloads to old route

vindex10:
the issue I observe (firefox):
when organization switch button (with two arrows) is pressed, page reloads, but organization stays the same.

I tried with debugger, it showed me "this.$router.currentRoute" to still be resolved to the old org

the updated version works fine though. do you have any idea why the currentRoute is not updated on "router.push"?

somewhat relevant discussion here:
https://github.com/vuejs/vue-router/issues/311#issuecomment-497182264

upd:

false alarm. when inside `then` it seems resolving correctly (so the PR is still relevant to have `then`)


vindex10:
@mvilanova I merged master

mvilanova:
Thanks for the PR. I'll test your changes and circle back.

mvilanova:
The changes seem to enable me to switch orgs, but only one time. Once I have switched from `default` to another newly created org, it doesn't allow me to switch to another one. It's a known issue we've had for a while, but we haven't prioritized a bug fix for it.

mvilanova:
This is where we added the logic to load the user settings (e.g. default projects for filtering purposes), which ended up breaking the ability to switch orgs, but we didn't notice it because we don't use the feature.

https://github.com/Netflix/dispatch/blob/master/src/dispatch/static/dispatch/src/router/index.js#L38

vindex10:
@mvilanova , I start with two already created orgs, and try to jump between them. It works for me, and didn't work before the fix introduced in this PR.

vindex10:
@mvilanova am I right that we don't contradict each other? and my fix is just a partial fix to the issue?

vindex10:
Hi! would you find time to check this? On my side it definitely improves the situation with the organization switching.

vindex10:
@mvilanova hi! I recorded a demo of how it works. I used our testing db, but switched frontend between versions: current master and this PR.

ezgif-1-8266b169fd

backend is running from this version:
https://github.com/Netflix/dispatch/commit/e085b0ce8b467b4221e46e27a343d7c85393ed38

mvilanova:
Thanks. I should have more time to re-test today. I'll circle back.

mvilanova:
Seems to be working as expected after further testing. Thanks again for fixing this pesky bug!

PR #2833, opened 2023-01-11
removes user_middleware from global message event dispatcher

wssheldon:
the `user_middleware` is not currently utilized in any of the `message_dispatcher` functions.


src/dispatch/plugins/dispatch_slack/bolt.py
```diff
@@ -11,10 +11,10 @@
 
 from .decorators import message_dispatcher
 from .middleware import (
+    configuration_middleware,
```
wssheldon, line 14:
```suggestion
```

PR #2840, opened 2023-01-12
add env var to disable registration

vindex10:
use case: we want to use basic auth but to forbid registration for new users (still keep the authentication)

there are two env vars:

* VITE_DISPATCH_AUTH_REGISTRATION_ENABLED - used for frontend, needs to be passed to the docker file as a build arg, so that vite can hardwire it into the app dist files (disable route, hide register link)
* DISPATCH_AUTH_REGISTRATION_ENABLED - used in the backend, to disable "register" endpoint

in the `config.py`, `DISPATCH_` variable is initialized from `VITE_`. I chose it this way, because then it is enough to have only one variable for both frontend and backend. 

config.py is not used to build frontend, while vite requires `VITE_` only variables, so generally `VITE_` can't be avoided from what I can see.

I can adapt though, for your preferences :)


kevgliss:
This looks good to me. If you can fix the linking issues we can get this merged. Thanks for the PR!

PR #2848, opened 2023-01-13
bugfix/handle-slash-command-timeouts-and-error-handling

wssheldon:
The slash commands are not bolt [lazy listeners](https://slack.dev/bolt-python/concepts#lazy-listeners) like the modal functions and as a result can raise `operation_timeout` errors in Slack to the user if they take longer than 3 seconds.

```
/dispatch-list-incidents failed with the error "operation_timeout"
```

One approach is to make everything a lazy listener. One known downside to lazy listeners is they swallow errors/exceptions and are hard to debug. We have the `@handle_lazy_error` decorator as a workaround. For context, we used to [acknowledge](https://github.com/Netflix/dispatch/blob/2390cc161f1b767e686ca5a68593fe5fca44fc7e/src/dispatch/plugins/dispatch_slack/events.py#L148) each request from Slack in the [handle_slack_event](https://github.com/Netflix/dispatch/blob/2390cc161f1b767e686ca5a68593fe5fca44fc7e/src/dispatch/plugins/dispatch_slack/events.py#L113) function that [would run immediately upon](https://github.com/Netflix/dispatch/blob/2390cc161f1b767e686ca5a68593fe5fca44fc7e/src/dispatch/plugins/dispatch_slack/views.py#L119) receiving a request from Slack.

I tested that this works by running `time.sleep(4)` within the `handle_list_incidents_command` function.

If we're good with this approach, I'll update the PR to do this for all slash commands.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -424,15 +424,14 @@ async def handle_list_incidents_command(
 
 @handle_lazy_error
 async def handle_list_participants_command(
-    ack: AsyncAck,
     respond: AsyncRespond,
     client: AsyncWebClient,
-    db_session: Session,
     context: AsyncBoltContext,
 ) -> None:
     """Handles list participants command."""
-    await ack()
     blocks = [Section(text="*Incident Participants*")]
+    db_session = refetch_db_session(context["subject"].organization_slug)
+
     participants = participant_service.get_all_by_incident_id(
         db_session=db_session, incident_id=context["subject"].id
     ).all()
@@ -516,17 +515,16 @@ def filter_tasks_by_assignee_and_creator(
 
 @handle_lazy_error
 async def handle_list_tasks_command(
-    ack: AsyncAck,
     body: dict,
     payload: dict,
     client: AsyncWebClient,
     context: AsyncBoltContext,
-    db_session: Session,
     respond: AsyncRespond,
 ) -> None:
     """Handles the list tasks command."""
-    await ack()
     blocks = []
+
+    db_session = refetch_db_session(context["subject"].organization_slug)
 
     caller_only = False
     if body["command"] == context["config"].slack_command_list_my_tasks:
@@ -586,11 +584,10 @@ async def handle_list_tasks_command(
 
 
 @handle_lazy_error
-async def handle_list_resources_command(
-    ack: AsyncAck, respond: AsyncRespond, db_session: Session, context: AsyncBoltContext
-) -> None:
+async def handle_list_resources_command(respond: AsyncRespond, context: AsyncBoltContext) -> None:
     """Handles the list resources command."""
-    await ack()
+
+    db_session = refetch_db_session(context["subject"].organization_slug)
 
     incident = incident_service.get(db_session=db_session, incident_id=context["subject"].id)
 
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -141,28 +141,38 @@ def configure(config):
     ]
 
     # don't need an incident context
+    app.command(config.slack_command_report_incident, middleware=[db_middleware])(
+        handle_report_incident_command
+    )
+
+    # non-sensitive-commands
     app.command(
         config.slack_command_list_incidents,
         middleware=middleware,
     )(ack=ack_command, lazy=[handle_list_incidents_command])
 
-    app.command(config.slack_command_report_incident, middleware=middleware)(
-        handle_report_incident_command
-    )
+    app.command(
+        config.slack_command_list_tasks,
+        middleware=middleware,
+    )(ack=ack_command, lazy=[handle_list_tasks_command])
 
-    # non-sensitive-commands
-    app.command(config.slack_command_list_tasks, middleware=middleware)(handle_list_tasks_command)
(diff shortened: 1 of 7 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -619,6 +621,11 @@ async def handle_list_resources_command(
     await respond(text="Incident Resources", blocks=blocks, response_type="ephemeral")
 
 
+async def ack_command(ack: AsyncAck) -> None:
+    """Handles request acknowledgement for slash commands."""
+    await ack()
```
wssheldon, line 626: We could send a message back to the user here, if we want.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -138,9 +138,11 @@ def configure(config):
     ]
 
     # don't need an incident context
-    app.command(config.slack_command_list_incidents, middleware=[db_middleware])(
-        handle_list_incidents_command
-    )
+    app.command(
+        config.slack_command_list_incidents,
+        middleware=[db_middleware],
+    )(ack=ack_command, lazy=[handle_list_incidents_command])
```
wssheldon, line 144: To cut down on duplicate code across the modal `ack` functions, It would be nice if we could inject variables into the `ack=` function (to build the modals).

src/dispatch/plugins/dispatch_slack/incident/interactive.py
kevgliss: In order to maintain current functionality; we should probably return a message similar to what we do now.... e.g. "Your command is running... " Using the mapped command strings as part of the message:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/plugin.py#L181

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -138,27 +141,39 @@ def configure(config):
...
-    app.command(config.slack_command_list_my_tasks, middleware=middleware)(
-        handle_list_tasks_command
-    )
-    app.command(config.slack_command_list_participants, middleware=middleware)(
-        handle_list_participants_command
-    )
+    app.command(
+        config.slack_command_list_incidents,
+        middleware=middleware,
+    )(ack=ack_command, lazy=[handle_list_incidents_command])
```
wssheldon, line 152: This prevents `list-incidents` from running outside of incident channels. 

You'll get:

```
Sorry, I can't determine the correct context to run the command /dispatch-list-incidents. Are you running this command in an incident channel?
```

This is caused by the `command_context_middleware`. Which we're using to get the org_slug to refetch the db. Need a better way here.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -132,54 +137,63 @@ class MonitorMetadata(SubjectMetadata):
...
-    app.command(config.slack_command_report_incident, middleware=[db_middleware])(
-        handle_report_incident_command
+    app.command(config.slack_command_list_incidents, middleware=middleware)(
+        ack=ack_command, lazy=[handle_list_incidents_command]
     )
 
     # non-sensitive-commands
-    app.command(config.slack_command_list_tasks, middleware=middleware)(handle_list_tasks_command)
+    middleware.remove(non_incident_command_middlware)
+    middleware.extend([command_context_middleware])
```
kevgliss, line 156: This is a bit confusing, should we build a new middleware list instead of modifying it?

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -292,6 +292,42 @@ async def get_user_avatar_url_async(client: Any, email: str):
...
 
 
+Conversations = list[dict[str, str]]
+
+
+async def get_conversations_by_user_id_async(
+    client: Any, user_id: str
+) -> tuple[Conversations, Conversations]:
+    """Gets the list of public and private conversations a user is a member of."""
+    result = await make_call_async(
```
kevgliss, line 302: These calls could be slow, do they make sense to cache?

kevgliss (approved):

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -2022,7 +2030,6 @@ async def handle_report_incident_submission_event(
 
     result = await client.views_update(
         view_id=result["view"]["id"],
-        hash=body["view"]["hash"],
```
wssheldon: This is the only way I've been able to resolve the hash_conflict errors. Removing it seems counter intuitive, would be good to check this.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -942,7 +950,7 @@ async def handle_member_joined_channel(
         participant.added_by = incident.commander
 
     # Message text when someone @'s a user is not available in body, use generic added by reason
-    participant.added_reason = f"Participant added by {participant.added_by}"
+    participant.added_reason = f"Participant added by {participant.added_by.individual.name}"
```
wssheldon, line 953: Name as str vs Obj

src/dispatch/plugins/dispatch_slack/incident/interactive.py
wssheldon: Agreed, made this change, and I think it's more simple.

PR #2849, opened 2023-01-13
Fixes incident feedback button and modal

mvilanova:


src/dispatch/database/core.py
```diff
@@ -188,3 +185,13 @@ def ensure_unique_default_per_project(target, value, oldvalue, initiator):
             if previous_default.id != target.id:
                 previous_default.default = False
                 session.commit()
+
+
+def refetch_db_session(organization_slug: str) -> Session:
```
wssheldon, line 190: This function is also [here](https://github.com/Netflix/dispatch/blob/f12474bd94b546d77dac7fe406463d368b4d2d14/src/dispatch/plugins/dispatch_slack/service.py#L47-L55). But, this looks like a better spot and name for it to me.

wssheldon (approved):

kevgliss (approved):

src/dispatch/database/core.py
mvilanova: Hmm I don't see it in the `service.py` module. Did you mean the `incident/interactive.py` one? IIRC I moved from there to here.

mvilanova: Ah, I see. You were referring to `get_organization_scope_from_slug`. For some reason, Github was not highlighting it.

mvilanova: I think it'd make more sense for both `get_organization_scope_from_slug` and `get_default_organization_scope` functions in `./src/dispatch/plugins/dispatch_slack/service.py` to live in `database/core.py` or `database/service.py`? I think we should re-org those two modules. It's not very clear to me when I should add a function to one vs the other.

mvilanova: I'm going to leave things are they are and we can tackle the re-org in a separate PR.

PR #2859, opened 2023-01-18
consolidate get_conversations service method and resolve cache issue

wssheldon:
This resolves a bug where non-incident commands incorrectly reported that the bot was not present in private channels. Consolidates `get_public/private_conversations_by_user_id_async` into a single function that accepts a `type` param and generates a unique cache key for public / private channels, which is the root cause of the bug.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/service.py
+++ b/src/dispatch/plugins/dispatch_slack/service.py
@@ -308,7 +308,7 @@ async def get_conversations_by_user_id_async(client: Any, user_id: str, type: st
             types=f"{type}_channel",
             exclude_archived="true",
         )
-        await cache.set(user_id + f"-{type}", result)
+        await cache.set(f"{user_id}-{type}", result)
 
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -298,46 +298,23 @@ async def get_user_avatar_url_async(client: Any, email: str):
...
-        private_result = await make_call_async(
+async def get_conversations_by_user_id_async(client: Any, user_id: str, type: str) -> Conversations:
+    result = await cache.get(user_id)
+    if not result:
+        result = await make_call_async(
             client,
             "users.conversations",
             user=user_id,
-            types="private_channel",
+            types=type + "_channel",
```
mvilanova, line 308:
```suggestion
            types=f"{type}_channel",
```

mvilanova (approved):

PR #2860, opened 2023-01-18
Creates default incident and case type on new project creation

mvilanova:


src/dispatch/project/flows.py
```diff
@@ -80,6 +100,23 @@ def project_init_flow(*, project_id: int, organization_slug: str, db_session=Non
...
+        exclude_from_metrics=default_case_type["exclude_from_metrics"],
+        default=default_case_type["default"],
+        enabled=default_case_type["enabled"],
+        project=project,
+    )
+    case_type = case_type_service.create(db_session=db_session, case_type_in=case_type_in)
+
+    # Map case type with incident type
+    case_type.incident_type = incident_type
+    db_session.add(case_type)
```
wssheldon, line 117: Do we need to add `incident_type` ?

wssheldon (approved):
LGTM

src/dispatch/project/flows.py
mvilanova: No, it gets added and committed in its create function above.

PR #2863, opened 2023-01-19
Adds handlers for incident join and subscribe button click events

mvilanova:


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -2097,3 +2101,70 @@ async def handle_report_incident_project_select_action(
...
+
+    if not incident:
+        message = "Sorry, we can't invite you to this incident. The incident does not exist."
+    elif incident.visibility == Visibility.restricted:
+        message = "Sorry, we can't invite you to this incident. The incident's visbility is restricted. Please, reach out to the incident commander if you have any questions."
+    elif incident.status == IncidentStatus.closed:
+        message = "Sorry, you can't join this incident. The incident has already been marked as closed. Please, reach out to the incident commander if you have any questions."
+    else:
+        user_id = context["user_id"]
+        await dispatch_slack_service.add_users_to_conversation_async(
```
kevgliss, line 2132: I don't think we need this function. You could call `client.conversations_invite().`

PR #2887, opened 2023-01-24
add task button handler and modal draw function, support front-end tasks

wssheldon:
- Support tasks created in the front-end (which don't have a `resource_id`)
- Automatically update the status within the modal and live update the modal
- Move modal builder to a draw function
- Create small open and close function in `task_service`

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -1,7 +1,6 @@
 import logging
 from datetime import datetime
 from typing import Any
-from typing import Any
 
 import pytz
 from blockkit import (
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -1,6 +1,6 @@
 import logging
 from datetime import datetime
-from typing import Any, Optional
+from typing import Any
 
 import pytz
 from blockkit import (
@@ -25,8 +25,7 @@ from slack_sdk.web.client import WebClient
 from sqlalchemy import func
 from sqlalchemy.orm import Session
 
-from dispatch.auth import service as user_service
-from dispatch.auth.models import DispatchUser, UserRegister
+from dispatch.auth.models import DispatchUser
 from dispatch.config import DISPATCH_UI_URL
 from dispatch.database.core import resolve_attr
 from dispatch.database.service import search_filter_sort_paginate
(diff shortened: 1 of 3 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -2140,3 +2166,73 @@ def monitor_link_button_click(
...
+            drive_task_plugin.instance.update(file_id, button.resource_id, resolved=resolve)
+    else:
+        task_service.resolve_or_reopen(
+            db_session=db_session, task_id=task.id, status=TaskStatus.resolved
+        ) if resolve is True else task_service.resolve_or_reopen(
+            db_session=db_session, task_id=task.id, status=TaskStatus.open
+        )
+
+    draw_task_modal(
+        caller_only=None,
```
wssheldon, line 2231: I think we need to retain this value across clicks so that we draw the correct tasks.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -510,70 +511,94 @@ def handle_list_tasks_command(
...
+    )
 
     caller_only = False
+    user_id = None
     if body["command"] == context["config"].slack_command_list_my_tasks:
+        user_id = payload["user_id"]
         caller_only = True
 
+    if caller_only:
+        user_id = user_id
```
mvilanova, line 526: Isn't this a no-op? What if `user_id` is `None`, won't the `client.users_info()` function fail?

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -510,70 +511,94 @@ def handle_list_tasks_command(
...
+        email = (client.users_info(user=user_id))["user"]["profile"]["email"]
+        user = user_service.get_or_create(
+            db_session=db_session,
+            organization=context["subject"].organization_slug,
+            user_in=UserRegister(email=email),
+        )
```
mvilanova, line 532: Tasks are scoped to incidents and created and owned by participants. I think using `participant_service.get_by_incident_id_and_email()` would be more appropriate here if you want to check if an incident participant with this email address exists. If you don't need to perform this check, then you can just pass the resulting email from the `client.users_info()` call directly to `filter_tasks_by_assignee_and_creator()`

src/dispatch/task/service.py
```diff
@@ -189,6 +189,12 @@ def update(*, db_session, task: Task, task_in: TaskUpdate, sync_external: bool =
     return task
 
 
+def resolve_or_reopen(*, db_session, task_id: int, status: str) -> None:
```
mvilanova, line 192: nit: missing docstring

src/dispatch/plugins/dispatch_slack/incident/interactive.py
wssheldon: This block of code was sloppy in general, addressed this issue, and cleaned it up.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
wssheldon: Makes sense, if a task owner is always a participant, we can probably just pass the email from `client.users_info()` safely.

PR #2899, opened 2023-01-26
Fixes timeline reaction event in incident conversation

mvilanova:


src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -112,6 +112,15 @@ def message_context_middleware(request: BoltRequest, context: BoltContext, next:
     next()
 
 
+def reaction_context_middleware(context: BoltContext, payload: dict, next: Callable) -> None:
```
mvilanova, line 115: @kevgliss @wssheldon I added a middleware for reactions. There's probably an opportunity to combine this one with the `command_context_middleware` and `message_context_middleware`. The code is almost the same, just different exception error message.

wssheldon, line 115:
```suggestion
def reaction_context_middleware(context: BoltContext, next: Callable) -> None:
```

I think payload is unused argument.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -209,7 +211,7 @@ def configure(config):
     # required to allow the user to change the reaction string
     app.event(
         {"type": "reaction_added", "reaction": config.timeline_event_reaction},
-        middleware=[db_middleware],
+        middleware=[reaction_context_middleware, db_middleware],
```
mvilanova, line 214: @kevgliss @wssheldon I kept the `db_middleware`, but Dispatch did not seem to have any issues without it. There are probably some situations where it's needed that I'm not aware of.

src/dispatch/plugins/dispatch_slack/middleware.py
wssheldon: Now that I see this, we may want to rethink the global error handler logic, specifically how we pick the message for the user:

TLDR; message is determined by exception type, not message raised with the exception. 

```python
def build_and_log_error(...snip) -> str:
    if isinstance(error, RoleError):
        message = build_role_error_message(payload)
        logger.info(error)
  
    ...snip
    
    elif isinstance(error, ContextError):
        message = build_context_error_message(payload, error)
        logger.info(error)
        
    ...snip

    else:
        guid = str(uuid.uuid4())
        message = build_unexpected_error_message(guid)
        logger.exception(error, extra=dict(slack_interaction_guid=guid))

    return message
```

src/dispatch/plugins/dispatch_slack/incident/interactive.py
wssheldon: db_middleware might not be needed lots of places now that I look at it, anything that calls `resolve_context_from_conversation` should already have a `db_session` in the `subject`:

https://github.com/Netflix/dispatch/pull/2899/files#diff-be932aea97556035df57c47ea4fa56c2ca3b7441226f65595ef4ef2f48c422abR40-R51

Probably some cleanup to do in middleware in general.

src/dispatch/plugins/dispatch_slack/middleware.py
mvilanova: Good catch.

mvilanova: Should we do it before pushing to prod?

src/dispatch/plugins/dispatch_slack/incident/interactive.py
mvilanova: Let's remove it if not needed then.

PR #2907, opened 2023-01-27
instantly acknowledge /command and /event requests from slack

wssheldon:
We still receive occasional command timeouts after running Slash commands or handling message events. 

This PR leverages Starlette background task to *instantly* return a `200` `OK` to Slack upon receipt of a `/command` or `/event` interactivity payload. This change ensures that the first thing we do is acknowledge the request and aren't waiting on Bolt at all.

In my tests, I discovered that calling `ack()` at the beginning of our middleware function does not *actually* send a `200` back to Slack *at that point*. The way bolt works, it doesn't matter when you call `ack()`, as long you call it at the beginning of your listener function ([like this](https://github.com/Netflix/dispatch/pull/2907/files#diff-d2495d41fbf8f5c5500a410dc9fdcbcb271af34de0a447a84c0ec7b7a200b67bR444)). If you call it before then, like in a middleware function, it will function exactly the same as if you had called it at beginning of the listener function:

```bash
DEBUG:0.0009812499999988233:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/app/app.py:dispatch:525
DEBUG:function.elapsed.time.dispatch.plugins.dispatch_slack.middleware.resolve_context_from_conversation: 0.08142970899999824:/Users/wshel/Projects/dispatch/src/dispatch/decorators.py:wrapper:111
DEBUG:Got middleware response None:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/app/app.py:dispatch:530
DEBUG:Running listener: handle_list_participants_command ...:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/app/app.py:dispatch:549
DEBUG:Begin if not request.lazy_only conditional:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/listener/thread_runner.py:run:112
DEBUG:Enter run_ack_function_asynchronously:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/listener/thread_runner.py:run:153
DEBUG:Start handler:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/listener/thread_runner.py:run_ack_function_asynchronously:117
DEBUG:Waiting while response is None and it's been less than 3 seconds:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/listener/thread_runner.py:run:172
DEBUG:Begin run_ack_function:/Users/wshel/.pyenv/versions/3.9.7/envs/dispatch/lib/python3.9/site-packages/slack_bolt/listener/thread_runner.py:run_ack_function_asynchronously:122
DEBUG:begin handle_list_participants_command: 1.013716875 seconds have elapsed since recieving trigger_id: 4722662541473.4543933691524.8025f946588c09b10ef50b7d7ebe20f3:/Users/wshel/Projects/dispatch/src/dispatch/plugins/dispatch_slack/incident/interactive.py:handle_list_participants_command:473
```

This flow has been annotated below:

[bolt-python/blob/main/slack_bolt/listener/thread_runner.py#L59-L109](https://github.com/slackapi/bolt-python/blob/c99c23fd056d26f8b1e39436bd1fcd2c83a3e1bd/slack_bolt/listener/thread_runner.py#L59-L109)
```python
    def run(  # type: ignore
        self,
        request: BoltRequest,
        response: BoltResponse,
        listener_name: str,
        listener: Listener,
        starting_time: Optional[float] = None,
    ) -> Optional[BoltResponse]:
    
           ...snip
          
            if not request.lazy_only:
                self.logger.debug("Begin if not request.lazy_only conditional")
                # start the listener function asynchronously
                def run_ack_function_asynchronously():
                    nonlocal ack, request, response
                    try:
                        self.logger.debug("Start handler")
                        self.listener_start_handler.handle(
                            request=request,
                            response=response,
                        )
                        self.logger.debug("Begin run_ack_function")
                        # 1. (wshel) This is when we finally start running our command logic (this is async)
                        listener.run_ack_function(request=request, response=response)
                        self.logger.debug("Done with run_ack_function")
                    except Exception as e:
                        ...snip
                    finally:
                        self.listener_completion_handler.handle(
                            request=request,
                            response=response,
                        )

                self.logger.debug("Enter run_ack_function_asynchronously")
                self.listener_executor.submit(run_ack_function_asynchronously)
```

Note that the code above (`run` function) begins [here](https://github.com/slackapi/bolt-python/blob/c99c23fd056d26f8b1e39436bd1fcd2c83a3e1bd/slack_bolt/app/app.py#L523), within the [dispatch function](https://github.com/slackapi/bolt-python/blob/c99c23fd056d26f8b1e39436bd1fcd2c83a3e1bd/slack_bolt/app/app.py#L454), strictly after all middleware has been called. Middleware [inherently cannot return a response](https://github.com/slackapi/bolt-python/blob/main/slack_bolt/app/app.py#L479), if it does, the listener will never be executed.

### Tests

Slash commands sends custom ephemeral message without middleware:

Screenshot 2023-01-27 at 7 59 14 AM

Modal submission:

Screenshot 2023-01-27 at 7 59 35 AM

Message event:

Screenshot 2023-01-27 at 8 00 21 AM

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/endpoints.py
+++ b/src/dispatch/plugins/dispatch_slack/endpoints.py
@@ -85,7 +85,7 @@ async def slack_events(request: Request, organization: str, body: bytes = Depend
     task = BackgroundTask(handler.handle, req=request, body=body)
     return JSONResponse(
         background=task,
-        content={"ok": ""},
+        content=HTTPStatus.OK.phrase,
         status_code=HTTPStatus.OK,
     )
 
--- a/src/dispatch/plugins/dispatch_slack/endpoints.py
+++ b/src/dispatch/plugins/dispatch_slack/endpoints.py
@@ -1,7 +1,10 @@
 from http import HTTPStatus
+import json
+import time
 
 from fastapi import APIRouter, HTTPException, Depends
-from slack_bolt.adapter.starlette.handler import SlackRequestHandler
+from starlette.background import BackgroundTask
+from starlette.responses import JSONResponse
 from slack_sdk.signature import SignatureVerifier
 from sqlalchemy import true
 from starlette.requests import Request, Headers
@@ -10,10 +13,12 @@ from dispatch.database.core import refetch_db_session
 from dispatch.plugin.models import Plugin, PluginInstance
 
 from .bolt import app
+from .case.interactive import configure as case_configure
+from .handler import SlackRequestHandler
 from .incident.interactive import configure as incident_configure
 from .feedback.interactive import configure as feedback_configure
 from .workflow import configure as workflow_configure
-from .case.interactive import configure as case_configure
+from .messaging import get_incident_conversation_command_message
 
 
 router = APIRouter()
@@ -21,6 +26,15 @@ router = APIRouter()
 
 async def get_body(request: Request):
     return await request.body()
+
+
+async def parse_request(request: Request):
+    request_body_form = await request.form()
+    try:
+        request = json.loads(request_body_form.get("payload"))
+    except Exception:
+        raise HTTPException(status_code=400, detail=[{"msg": "Bad Request"}])
+    return request
 
 
 def is_current_configuration(
@@ -46,10 +60,11 @@ def get_request_handler(request: Request, body: bytes, organization: str) -> Sla
     )
     for p in plugin_instances:
         if is_current_configuration(body=body, headers=request.headers, plugin_instance=p):
-            incident_configure(p.configuration)
-            workflow_configure(p.configuration)
             case_configure(p.configuration)
             feedback_configure(p.configuration)
+            incident_configure(p.configuration)
+            workflow_configure(p.configuration)
+            app._configuration = p.configuration
             app._token = p.configuration.api_bot_token.get_secret_value()
             app._signing_secret = p.configuration.signing_secret.get_secret_value()
             session.close()
@@ -67,7 +82,12 @@ def get_request_handler(request: Request, body: bytes, organization: str) -> Sla
 async def slack_events(request: Request, organization: str, body: bytes = Depends(get_body)):
     """Handle all incoming Slack events."""
     handler = get_request_handler(request=request, body=body, organization=organization)
-    return await handler.handle(request)
+    task = BackgroundTask(handler.handle, req=request, body=body)
+    return JSONResponse(
+        background=task,
+        content={"ok": ""},
+        status_code=HTTPStatus.OK,
+    )
 
 
 @router.post(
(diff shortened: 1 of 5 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -337,9 +329,16 @@ def handle_update_incident_project_select_action(
...
-    body: dict, payload: dict, db_session: Session, context: BoltContext, client: WebClient
+    ack: Ack,
+    body: dict,
+    payload: dict,
+    db_session: Session,
+    context: BoltContext,
+    client: WebClient,
 ) -> None:
     """Handles the list incidents command."""
+    ack()
```
kevgliss, line 340: Are these acks necessary if we've already returned a 200 ok? Are they simply a way to avoid bolt warning us that we haven't called `ack`?

wssheldon: The latter, they're necessary to avoid bolt warning that we "haven't called ack()" and then exiting: https://github.com/slackapi/bolt-python/blob/c99c23fd056d26f8b1e39436bd1fcd2c83a3e1bd/slack_bolt/listener/thread_runner.py#L162-L168

We could still use the ack() middleware for this. I personally prefer this pattern but neither here nor there 🤷‍♀️

src/dispatch/plugins/dispatch_slack/endpoints.py
```diff
@@ -46,10 +60,11 @@ def get_request_handler(request: Request, body: bytes, organization: str) -> Sla
     )
     for p in plugin_instances:
         if is_current_configuration(body=body, headers=request.headers, plugin_instance=p):
-            incident_configure(p.configuration)
-            workflow_configure(p.configuration)
             case_configure(p.configuration)
             feedback_configure(p.configuration)
+            incident_configure(p.configuration)
+            workflow_configure(p.configuration)
+            app._configuration = p.configuration
```
wssheldon, line 67: Can we just put `app._configuration` in `context_properties` dictionary and remove `configuration_middleware()`?

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -435,12 +434,18 @@ def handle_list_incidents_command(
...
 def handle_list_participants_command(
+    ack: Ack,
     body: dict,
     client: WebClient,
     context: BoltContext,
     db_session: Session,
 ) -> None:
     """Handles list participants command."""
+    ack()
+    log.debug(
```
wssheldon, line 445: I can remove these after testing some (or add it to other commands if we want)

PR #2912, opened 2023-01-27
clean up db_middleware where we already have a session in context obj

wssheldon:


wssheldon:
@kevgliss -- I think this is OK, because the "scoped_db_session" is always "default". I am a little confused by the `resolve_context_from_conversation` function. 

```python
    db_session = SessionLocal()
    organization_slugs = [o.slug for o in organization_service.get_all(db_session=db_session)]
    db_session.close()
    organization_slugs = ["default"]  #  <------ 
    for slug in organization_slugs:
        scoped_db_session = refetch_db_session(slug)
        conversation = conversation_service.get_by_channel_id_ignoring_channel_type(
            db_session=scoped_db_session, channel_id=channel_id
        )
        if conversation:
            subject = SubjectMetadata(
                type="incident",
                id=conversation.incident_id,
                organization_slug=slug,
                project_id=conversation.incident.project_id,
            )
            return Subject(subject, db_session=scoped_db_session)

        scoped_db_session.close()
```

Should we be iterating through all of them and if so, would this PR still work?

kevgliss:
>     organization_slugs = ["default"]  #  <------ 
This should probably be fixed to use the actual list of slugs and is likely left over from me trying to optimize calls.

I don't have a strong opinion on using `context["db_session"]` vs `db_session: Session` the way that bolt unpacks it's args; these are the same variable, and you can inject anything in the function that's in the context as a function arg. I think `db_session: Session` is a bit cleaner as we can add some typing to it but as long as we're consistent, I'm fine with either approach.

wssheldon:
> This should probably be fixed to use the actual list of slugs and is likely left over from me trying to optimize calls.

I went ahead and removed the 'default' line here and reverted back to `db_session: Session` in the `*args`, which I also prefer (didn't know that could work). Main goal with this diff is to avoid unnecessary use of the `db_middleware` when we already have a `db_session` from `resolve_context_from_conversation`, which is used within the `message_context_middleware` and a couple other places. Trying to simplify the middleware/make the request flow faster.

Haven't had a chance to test this most recent commit https://github.com/Netflix/dispatch/pull/2912/commits/623a5470237780ca36d3d095b7051cbb161ba6c5 due to some network issues.

wssheldon:
Tested this and didn't run into any issues.

PR #2919, opened 2023-01-30
Adds dialog to handoff incidents in bulk

mvilanova:
Screenshot 2023-02-10 at 12 00 04 PM

Screenshot 2023-02-13 at 12 19 47 PM


kevgliss:
Is it common to hand off incidents across different projects? I would have thought that a "bulk" re-assign feature (similar to what we do with `marked closed`) would be the more straightforward way to select incidents and accomplish the hand-off

PR #2921, opened 2023-01-31
testing/case-tests

wssheldon:
Work in-progress to add tests for the `case` service. Pushing early as I did not see a similar suite for the `incident` service and want to avoid going down the wrong path for too long.


kevgliss:
lgtm

PR #2927, opened 2023-02-01
Fixes issue when selects have null values

kevgliss:


wssheldon:
Resolves: `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'id')`

wssheldon (approved):

PR #2936, opened 2023-02-02
tests/end-to-end-test-for-report-submission-using-playwright-in-ci

wssheldon:
Tests work great locally, need to brainstorm how to fully automate authentication in CI, and how to run the development server in CI on GH Actions.

I recreated the bug where the "Load More" button did not appear for the Project select button, introduced [here](https://github.com/Netflix/dispatch/pull/2770). The test will check for the "Load More" text and fail if it can't be found after selecting the `Type` button, this failed as expected:

```typescript
    // check that the 'Load More' selector is available upon opening the Type dropdown
    await expect(
      page.getByText("Load More"),
      "The 'Load More' selector should be visibile."
    ).toBeVisible()
```

```bash
Error: The 'Load More' selector should be visibile.

Call log:
  - expect.toBeVisible with timeout 10000ms
  - waiting for getByText('Load More')
  -   unexpected value "hidden"

  35 |       page.getByText("Load More"),
  36 |       "The 'Load More' selector should be visibile."
> 37 |     ).toBeVisible()
     |       ^
  38 |
  39 |     await page.getByText(type, { exact: true }).click()

    at /Users/wshel/Projects/dispatch/tests/e2e/report-submission.spec.ts:37:7
```


wssheldon:
> I think you should be able to run these via a github action:

I think I'll need to use the `.devcontainer` to be able to run `dispatch server develop`

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    await page.getByRole("combobox").filter({ hasText: "Project" }).locator("i").click()
+    await page.getByText(project, { exact: true }).click()
+
+    // Type
+    await page.getByRole("button", { name: "Type" }).click()
+    await page.getByText(type, { exact: true }).click()
+
+    // Priority
+    await page.getByRole("button", { name: "Priority" }).click()
+    await page.getByText(priority).click()
```
mvilanova, line 36: Shouldn't these be of type combobox or are you trying to mock clicking on it and loading a value?

wssheldon: This clicks the button and types a value.

wssheldon: It could be done either way though like you said.

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // Open the dropdown selection for the Tags field
+    await page.getByRole("combobox").filter({ hasText: "Tags" }).locator("i").click()
+
+    // Soft check that the 'Load More' selector is available upon opening the Tags dropdown
+    await expect
+      .soft(page.getByText("Load More"), "The 'Load More' selector should be visibile.")
```
wssheldon, line 43:
```suggestion
      .soft(page.getByText("Load More"), "The 'Load More' selector should be visible when there are more than 5 options in the Tags combobox.")
```

Also, can we check if there's more than 5 options in a smart way first to make this a conditional check?

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    // Wait for the incident to be created
+    await page.waitForLoadState("networkidle")
+
+    // Soft validate that we get redirected to the incident submission form
+    await expect
+      .soft(page)
+      .toHaveURL(
+        encodeURI(
+          `./default/incidents/report?project=${project}&incident_priority=${priority}&incident_type=${type}`
+        )
```
wssheldon, line 61: The [custom exception messages](https://github.com/Netflix/dispatch/pull/2936/files#diff-5221afac34a469ab3bdfa7bd61adfa4d2093bb89bd78506d6dfd48434a917280R43) are pretty nice. We should consider adding one here.

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // Soft validate that we recieve the report form.
+    await expect.soft(page.getByText("Incident Report")).toBeVisible()
+
+    // Soft validate that we recieve the report form.
+    await expect
+      .soft(
+        page.getByText(
+          "This page will be populated with incident resources as they are created (if available). If you have any questions, please feel free to review the Frequently Asked Questions (FAQ) document linked below, and/or reach out to the listed Incident Commander."
+        )
```
wssheldon, line 72: The [custom exception messages](https://github.com/Netflix/dispatch/pull/2936/files#diff-5221afac34a469ab3bdfa7bd61adfa4d2093bb89bd78506d6dfd48434a917280R43) are pretty nice. We should consider adding one here.

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  })
+  test("Should allow me to report an incident", async ({ page }) => {
+    /* The ability to report an incident is one of the most critical
+      user stories in the Dispatch application. */
+
+    const project = "Test"
+    const type = "Customer Data"
+    const priority = "Low"
+
+    await page.getByRole("link", { name: "Report Incident" }).click()
```
wssheldon, line 16: I wonder if there might be a reason to visit the URL directly as opposed to clicking the `Report Incident` button. I suspect the direct URL visit may be more reliable and this test might be better isolated if we avoid any side effects of this button click.

tests/e2e/report-submission.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // Soft validate that we recieve the report form.
+    await expect
+      .soft(
+        page.getByText(
+          "This page will be populated with incident resources as they are created (if available). If you have any questions, please feel free to review the Frequently Asked Questions (FAQ) document linked below, and/or reach out to the listed Incident Commander."
+        )
+      )
+      .toBeVisible()
+  })
```
wssheldon, line 75: We might want to extend this test to validate the submission report card has not regressed in anyway. This pattern will probably come up a lot and it feels very verbose and maybe flakey to do it this way. Maybe a todo could be to investigate using something like this: https://www.thoughtworks.com/en-us/radar/techniques/component-visual-regression-testing

wssheldon: https://storybook.js.org/tutorials/intro-to-storybook/vue/en/get-started/

src/dispatch/static/dispatch/package-lock.json
```diff
@@ -7047,8 +7085,7 @@
       "version": "18.11.7",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-18.11.7.tgz",
       "integrity": "sha512-LhFTglglr63mNXUSRYD8A+ZAIu5sFqNJ4Y2fPuY7UlrySJH87rRRlhtVmMHplmfk5WkoJGmDjE9oiTfyX94CpQ==",
-      "dev": true,
-      "peer": true
```
wssheldon, line 7051: Not sure why that changed

PR #2937, opened 2023-02-02
Bookmark Incident Review document in the conversation after its creation

mvilanova:


src/dispatch/incident/flows.py
```diff
@@ -1125,6 +1125,29 @@ def incident_stable_status_flow(incident: Incident, db_session=None):
...
+                title="Incident Review Document",
+            )
+        except Exception as e:
+            event_service.log_incident_event(
+                db_session=db_session,
+                source="Dispatch Core App",
+                description=f"Bookmarking the incident review document failed. Reason: {e}",
+                incident_id=incident.id,
+            )
+            log.exception(e)
```
mvilanova, line 1149: @wssheldon @kevgliss I think there's an opportunity to make this its own function that we can call as many times as resources need to be bookmarked in `set_conversation_bookmarks` for example and it can be used in other places. We could tackle it in a separate PR.

PR #2945, opened 2023-02-06
Refact E2E tests to use Page Object Model (POM), fixtures, and add additional tests

wssheldon:


src/dispatch/static/dispatch/src/incident/Table.vue
```diff
@@ -39,6 +39,7 @@
             :sort-by.sync="sortBy"
             :sort-desc.sync="descending"
             :loading="loading"
+            data-testid="incident-data-table"
```
wssheldon, line 42: We may want to consider a single SOT for these `data-testid` strings as an `Enum` that we can use here and in tests `Pages`

tests/static/e2e/fixtures/dispatch-fixtures.ts
```diff
@@ -0,0 +1,27 @@
+// my-test.ts
```
wssheldon, line 1:
```suggestion
```

tests/static/e2e/routes.ts
```diff
@@ -0,0 +1,9 @@
+export const orgSlug = "default"
+
+export enum Routes {
```
wssheldon, line 3: Would be nice if we could import these directly from the Vue router. I was unable to import the routes because of an issue with this line and others upon import (can't use this outside module, or similar error):

```
const authProviderSlug =
  import.meta.env.VITE_DISPATCH_AUTHENTICATION_PROVIDER_SLUG || "dispatch-auth-provider-basic"
```

tests/static/e2e/pages/incidents-page.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly OtherRow: Locator
+  readonly NextPage: Locator
+  readonly EditKebab: Locator
+  readonly EditMenu: Locator
+  readonly EditViewEdit: Locator
+  readonly EditCreateReport: Locator
+  readonly EditRunWorkflow: Locator
+  readonly EditDelete: Locator
+
+  constructor(page: Page, incident: string = `dispatch-${orgSlug}-${orgSlug}-2`) {
```
wssheldon, line 18: The default incident here is a bit fragile to it being present in the table, there's probably a better way.

PR #2949, opened 2023-02-07
Allow cases to be re-assigned after addition of participant model

wssheldon:
Still testing this, wanted to push changes in case @kevgliss is working on it.

PR #2952, opened 2023-02-07
Direct add participants and allow for new participants via messages.

kevgliss:


src/dispatch/case/flows.py
```diff
@@ -282,6 +282,16 @@ def case_new_create_flow(*, case_id: int, organization_slug: OrganizationSlug, d
                 )
                 # wait until all resources are created before adding suggested participants
                 individual_participants = [x.email for x, _ in individual_participants]
+
+                for email in individual_participants:
+                    # we don't rely on on this flow to add folks to the conversation because in this case
+                    # we want to do it in bulk
+                    case_add_or_reactive_participant_flow(
```
wssheldon, line 289:
```suggestion
                    case_add_or_reactivate_participant_flow(
```

I updated this in https://github.com/Netflix/dispatch/pull/2949/files

PR #2953, opened 2023-02-08
Propose the adoption of lightweight architecture decision records

wssheldon:
Storing these in `git` is useful in my experience -- they work best when kept near the code they pertain to because developers will find them in the course of working, and they can be created as part of the PR introducing a change—all ensuring the pertinent context is available with minimal searching.  You can see an example of this in use by the [UK government](https://github.com/alphagov/govuk-aws/tree/master/doc/architecture/decisions).

### DIFF
--- a/docs/architecture/decisions/0001-record-architecture-decisions.md
+++ b/docs/architecture/decisions/0001-record-architecture-decisions.md
@@ -1,0 +1,20 @@
+# 1. Record architecture decisions
+
+Date: 2023-02-07
+
+## Status
+
+Proposed
+
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will use Architecture Decision Records, as described by Michael Nygard in
+[documenting architecture decisions](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)
+
+## Consequences
+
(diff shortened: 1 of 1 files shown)


wssheldon:
cc @kevgliss 

(Can't find you in reviewer dropdown for some reason)

kevgliss:
I'm not against this, but from your linked example all these decisions seem rather deployment oriented. e.g. remove load balancers, and migrate to redis which I'm not sure makes sense to document publically as it's not likely to be useful to other users of dispatch. 

Do you have some other examples of the types of decisions that might fit into these records?

mvilanova:
+1 on what @kevgliss said. These seem to be more appropriate for our internal documentation.

wssheldon:
> Do you have some other examples of the types of decisions that might fit into these records?

Some examples would be "use playwright for end-to-end UI tests", "migrate to Slack Bolt SDK for the conversation plugin", "use Fast API for the REST API / web framework", etc.

Either way, I'm good with storing them internally, if really whatever y'all prefer. It's a 2 way door decision, so we can always revisit if we want. Here's another UK GOV example that's less deployment oriented: https://github.com/alphagov/govuk-aws/blob/main/docs/architecture/decisions/0017-terraform-data-structure.md -- I could see a similar thing being helpful for the dispatch service/views/flows/etc architecture model.

PR #2960, opened 2023-02-08
Only display active roles in CaseParticipantsTab component, add Vitest, and unit test(s)

wssheldon:
This PR prevents renounced roles from being displayed in the CaseParticipantsTab component. Previously, when someone has frequently changed roles, their roles will appear as `('Reporter', 'Assignee', 'Participant', 'Assignee'), for example. We only display roles that haven't been renounced and add support for testing Vue 2 components with Vitest [0]. Finally, we add a unit test for this specific case and others for the CaseParticipantsTab component.

```bash
 ✓ tests/participants-tab.spec.js (2)

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Start at  08:36:41
   Duration  778ms (transform 211ms, setup 0ms, collect 191ms, tests 14ms)


 PASS  Waiting for file changes...
       press h to show help, press q to quit
```

[0] https://vitest.dev/


.github/workflows/javascript.yml
```diff
@@ -21,3 +21,7 @@ jobs:
         working-directory: src/dispatch/static/dispatch
         run: |
           npm run lint
+      - name: Run Vitest
+        working-directory: src/dispatch/static/dispatch
+        run: |
+          npm run test:coverage
```
wssheldon, line 27:
```suggestion
          npm run test
```

Try running w/o coverage to resolve error.

.github/workflows/javascript.yml
```diff
@@ -21,3 +21,7 @@ jobs:
         working-directory: src/dispatch/static/dispatch
         run: |
           npm run lint
+      - name: Run Vitest
+        working-directory: src/dispatch/static/dispatch
+        run: |
+          npm run test
```
wssheldon, line 27:
```suggestion
          npm run test:coverage
```

Run coverage again.

wssheldon:
@kevgliss @mvilanova Any opinions on where the tests for front-end should be stored? Here they are stored in the static/ dir and E2E tests are currently in the gitroot tests/ dir. Which do you prefer?

src/dispatch/static/dispatch/tests/participants-tab.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      },
+    }
+
+    store = new Vuex.Store({ state })
+  })
+
+  it("renders the correct number of participants", () => {
+    const wrapper = mount(CaseParticipantsTab, { store, localVue })
+    const participants = wrapper.vm.$store.state.case_management.selected.participants
+    expect(participants).to.have.lengthOf(2)
```
kevgliss, line 61: Doesn't this just check that we have an array of two in the participant store (which you set directly above)? I don't see how this actually checks to see if the dom actually rendered two participants.

kevgliss: > @kevgliss @mvilanova Any opinions on where the tests for front-end should be stored? Here they are stored in the static/ dir and E2E tests are currently in the gitroot tests/ dir. Which do you prefer?

I'm fine with those locations, they make sense if asking a question "what do these tests cover"? e.g. integration is a top level but various unit or lower level tests are located closer to the code under test.

src/dispatch/static/dispatch/src/case/ParticipantsTab.vue
```diff
@@ -39,13 +39,12 @@ export default {
...
-          return v.role
-        })
-        .join(", ")
-    },
+    activeRoles: function (participantRoles) {
+        return participantRoles
+          .filter(role => !role.renounced_at)
+          .map(role => role.role)
+          .join(", ");
+    }
```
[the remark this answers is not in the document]
wssheldon: Will do

src/dispatch/static/dispatch/tests/participants-tab.spec.js
wssheldon: Yeah I think the test is wrong. I'll dig into it more.

wssheldon: Should be fixed now.

src/dispatch/static/dispatch/src/case/ParticipantsTab.vue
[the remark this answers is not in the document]
wssheldon: Btw, I found out that after moving the filter to a global one (so we can use it in incidents), that you need to import it into the test like so:

https://github.com/Netflix/dispatch/pull/2960/files#diff-f49aa04a05e87b354472293a74ab6cea2fea115c2502b6fe1244fdd37e933ed8R11-R12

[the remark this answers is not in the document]
wssheldon: That was why the test didn't work.

PR #2966, opened 2023-02-09
Feature/markdown in incident desc

vindex10:
Hi!

In this pull request I introduce a special flag that enables markdown for incident descriptions:

`DISPATCH_MARKDOWN_IN_INCIDENT_DESC`

and 

`DISPATCH_ESCAPE_HTML`

that generally allows HTML in the text fields. The latter, if not specified explicitly is switched off by markdown.

I assume some plugins may break when descriptions are not plain text. But those who use these plugins now, stick to plain text in the incident descriptions as well so won't enable this options anyways.


vindex10:
I fixed linting, and updated requirements.txt. I hope end to end tests will pass through now. If not, I will try running some tests with local deployment. (I only tested the email formatting part, so missed the requirements initially)

vindex10:
I managed to send email with markdown in the version: [v20221207](https://github.com/Netflix/dispatch/releases/tag/v20221207)

vindex10:
@mvilanova could you please consult me on what would be the right way to add an extra dependency to the project?

The difficulty is that there are two files: requirements-base.in and requirements-dev.in and then with `pip-compile` both produce  ".txt" snapshots of versions with "==" specification for each package version.

Now, when we are installing the test dependencies `pip install -e ".[dev]"` some packages that are not direct dependencies (not mentioned in ".in" files), but present in both "-base.txt" and "-dev.txt" may have different versions, because of different constraints implied from the base dependencies.

Therefore, pip cannot resolve dependencies like:

```
    dispatch[dev] 0.1.0.dev0 depends on attrs==22.1.0
    dispatch[dev] 0.1.0.dev0 depends on attrs==22.2.0; extra == "dev"
```

And I don't really know how to properly solve this properly. What I did in this commit: [0e3b894](https://github.com/Netflix/dispatch/pull/2966/commits/0e3b894169a31d59922fb8d2ab7f0361269e28f6), I edited the generated requirements-dev.txt by hand, and replaced `attrs==22.2.0` with `22.1.0`.. I hope this is quite fine.

vindex10:
@mvilanova could you please run tests? :)

mvilanova:
> @mvilanova could you please consult me on what would be the right way to add an extra dependency to the project?
> 
> The difficulty is that there are two files: requirements-base.in and requirements-dev.in and then with `pip-compile` both produce ".txt" snapshots of versions with "==" specification for each package version.
> 
> Now, when we are installing the test dependencies `pip install -e ".[dev]"` some packages that are not direct dependencies (not mentioned in ".in" files), but present in both "-base.txt" and "-dev.txt" may have different versions, because of different constraints implied from the base dependencies.
> 
> Therefore, pip cannot resolve dependencies like:
> 
> ```
>     dispatch[dev] 0.1.0.dev0 depends on attrs==22.1.0
>     dispatch[dev] 0.1.0.dev0 depends on attrs==22.2.0; extra == "dev"
> ```
> 
> And I don't really know how to solve this properly. What I did in this commit: [0e3b894](https://github.com/Netflix/dispatch/pull/2966/commits/0e3b894169a31d59922fb8d2ab7f0361269e28f6), I edited the generated requirements-dev.txt by hand, and replaced `attrs==22.2.0` with `22.1.0`.. I hope this is quite fine.

You shouldn't need to run `pip-compile` against the `requirements-dev.in` file. Just add the dependencies in the `requirements-base.in` file and run `pip-compile` against it.

mvilanova:
I'll try to find time to test this locally and circle back.

vindex10:
@mvilanova  All checks have passed, but branch got out of date. When I "synced", tests got discarded :(  could you please re-run when you have time :) thank you!

vindex10:
apparently "automatic sync" broke requirements, some tests failed because of this. I rebuilt requirements and resolved conflicts manually

mvilanova:
Thanks for the PR!

PR #2977, opened 2023-02-10
add base concept of entities to signals

wssheldon:
Right now, the only type of Entity implemented will recursively search a signal for a entity that matches a regular expression (as opposed to specify, say the field/key name, where that entity should be). I will end up implementing both as an option. There is also the concept of a "global" entity, where we will search all signals, recursively, for that entity. 

We should prioritize "time-to-first-render" and not let entity correlation be a bottleneck, although I expect this will still be fairly performant, all things considered. Only after a signal/case has been fully processed, will we correlate entities with other cases and signals. We can use UI tricks to try to make this invisible to the user and our UI is already set up well for this, by putting information with tabs on the CaseEditSheet.

### Case View Mockup

Screenshot 2023-02-10 at 1 13 12 PM

### Implemented Entity Case View

case_entitiy_tab

You can click the red context cards and view the signals. Cases table dialog not implemented yet.


src/dispatch/entity/service.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def delete(*, db_session, entity_id: int):
+    """Deletes an existing entity."""
+    entity = db_session.query(Entity).filter(Entity.id == entity_id).one_or_none()
+    db_session.delete(entity)
+    db_session.commit()
+
+
+def get_cases_with_entity(db: Session, entity_id: int, days_back: int) -> int:
```
wssheldon, line 130:
```suggestion
def get_cases_with_entity(db: Session, entity_id: int, days_back: int) -> list[Case]:
```

src/dispatch/signal/service.py
```diff
@@ -326,3 +341,110 @@ def supress(
...
+        # If the value is a string, search it for entity matches
+        elif isinstance(val, str):
+            for entity_type, entity_regex, field in entity_type_pairs:
+                # If a field was specified for this entity type, only search that field
+                if not field or key == field:
+                    # Search the string for matches to the entity type's regular expression
+                    if match := entity_regex.search(val):
+                        # If a match was found, create a new Entity object for it
+                        entity = EntityCreate(
+                            value=match.group(0),
```
wssheldon, line 414: We might want to add the fact that we don't really support capture groups to the regex tooltip until we figure out how we can let users add them.

src/dispatch/signal/flows.py
```diff
@@ -50,7 +58,7 @@ def create_signal_instance(
     if duplicate:
         return
 
-    # create a case if not duplicate or supressed
+    # # create a case if not duplicate or supressed
```
wssheldon, line 61:
```suggestion
    # create a case if not duplicate or suppressed
```

src/dispatch/entity/views.py
```diff
@@ -0,0 +1,79 @@
...
+
+@router.get("/{entity_id}/cases", response_model=None)
+def count_cases_with_entity(*, db_session: Session = Depends(get_db), entity_id: PrimaryKey):
+    cases = get_cases_with_entity(db=db_session, entity_id=entity_id, days_back=30)
+    return {"cases": cases}
+
+
+@router.get("/{entity_id}/signal_instances", response_model=None)
+def get_signal_instances_by_entity(entity_id: int, db: Session = Depends(get_db)):
+    instances = get_signal_instances_with_entity(db, entity_id, days_back=30)
```
wssheldon, line 78: Do we want to ship this with the 30 day default for now and then add a datetime select after mvp?

wssheldon:
Global entity types are not implemented. I will remove from front-end, but keep it in the model.

src/dispatch/static/dispatch/src/entity_type/api.js
```diff
@@ -0,0 +1,29 @@
+import API from "@/api"
+
+const resource = "/entity_type"
+
+export default {
+  getAll(options) {
+    return API.get(`${resource}`, { params: { ...options } })
+  },
+
+  get(EntityTypeId) {
```
kevgliss, line 10: nit: lowerCamelCase

kevgliss (approved):
lgtm, some small suggestions

src/dispatch/entity/service.py
```diff
@@ -0,0 +1,269 @@
...
+            for item in val:
+                entities.extend(search(None, item, entity_type_pairs))
+
+        # If the value is a string, search it for entity matches
+        elif isinstance(val, str):
+            for entity_type, entity_regex, field in entity_type_pairs:
+                # If a field was specified for this entity type, only search that field
+                if not field or key == field:
+                    # Search the string for matches to the entity type's regular expression
+                    if match := entity_regex.search(val):
```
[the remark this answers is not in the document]
wssheldon: Right now setting a regex is required. That's a good point though if we just want the whole value and know it's always going to be in a certain field in a certain signal. I guess right now you'd just set the field and the regex to `.*` or a pattern for the value. We should probably support no regex though.

src/dispatch/entity/views.py
wssheldon: the `days_back` was there to add support for a date range picker on the entity tab (not implemented yet). we probably do care about too much data though as well.

src/dispatch/static/dispatch/src/entity/EntityCard.vue
```diff
@@ -0,0 +1,119 @@
...
+        const [casesResponse, signalResponse] = await Promise.all([
+            casePromise,
+            signalPromise,
+        ]);
+        this.caseCount = casesResponse.cases.length;
+        this.signalInstanceCount = signalResponse.instances.length;
+        this.isLoading = false;
+    },
+    methods: {
+      async getSignalInstances() {
```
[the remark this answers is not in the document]
wssheldon: I wanted the code above to be async and didn't have another non-async version of the getSignal... api call. We could add a non-async version probably.

wssheldon:
> lgtm, some small suggestions

thanks for looking at it

PR #3003, opened 2023-02-16
Create an entity type creation dialog flow that embeds a regex and jsonpath playground

wssheldon:
### Mockup

Divided areas are expandable with arrow buttons. Regex and JSONPath text area's auto-grow. Option to load signal data from prior signals or just put any arbitrary json blob. Highlight matching for values that match the supplied entity type pair (jsonpath +/- regex). 

Retrieve list of all signal instances over last 30 days and run javascript version of `find_entities` function that checks the entity type pair supplied in realtime as the user updates the `regex` or `jsonpath` text area. Matches are returned and update the playground JSON object with any matches via CSS highlighting.

Screenshot 2023-02-16 at 9 23 28 AM

#### Create Dialog

Screenshot 2023-02-16 at 9 48 45 AM

### MVP

The MVP here is everything minus the right hand side JSON playground. We should still check for matches with the supplied values and display a count + table of matches.


wssheldon:
Abandoning for new PR

PR #3028, opened 2023-02-28
Allows tags to have external IDs

kevgliss:


wssheldon (approved):
Is this related to MITRE tags?

kevgliss:
> Is this related to MITRE tags?

Yup, external id allows us to map that mitre tactic tag to its mitre id.

PR #3035, opened 2023-03-01
Creates a new plugin for Duo that provides on-demand MFA across Dispatch

wssheldon:
This PR introduces a new plugin for Duo that provides a push notification method that can be used across the application to gate certain functionality behind an additional MFA prompt. We also include an example Slack flow to visualize how this would work within the Slack plugin. 

The example is complete and shows how we can change behavior based on the result of the Duo response. I'll remove the example if we are OK with the structure of the plugin overall.


src/dispatch/plugins/dispatch_duo/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+                {"response": {"result": "allow", "status": "allow", "status_msg": "Success. Logging you in..."}, "stat": "OK"}
+
+        Example:
+            >>> plugin = DuoMfaPlugin()
+            >>> result = plugin.send_push_notification(username='wshel@netflix.com', type='Login Request')
+            >>> result
+            {'response': {'result': 'allow', 'status': 'allow', 'status_msg': 'Success. Logging you in...'}, 'stat': 'OK'}
+
+        Notes:
+            For more information, see https://duo.com/docs/authapi#/auth.
```
wssheldon, line 75:
```suggestion
            For more information, see https://duo.com/docs/authapi#/auth
```

wssheldon:
Chatted w/ @kevgliss about this PR, updated plugin name to `auth-mfa` and will merge this.

PR #3073, opened 2023-03-07
Add snooze button to signal instance message in Slack plugin and add Case/Signal subject enum

wssheldon:
Screenshot 2023-03-07 at 12 04 40 PM


src/dispatch/plugins/dispatch_slack/models.py
```diff
@@ -29,3 +31,12 @@ class SubjectMetadata(BaseModel):
 
     project_id: Optional[str]
     channel_id: Optional[str]
+
+
+class CaseSubjects(DispatchEnum):
```
wssheldon, line 36: @mvilanova @kevgliss prefer this pattern over the current raw str? I can add to everything.

kevgliss: LGTM

kevgliss:
For the layout, I wonder if we should have an `actions` pull down? e.g. it would have "view raw" and "snooze" as options. I would like to be cautious of the amount of vertical space we take up.

kevgliss (approved):

PR #3074, opened 2023-03-07
check for entity types when we create a signal

wssheldon:


src/dispatch/static/dispatch/src/signal/store.js
```diff
@@ -141,6 +141,7 @@ const actions = {
     commit("RESET_SELECTED")
   },
   save({ commit, dispatch }) {
+    console.log("Saving state: %O", state)
```
wssheldon, line 144:
```suggestion
```

PR #3077, opened 2023-03-08
Adds testing to signal instance creation

kevgliss:


src/dispatch/signal/flows.py
```diff
@@ -1,20 +1,26 @@
...
+    current_user: DispatchUser = None,
+):
     """Creates a signal and a case if necessary."""
     signal = signal_service.get_by_variant_or_external_id(
         db_session=db_session,
         project_id=project.id,
-        external_id=signal_instance_data["id"],
-        variant=signal_instance_data.variant,
+        external_id=signal_instance_data.get("id"),
+        variant=signal_instance_data["variant"],
```
wssheldon, line 23: should it be `.get("variant")` since it's optional field?

kevgliss: I think so, technically it's one or the other, I think we may need to eventually re-add something like raw signal back and enforce either 'id' or 'variant' (otherwise we can't do anything with it). But I need to figure out how to add arbitrary keys in a pydantic model (if that's even possible). 

I'm going to merge this for the time being until I can figure that out.

PR #3089, opened 2023-03-09
update action context middleware to account for other data in private_metadata

wssheldon:


kevgliss:
I think this looks okay. I'm still not entirely clear on how the subject is being nested. I would have preferred the subject is flattened before it's embedded in the button itself. For instance, it feels like we are convoluting something because we do even more unpacking of the metadata and subject [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/case/interactive.py#L550), where I thought the middleware would largely abstract that for you. 

Is it because you need the`project_id` here?
https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/case/interactive.py#L591
In that case you already have the `signal` object so you'd be able to fetch it via `signal.project.id`.

wssheldon:
> I think this looks okay. I'm still not entirely clear on how the subject is being nested. I would have preferred the subject is flattened before it's embedded in the button itself. For instance, it feels like we are convoluting something because we do even more unpacking of the metadata and subject [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/case/interactive.py#L550), where I thought the middleware would largely abstract that for you.
> 
> Is it because you need the`project_id` here? https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/case/interactive.py#L591 In that case you already have the `signal` object so you'd be able to fetch it via `signal.project.id`.

This probably makes it even more confusing. But, the middleware change is actually to satisfy what `user_middleware` expects here:

```python
        participant = participant_service.get_by_case_id_and_conversation_id(
            db_session=db_session, case_id=context["subject"].id, user_conversation_id=user_id
        )
```

We use `user_middleware` to populate `creator` here:

```python
            new_filter = signal_service.create_signal_filter(
                db_session=db_session, creator=user, signal_filter_in=filter_in
            )
```

The `subject` within `context` isn't actually used in `handle_snooze_submission_event`. We could, but it's in the `private_metadata` as well:

```python
def handle_snooze_submission_event(
    ack: Ack,
    body,
    client: WebClient,
    context: BoltContext,
    db_session: Session,
    user: DispatchUser,
) -> None:
    """Handle the submission event of the snooze modal.

    This function is executed when a user submits the snooze modal. It first
    sends an MFA push notification to the user to confirm the action. If the
    user accepts the MFA prompt, the function retrieves the relevant information
    from the form data and creates a new signal filter. The new filter is then
    added to the existing filters for the signal. Finally, the function updates
    the modal view to show the result of the operation.

    Args:
        ack (Ack): The acknowledgement function.
        body (dict): The request body.
        client (WebClient): The Slack API client.
        context (BoltContext): The context data.
        db_session (Session): The database session.
        user (DispatchUser): The Dispatch user who submitted the form.
    """
    metadata = json.loads(body["view"]["private_metadata"])
    subject = SubjectMetadata.parse_raw(metadata["subject"])
```

So the issue stems from the middleware (`action_context_middleware`) expecting subject to be the only value in `private_metadata`, but we also need to use `private_metadata` as plumbing for the `form_data` in this situation, and then `user_middleware` need the subject context (and in a specific way).

It's all fairly confusing to me, hopefully this helps come up with a way we can simplify here, maybe in a separate PR.

kevgliss:
Thanks, that helps makes things clearer for me. My preference is that we are _explicit_ in supporting the form data use case for private metadata. The only use case originally for private metadata was to get a messages/button's context via SubjectMeta, it seems we now have an additional use case for storing form-data state (in this case between modals). For this, we should introduce another model may be called FormData, and make sure that the middleware can handle it (or ignore it) as appropriate.

wssheldon:
That helps a lot, I forgot we have a similar pattern for `TaskMetadata`. Current iteration should cover what you suggested.

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -662,9 +658,9 @@ def _create_snooze_filter(
             # Get the existing filters for the signal
             _create_snooze_filter(
                 db_session=db_session,
-                form_data=metadata["form_data"],
+                form_data=context["subject"].form_data,
                 user=user,
-                subject=subject,
+                subject=context["subject"],
```
kevgliss, line 663: nit: could the `form_data` arg be removed in as the `subjet` arg would also have the form_data values/

wssheldon: yeah, good catch. i'll update in a new pr.

PR #3090, opened 2023-03-09
Adding case resolution reason

kevgliss:


tests/factories.py
```diff
@@ -727,6 +727,8 @@ class CaseFactory(BaseFactory):
     name = FuzzyText()
     title = FuzzyText()
     description = FuzzyText()
+    resolution = FuzzyText()
+    resolution_reason = FuzzyChoice(["False Positivie", "User Acknowleged"])
```
wssheldon, line 731:
```suggestion
    resolution_reason = FuzzyChoice(["False Positive", "User Acknowledged"])
```

src/dispatch/plugins/dispatch_slack/fields.py
```diff
@@ -360,6 +362,28 @@ def resolution_input(
...
 
+def case_resolution_reason_select(
+    db_session: SessionLocal,
+    action_id: str = DefaultActionIds.case_resolution_reason_select,
+    block_id: str = DefaultBlockIds.case_resolution_reason_select,
+    label: str = "",
+    initial_option: dict = None,
+    **kwargs,
+):
+    """Creates an incident priority select."""
```
mvilanova, line 373: This docstring needs to be updated.

src/dispatch/static/dispatch/src/case/DetailsTab.vue
```diff
@@ -29,6 +29,14 @@
           />
         </ValidationProvider>
       </v-flex>
+      <v-flex xs6>
+        <v-select
+          v-model="resolution_reason"
+          label="Resolution Reason"
+          :items="visibilities"
```
mvilanova, line 36: `resolutionReasons` instead?

mvilanova (changes requested):

src/dispatch/static/dispatch/src/case/DetailsTab.vue
kevgliss: That doesn't map to what the api returns.

mvilanova: I don't understand. You're providing visibilities not resolution reasons defined in https://github.com/Netflix/dispatch/pull/3090/files#diff-15e369cbda358d580577a6be233ec92377ef358ce0797560ff0f221e263fd47aR161 as an options to select from?

src/dispatch/plugins/dispatch_slack/fields.py
mvilanova: Forgot to update it?

src/dispatch/static/dispatch/src/case/DetailsTab.vue
mvilanova: Screenshot 2023-03-10 at 1 47 00 PM

PR #3093, opened 2023-03-10
Enable workflow notification via ENV

Cinojose:
Right now workflow notification seems to be hardcoded to False. This PR is make it configurable via env.

### DIFF
--- a/src/dispatch/static/dispatch/src/signal/RawSignalViewer.vue
+++ b/src/dispatch/static/dispatch/src/signal/RawSignalViewer.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-dialog v-model="dialog" persistent max-width="800px">
+  <v-dialog v-model="dialog" persistent max-width="1000px">
     <template v-slot:activator="{ on }">
       <v-btn class="mr-2" icon v-on="on"> <v-icon>mdi-card-search-outline</v-icon> </v-btn>
     </template>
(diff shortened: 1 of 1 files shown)


kevgliss:
Hi @Cinojose, this is a good catch and is a bug. But I think the fix might be a bit more nuanced. I think what we really want here is to only send workflow notifications to non-closed incidents. This would be the same pattern as task syncing here:
https://github.com/Netflix/dispatch/blob/master/src/dispatch/task/scheduled.py#L85 

We could solve this a couple different ways. One would be to full adopt the task pattern for workflows, e.g.:

1) Sync workflows for active stable instances more frequently (and notify), and then do a daily sync (without notifications).
2) Or we could simple check the `incident` for it's current status [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/workflow/scheduled.py#L58) and use that to decide when to notify.

kevgliss:
Unless of course, you wish to be able to fully disable workflow notification, which I think we don't have a great story for (notification management).

Cinojose:
> Hi @Cinojose, this is a good catch and is a bug. But I think the fix might be a bit more nuanced. What we really want here is to only send workflow notifications to non-closed incidents. This would be similar to the pattern we use for task syncing: https://github.com/Netflix/dispatch/blob/master/src/dispatch/task/scheduled.py#L85
> 
> We could solve this in a couple of different ways:
> 
> 1. Sync workflows for active stable instances more frequently (and notify), and then do a daily sync (without notifications).
> 2. Or we could simply check the `incident` for its current status [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/workflow/scheduled.py#L58) and use that to decide when to notify.
> 
> I think 1) is probably the better long-term solution but is more work and would be okay with 2) if it unblocks you in the near term.

Thanks, option 1 make sense to me as well.

 I will add the check for the active/stable for this current notification first and then we can add another scheduled task for closed ones. 
 
 BTW another thought/question, may be a separate conversation. Are we planning to add workflow execution to incident timeline ?

PR #3128, opened 2023-03-17
Auto-run workflows associated with Signals and auto-parameterize with Entity Type values

wssheldon:
We parameterize workflows/notebooks with values from entities by assigning the entity type name as the parameter value and then checking if that name exists in the entities for a consumed signal.

### Configure Workflows for a Signal Definition

Screenshot 2023-03-23 at 3 04 44 PM

### Create Workflows using Entity Types in a Dialog

Screenshot 2023-03-23 at 3 06 47 PM

```bash
Creating signal..
Enter instance flow..
Looking for signal_instance.signal.entity_types=[<EntityType #6>]
signal_instance.signal.workflows=[<Workflow #42>]
Running workflow=<Workflow #42>
instance=<WorkflowInstance #48>
entities=[<Entity #2>]
workflow.parameters=[{'key': 'apps', 'value': 'MetaflowARNTest'}]
Got value='MetaflowARNTest' for p={'key': 'apps', 'value': 'MetaflowARNTest'}
Got entity.entity_type.name='MetaflowARNTest' for value='MetaflowARNTest'
params={'apps': 'arn:aws:iam::123456789012:role/Test'}
```


src/dispatch/signal/flows.py
```diff
@@ -46,6 +46,21 @@ def create_signal_instance(
     db_session.commit()
 
     if signal_service.apply_filter_actions(db_session=db_session, signal_instance=signal_instance):
+        # is_configured_workflow()
```
kevgliss, line 49: FYI this section changes a bit in #3131

kevgliss (approved):

PR #3192, opened 2023-03-30
Use Annotated types for FastAPI `db_session`, `get_current_user` parameter

wssheldon:
Initial commit uses Annotated types for DispatchUser and Session in Case view as an example. If the approach looks good to folks (namely where the annotated types live in the codebase) I will add this approach to all the views for consistency.

Reference: https://www.reddit.com/r/Python/comments/11v0kcb/fastapi_0950_supports_and_recommends_annotated/


wssheldon:
Looks like the `*` symbol at the beginning of the view functions can also be removed if we use Annotated types

https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#order-the-parameters-as-you-need-tricks

mvilanova (approved):
Looks great. Thanks for working on it.

PR #3198, opened 2023-03-31
generate a unique monaco uri using a uuid for entity type playground

wssheldon:
logged the uri to test:

`playground-25cb1648-7c86-4616-97ec-d62896b30fc8`


kevgliss:
Could we use the built in browser support for this instead of a package?
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

wssheldon:
For sure, swapped it.

PR #3200, opened 2023-03-31
Remove unnecessary `*` parameter in views, create `CommonParameters` and `CurrentIncident` Annotated types

wssheldon:
unnecessary `*` parameter in views: https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#order-the-parameters-as-you-need-tricks

### DIFF
--- a/src/dispatch/database/service.py
+++ b/src/dispatch/database/service.py
@@ -427,12 +427,12 @@ def get_all(*, db_session, model):
 def common_parameters(
     current_user: CurrentUser,
     db_session: DbSession,
-    page: Annotated[int, Query(gt=0, lt=2147483647)] = 1,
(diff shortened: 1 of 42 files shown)


src/dispatch/database/service.py
```diff
@@ -448,6 +448,12 @@ def common_parameters(
     }
 
 
+CommonParameters = Annotated[
+    dict[str, int | CurrentUser | DbSession | QueryStr | Json | List[str] | List[bool] | UserRoles],
```
wssheldon, line 452: This should likely be a Pydantic model or TypedDict

mvilanova: I'd lean on using a Pydantic model so we can do data validation.

mvilanova (approved):

wssheldon:
Python CI test failure is not related to this change and is resolved in: https://github.com/Netflix/dispatch/pull/3207

PR #3224, opened 2023-04-05
Fixing case creation logic

kevgliss:


src/dispatch/signal/flows.py
```diff
@@ -31,52 +31,52 @@ def signal_instance_create_flow(
...
-
-            conversation_target = None
-            if signal_instance.signal.conversation_target:
-                conversation_target = signal_instance.signal.conversation_target
-
-            case_flows.case_new_create_flow(
+    if signal_service.apply_filter_actions(db_session=db_session, signal_instance=signal_instance):
+        return signal_instance
+
+    if signal_instance.signal.create_case:
```
wssheldon, line 37: I would probably move this up before the filter, my understanding is if this is set to `False` there's really no need to check the filters since we will not be creating a case in any situation.


```suggestion
if not signal_instance.signal.create_case:
    return signal instance
```

src/dispatch/signal/flows.py
```diff
@@ -31,52 +31,52 @@ def signal_instance_create_flow(
...
-            service_id = None
-            if signal_instance.signal.oncall_service:
-                service_id = signal_instance.signal.oncall_service.external_id
-
-            conversation_target = None
-            if signal_instance.signal.conversation_target:
-                conversation_target = signal_instance.signal.conversation_target
-
-            case_flows.case_new_create_flow(
+    if signal_service.apply_filter_actions(db_session=db_session, signal_instance=signal_instance):
```
wssheldon, line 34: This name feels more explicit / easier to read to me

```suggestion
    if signal_service.filter_signal(db_session=db_session, signal_instance=signal_instance):
```

src/dispatch/signal/service.py
```diff
@@ -354,7 +354,7 @@ def apply_filter_actions(
 ) -> Literal[True] | None:
```
wssheldon, line 354:
```suggestion
) -> bool
```

PR #3225, opened 2023-04-05
Adds new command for bulk PR merging and improves release notes command

mvilanova:


wssheldon (approved):
Are there any docs we should update that reference `utils/release_notes.py`?

mvilanova:
> Are there any docs we should update that reference `utils/release_notes.py`?

No OSS docs need to be updated.

PR #3226, opened 2023-04-05
Add Filter Action column to SignalInstance table

wssheldon:
Screenshot 2023-04-05 at 10 54 12 AM

### DIFF
--- a/src/dispatch/static/dispatch/src/signal/TableInstance.vue
+++ b/src/dispatch/static/dispatch/src/signal/TableInstance.vue
@@ -30,6 +30,21 @@
                 {{ item.project.name }}
               </v-chip>
             </template>
+            <template v-slot:item.filter_action="{ item }">
+              <v-chip
+                small
+                text-color="white"
+                :color="
+                  item.filter_action === 'snooze'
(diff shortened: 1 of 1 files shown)


src/dispatch/static/dispatch/src/signal/TableInstance.vue
```diff
@@ -30,6 +30,21 @@
                 {{ item.project.name }}
               </v-chip>
             </template>
+            <template v-slot:item.filter_action="{ item }">
+              <v-chip
+                small
+                text-color="white"
+                :color="
+                  item.filter_action === 'snooze'
+                    ? 'red'
```
kevgliss, line 39: Hmm, I wonder if we should pick something other than red? It's not anything bad and red is our primary color.

kevgliss (approved):

src/dispatch/static/dispatch/src/signal/TableInstance.vue
wssheldon: Yeah I wasn't sure what color to use, I'll do blue / light-blue.

PR #3239, opened 2023-04-07
Fixes an potential race condition when multiple signals were sent at the same time.

kevgliss:


src/dispatch/signal/scheduled.py
```diff
@@ -49,3 +49,25 @@ def consume_signals(db_session: SessionLocal, project: Project):
 
         if signal_instances:
             plugin.instance.delete()
+
+
+@scheduler.add(every(1).minutes, name="signal-process")
+@scheduled_project_task
+def process_signals(db_session: SessionLocal, project: Project):
+    """Processes signals and create cases if appropriate."""
+    signal_instances = (
```
wssheldon, line 58: should we constrain it more somehow? i think this will grow forever and potentially get slow

kevgliss: Anytime a signal is processed it should now have a value for `filter_action`. So only unprocessed signals should get returned here.

PR #3241, opened 2023-04-07
add a button to visit a case in the case popover

wssheldon:


kevgliss:
I think you can just use the `to` param for buttons:

```
<v-btn :to="{
                      name: 'CaseTableEdit',
                      params: { name: item.name },
                    }"
```

wssheldon:
> name: 'CaseTableEdit',
>         params: { name: item.name },

Ah, thanks

PR #3272, opened 2023-04-14
Ensuring looped in users are in the channel

kevgliss:


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -231,6 +231,8 @@ def add_users_to_conversation_thread(
     """Adds user to a threaded conversation."""
     users = [f"<@{user_id}>" for user_id in user_ids]
     if users:
+        # @'ing them isn't enough if they aren't already in the channel
+        add_users_to_conversation(client=client, conversation_id=conversation_id, user_ids=user_ids)
```
wssheldon, line 235: Should we check they're not in the channel already? Looks like it will throw an error here if they are in there already:

`already_in_channel`

Can use: https://api.slack.com/methods/conversations.members

wssheldon: Oh nvm, we already handle that.

wssheldon (approved):

PR #3290, opened 2023-04-21
Aggregate signal instance messages in Case thread

wssheldon:
Screenshot 2023-04-21 at 10 51 04 AM


wssheldon:
@kevgliss The only trade-off introduced here is that we can only show 1 raw signal data (the first one) in slack now, this is OK I think. But, we could probably implement a paging functionality like the list signals modal if we wanted to be able to see them all, barring any blockkit limitations.

kevgliss:
This looks good, I think there is probably a way to clean up the `Seen in...` language to avoid so much repetition but I can't really think of a good way right now.

wssheldon:
> This looks good, I think there is probably a way to clean up the `Seen in...` language to avoid so much repetition but I can't really think of a good way right now.

Agreed, here's a potential alternative, although this feels worse somehow to me. 🤷‍♂️

There's probably some clever blockkit solve here.

Screenshot 2023-04-21 at 11 16 14 AM

wssheldon:
It could be a button that just pops open a modal with this information.

kevgliss:
Yeah, or maybe just pick the most important one? (other cases)?

PR #3314, opened 2023-04-26
Allows for a signal's raw uuid to be our raw id

kevgliss:


src/dispatch/signal/service.py
```diff
@@ -450,6 +450,11 @@ def create_instance(
         signal=signal,
     )
 
+    # if the signal has an existing uuid we propgate it as our primary key
+    if signal_instance_in.raw:
+        if signal_instance_in.raw.get("id"):
+            signal_instance.id = signal_instance_in.raw.id
```
wssheldon, line 456:
```suggestion
            signal_instance.id = signal_instance_in.raw["id"]
```

wssheldon:
`external_id` using the existing `id` format so we'll need to account for that

```python
  if not signal_instance_in.signal:
      external_id = signal_instance_in.raw["id"]
      variant = signal_instance_in.raw["variant"]

      signal = signal_service.get_by_variant_or_external_id(
          db_session=db_session,
          project_id=project.id,
          external_id=external_id,
          variant=variant,
      )

      signal_instance_in.signal = signal
```

PR #3321, opened 2023-04-28
Fixes breaking changes impacting incident ban threads feature

mvilanova:
This PR fixes breaking changes ([here](https://github.com/Netflix/dispatch/pull/2913/files#diff-be932aea97556035df57c47ea4fa56c2ca3b7441226f65595ef4ef2f48c422ab) and [here](https://github.com/Netflix/dispatch/pull/2913/files#diff-4cef4776fc86766f7ccd0412bfe00a095a19cc90145629e0a3cf6b35081c8913)) introduced in this [PR](https://github.com/Netflix/dispatch/pull/2913).

Screenshot 2023-04-28 at 11 15 26 AM

### DIFF
--- a/src/dispatch/conversation/service.py
+++ b/src/dispatch/conversation/service.py
@@ -1,57 +1,33 @@
 from typing import Optional
 
-from dispatch.case.models import Case
-from dispatch.event import service as event_service
-from dispatch.incident.models import Incident
-from dispatch.project.models import Project
-
 from .models import Conversation, ConversationCreate, ConversationUpdate
 
 
 def get(*, db_session, conversation_id: int) -> Optional[Conversation]:
-    """Fetch a conversation by its `conversation_id`."""
+    """Gets a conversation by its id."""
     return db_session.query(Conversation).filter(Conversation.id == conversation_id).one_or_none()
 
 
-def get_by_channel_id_ignoring_channel_type(
-    db_session, channel_id: str, thread_id: Optional[str] = None
-) -> Optional[Conversation]:
-    """Fetch a conversation by its `channel_id` ignoring the channel type
-    and update the channel id in the database if the channel type has changed."""
-    if thread_id:
-        conversation = (
-            db_session.query(Conversation)
-            .join(Case)
-            .join(Project)
-            .filter(Conversation.channel_id.contains(channel_id[1:]))
-            .filter(Conversation.thread_id == thread_id)
-            .one_or_none()
-        )
-
-    else:
-        conversation = (
-            db_session.query(Conversation)
-            .join(Incident)
-            .join(Project)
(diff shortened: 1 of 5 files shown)


src/dispatch/participant/flows.py
```diff
@@ -12,10 +12,9 @@
     ParticipantRoleType,
     ParticipantRoleCreate,
 )
+from dispatch.participant import service as participant_service
```
mvilanova, line 15: I had to import the service module this way to fix a circular import I stumbled upon. Not sure why it did not manifest before.

src/dispatch/conversation/service.py
```diff
@@ -1,59 +1,35 @@
...
                 conversation=conversation,
                 conversation_in=conversation_in,
             )
 
-            event_service.log_incident_event(
-                db_session=db_session,
-                source="Dispatch Core App",
-                description=f"Slack conversation type has changed ({channel_id[0]} -> {conversation.channel_id[0]})",
-                incident_id=conversation.incident_id,
-            )
```
mvilanova, line 55: Removed the event to avoid having to add logic to determine if we should record it in a case or incident.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -780,15 +780,14 @@ def handle_thread_creation(
     if not context["config"].ban_threads:
         return
 
-    if context["subject"].type == "incident":
```
mvilanova, line 783: I think this was leftover that wasn't removed.

src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -27,33 +27,33 @@
...
                     organization_slug=slug,
-                    project_id=conversation.case.project_id,
+                    project_id=conversation.incident.project_id,
                 )
             else:
                 subject = SubjectMetadata(
-                    type="incident",
-                    id=conversation.incident_id,
+                    type="case",
+                    id=conversation.case_id,
```
wssheldon, line 54: I added some Subject type enums we could leverage here:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/models.py#L52-L58

Although, these should probably be moved to the top level `enum.py` file as they're not specific to Slack.

wssheldon (approved):
Everything looks good on the Case side. I tested:

- [x] Case creation
- [x] Main case message updates (edit, resolve, etc)
- [x] Signal deduplication thread message updates
- [x] Engagement thread messages

src/dispatch/plugins/dispatch_slack/middleware.py
mvilanova: Ok, I'll address it in a separate PR.

PR #3331, opened 2023-05-01
Conditional creation of case resources for perf improvement on signal ingest

wssheldon:
Resources are still created successfully on case escalation

Screenshot 2023-05-01 at 3 34 46 PM

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -208,8 +208,17 @@ def case_new_create_flow(
     # we create the ticket
     ticket_flows.create_case_ticket(case=case, db_session=db_session)
 
+    individual_participants, team_participants = get_case_participants(
+        case=case, db_session=db_session
+    )
+
     if create_resources:
-        case_create_resources_flow(db_session=db_session, case_id=case.id)
+        case_create_resources_flow(
+            db_session=db_session,
+            case_id=case.id,
+            individual_participants=individual_participants,
+            team_participants=team_participants,
+        )
 
     if case.case_priority.page_assignee:
         if not service_id:
@@ -260,8 +269,6 @@ def case_new_create_flow(
                     description="Conversation added to case",
                     case_id=case.id,
                 )
-                individual_participants, _ = get_case_participants(case=case, db_session=db_session)
-
                 # wait until all resources are created before adding suggested participants
                 individual_participants = [x.email for x, _ in individual_participants]
 
@@ -669,13 +676,11 @@ def case_assign_role_flow(
     role_flow.assign_role_flow(case, assignee_email, assignee_role, db_session)
 
 
-def case_create_resources_flow(db_session: Session, case_id: int) -> None:
+def case_create_resources_flow(
+    db_session: Session, case_id: int, individual_participants: list, team_participants: list
+) -> None:
     """Runs the case resource creation flow."""
     case = get(db_session=db_session, case_id=case_id)
-
-    individual_participants, team_participants = get_case_participants(
-        case=case, db_session=db_session
-    )
 
     if case.assignee:
         individual_participants.append((case.assignee.individual, None))
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -623,23 +623,23 @@ def case_to_incident_endpoint_escalate_flow(
     case.status = CaseStatus.escalated
 
     # we run the incident create flow
-    incident_flows.incident_create_flow(
+    incident = incident_flows.incident_create_flow(
         incident_id=incident_id, organization_slug=organization_slug, db_session=db_session
     )
 
     # fetch the incident from the database after the background task has completed
     incident = incident_service.get(db_session=db_session, incident_id=incident_id)
-    if incident:
-        case.incidents.append(incident)
-        db_session.add(case)
-        db_session.commit()
+    case.incidents.append(incident)
 
-        event_service.log_case_event(
-            db_session=db_session,
-            source="Dispatch Core App",
-            description=f"The case has been linked to incident {incident.name} in the {incident.project.name} project",
-            case_id=case.id,
-        )
+    db_session.add(case)
+    db_session.commit()
+
+    event_service.log_case_event(
+        db_session=db_session,
+        source="Dispatch Core App",
+        description=f"The case has been linked to incident {incident.name} in the {incident.project.name} project",
+        case_id=case.id,
+    )
 
     # we add the incident's tactical group to the case's storage folder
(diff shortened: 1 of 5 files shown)


src/dispatch/case/flows.py
```diff
@@ -660,20 +623,23 @@ def case_to_incident_endpoint_escalate_flow(
...
-    case.incidents.append(incident)
 
-    db_session.add(case)
-    db_session.commit()
+    # fetch the incident from the database after the background task has completed
+    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+    if incident:
+        case.incidents.append(incident)
+        db_session.add(case)
+        db_session.commit()
```
wssheldon, line 635:
```
ERROR:Can't flush None value found in collection Case.incidents:/Users/wshel/Projects/dispatch/src/dispatch/decorators.py:wrapper:92


File "/Users/wshel/Projects/dispatch/src/dispatch/case/flows.py", line 632, in case_to_incident_endpoint_escalate_flow
    db_session.commit()
```

Potential solution to race condition

src/dispatch/case/flows.py
```diff
@@ -660,20 +623,23 @@ def case_to_incident_endpoint_escalate_flow(
...
+    incident_flows.incident_create_flow(
         incident_id=incident_id, organization_slug=organization_slug, db_session=db_session
     )
-    case.incidents.append(incident)
 
-    db_session.add(case)
-    db_session.commit()
+    # fetch the incident from the database after the background task has completed
+    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+    if incident:
```
wssheldon, line 632:
```suggestion
    if incident := incident_service.get(db_session=db_session, incident_id=incident_id):
```

Use walrus operator as this var is used nowhere else

src/dispatch/case/flows.py
```diff
@@ -197,6 +199,7 @@ def case_new_create_flow(
     conversation_target: str = None,
     service_id: int = None,
     db_session=None,
+    create_resources: bool = True,
```
mvilanova, line 202: Should we make this configurable in the UI at case type level?

src/dispatch/case/flows.py
```diff
@@ -209,48 +212,8 @@ def case_new_create_flow(
         case=case, db_session=db_session
     )
```
mvilanova, line 213: Should we remove this code block? We already have it inside the case resources creation flow.
```
    individual_participants, team_participants = get_case_participants(
        case=case, db_session=db_session
    )
```

src/dispatch/individual/service.py
```diff
@@ -74,6 +77,11 @@ def get_all(*, db_session) -> List[Optional[IndividualContact]]:
     return db_session.query(IndividualContact)
 
 
+@lru_cache(maxsize=1000)
+def fetch_individual_info(contact_plugin: PluginInstance, email: str, db_session: Session):
```
mvilanova, line 81: What's the return type here, `dict`?

wssheldon: Yeah, I think `dict[str, str]`. Although it depends on your plugin I suppose, so I didn't add it.

src/dispatch/case/flows.py
```diff
@@ -209,48 +212,8 @@ def case_new_create_flow(
...
-
-    # we update the ticket
-    ticket_flows.update_case_ticket(case=case, db_session=db_session)
-
-    # we update the case document
-    document_flows.update_document(
-        document=document, project_id=case.project.id, db_session=db_session
-    )
+    if create_resources:
+        case_create_resources_flow(db_session=db_session, case_id=case.id)
```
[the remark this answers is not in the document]
wssheldon: I was wondering that but I wasn't sure. It's core to the workflow and if we moved it to `case_create_resources_flow` it would be too binary.  We have the `conversation_target` flag in the Signal, which is our conditional for creating a conversation or not. 

```python
        if conversation_target:
            try:
                # TODO: Refactor conversation creation using conversation_flows module
                conversation = create_conversation(case, conversation_target, db_session)
                conversation_in = ConversationCreate(
```
Also, we should probably use convo flow here or create a new flow for this.

src/dispatch/case/flows.py
wssheldon: I kept it here because of the dependency the conversation creation has on `individual_participants`. This could be handled better.

src/dispatch/case/flows.py
wssheldon: Yeah, I can add in a follow up PR.

src/dispatch/case/flows.py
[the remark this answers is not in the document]
mvilanova: I thought the plan was to not create new conversations, but leveraging existing and configured ones in Dispatch for sending notifications and threading them. What am I missing?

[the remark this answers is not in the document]
mvilanova: If we still need to create a conversation and this depends on `individual_participants` and `team_participants`, should these be passed as parameters to `case_create_resources_flow` to avoid fetching them twice?

src/dispatch/case/flows.py
```diff
@@ -660,20 +623,23 @@ def case_to_incident_endpoint_escalate_flow(
     case.status = CaseStatus.escalated
 
     # we run the incident create flow
-    incident = incident_flows.incident_create_flow(
+    incident_flows.incident_create_flow(
```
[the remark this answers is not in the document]
mvilanova: Were you able to repro?

src/dispatch/case/flows.py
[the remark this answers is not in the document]
wssheldon: > I thought the plan was to not create new conversations, but leveraging existing and configured ones in Dispatch for sending notifications and threading them. What am I missing?

Only net new (non-deduplicated/snooze) Signals should ever enter the `case_new_create_flow`.

Filtered ones update the existing thread:

```python
    # we don't need to continue if a filter action took place
    if signal_service.filter_signal(db_session=db_session, signal_instance=signal_instance):
        # If the signal was deduplicated, we can assume a case exists,
        # and we need to update the corresponding signal message
        if (
            signal_instance.filter_action == SignalFilterAction.deduplicate
            and signal_instance.case.signal_thread_ts  # noqa
        ):
            update_signal_message(
                db_session=db_session,
                signal_instance=signal_instance,
            )
        return signal_instance
```

src/dispatch/case/flows.py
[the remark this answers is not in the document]
wssheldon: I removed the handling, will handle as separate PR.

PR #3337, opened 2023-05-02
Optimize signal ingestion pipeline and create perf-test cli util

wssheldon:
This PR introduces a blocking thread-safe FIFO queue in the signal ingestion pipeline. This ensures signals are processed in the order they were received and prevents race conditions where a signal instance can override a shared value (e.g. case associations). The queue has a max size of 500 that is also used to limit the number of instances retrieved by the query in `process_signals`. We also create a `scoped_session` for each run of the scheduler.

A boundary of `.filter(SignalInstance.created_at >= one_hour_ago)` was added to the ingestion query to prevent cases where we are unable to keep up for an hour, which should never happen, and if it does is an acceptable scenario to prevent performance issues and outages.

Caching logic was introduced in `_should_update_signal_message` to prevent the Slack `chat.update` method from being ratelimited. 

Finally, a stress test CLI utility was created that attempts to replicate signal ingestion via the `/instances` API endpoint. 

Usage:

```bash
dispatch scheduler perf-test --api-token redacted` --num-instances 5000
...
Sent 5000 of 5000 signal instances:/Users/wshel/Projects/dispatch/src/dispatch/cli.py:send_signal_instances:769
Elapsed time: 281.17 seconds
```

Screenshot 2023-05-02 at 8 56 19 AM

I tested a push of 5000 instances from my M1 MBP.

timer results:

Non-deduplicated signal instance (calls many external API's) and results in *new* case.

```bash
DEBUG:function.elapsed.time.dispatch.signal.scheduled.process_signal_instance: 29.435003541992046
```

Deduplicated signal instance.

```bash
DEBUG:function.elapsed.time.dispatch.signal.scheduled.process_signal_instance: 0.0374222089885734
```

We can play around with these configurable options to see what makes the most sense:

`@scheduler.add(every(1).minutes, name="signal-process")`
`MAX_SIGNAL_INSTANCES = 500`

### DIFF
--- a/src/dispatch/cli.py
+++ b/src/dispatch/cli.py
@@ -1,7 +1,9 @@
 import logging
 import os
+import time
 
 import click
+import requests
 import uvicorn
 from dispatch import __version__, config
 from dispatch.enums import UserRoles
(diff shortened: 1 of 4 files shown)


src/dispatch/cli.py
```diff
@@ -689,6 +691,101 @@ def start_tasks(tasks, exclude, eager):
     scheduler.start()
 
 
+@dispatch_scheduler.command("perf-test")
+@click.option("--num-instances", default=1000, help="Number of signal instances to send.")
+@click.option(
+    "--api-endpoint",
+    default="http://localhost:8080/api/v1/default/signals/instances",
```
mvilanova, line 698:
```suggestion
    default=f"{DISPATCH_UI_URL}/api/v1/default/signals/instances",
```

mvilanova: We will need to add an import for it.

src/dispatch/cli.py
```diff
@@ -689,6 +691,101 @@ def start_tasks(tasks, exclude, eager):
...
+@click.option(
+    "--api-token",
+    required=True,
+    help="API token to use.",
+)
+@click.option(
+    "--project",
+    default="Test",
+    required=True,
+    help="The Dispatch project to send to send the instances to",
```
mvilanova, line 711:
```suggestion
    help="The Dispatch project to send the instances to",
```

src/dispatch/cli.py
```diff
@@ -689,6 +691,101 @@ def start_tasks(tasks, exclude, eager):
...
+)
+def perf_test(num_instances: int, api_endpoint: str, api_token: str, project: str) -> None:
+    """Performance testing utility for creating signal instances."""
+    import concurrent.futures
+    from fastapi import status
+
+    NUM_SIGNAL_INSTANCES = num_instances
+    NUM_WORKERS = 10
```
mvilanova, line 719: Should we parameterize this one as well?

src/dispatch/cli.py
```diff
@@ -689,6 +691,101 @@ def start_tasks(tasks, exclude, eager):
...
+        {
+            "project": {"name": project},
+            "raw": {},
+        },
+    ] * NUM_SIGNAL_INSTANCES
+
+    send_signal_instances(api_endpoint, api_token, signal_instances)
+
+    elapsed_time = time.time() - start_time
+    print(f"Elapsed time: {elapsed_time:.2f} seconds")
```
mvilanova, line 786:
```suggestion
    click.echo(f"Elapsed time: {elapsed_time:.2f} seconds")
```

mvilanova (changes requested):

PR #3338, opened 2023-05-02
Adds 250/minute rate limit to SignalInstance POST API endpoint using `slowapi`

wssheldon:


src/dispatch/signal/views.py
```diff
@@ -52,11 +53,14 @@ def get_signal_instances(common: CommonParameters):
 
 
 @router.post("/instances", response_model=SignalInstanceRead)
+@limiter.limit("250/minute")
```
mvilanova, line 56: Why 250 and not a different value? Should we make it configurable via the Web UI?

mvilanova (approved):

src/dispatch/signal/views.py
wssheldon: I chose this since it's only a bit higher to our limit set in `signal-process` without being too restrictive (it is more than double that limit, to be fair). I think it will take some tuning to get it just right, so open to ideas here.

wssheldon:
```
DEBUG:http://localhost:8080 "POST /api/v1/default/signals/instances HTTP/1.1" 429 28:/Users/wshel/.pyenv/versions/3.11.2/envs/dispatch-prod/lib/python3.11/site-packages/urllib3/connectionpool.py:_make_request:456
INFO:Response: {'detail': '10 per 1 minute'}:/Users/wshel/Projects/dispatch/src/dispatch/cli.py:_send_signal_instance:749
ERROR:Unable to send finding. Reason: 429 Client Error: Too Many Requests for url: http://localhost:8080/api/v1/default/signals/instances Response: N/A:/Users/wshel/Projects/dispatch/src/dispatch/cli.py:_send_signal_instance:758
```

PR #3339, opened 2023-05-02
limit entity select to 100 and only show entities for signal

wssheldon:


src/dispatch/plugins/dispatch_slack/fields.py
```diff
@@ -594,22 +594,22 @@ def case_type_select(
...
     entity_options = [
         {"text": entity.value, "value": entity.id}
-        for entity in entity_service.get_all(db_session=db_session, project_id=project_id).all()
+        for entity in entity_service.get_all_by_signal(db_session=db_session, signal_id=signal_id)
     ]
 
     return multi_select_block(
         placeholder="Select Entities",
-        options=entity_options,
+        options=entity_options[:100],  # Limit the entities to the first 100
```
mvilanova, line 612: Is there a way we can be smarter on which ones we offer as an option?

mvilanova (approved):

src/dispatch/plugins/dispatch_slack/fields.py
wssheldon: Updated to return the 100 most recent entities. There's probably something clever we can do for common use cases (people likely want to snooze when their username shows up, for example). But, this is probably good for now.

PR #3345, opened 2023-05-02
Filters out Slack events from USLACKBOT

mvilanova:


wssheldon:
This message will be surface to the user:

```python
def build_context_error_message(payload: dict, error: Any) -> str:
    message = (
        f"""I see you tried to run `{payload['command']}` in an non-incident conversation. Incident-specifc commands can only be run in incident conversations."""  # command_context_middleware()
        if payload.get("command")
        else str(error)  # everything else
    )
    return message
```

Is that appropriate for this error case or should we add a new exception type?

mvilanova:
> This message will be surface to the user:
> 
> ```python
> def build_context_error_message(payload: dict, error: Any) -> str:
>     message = (
>         f"""I see you tried to run `{payload['command']}` in an non-incident conversation. Incident-specifc commands can only be run in incident conversations."""  # command_context_middleware()
>         if payload.get("command")
>         else str(error)  # everything else
>     )
>     return message
> ```
> 
> Is that appropriate for this error case or should we add a new exception type?
Yeah, I think that would make sense, as I'm seeing `ContextError()` being used in other places in the middleware module and I'm not sure there's a lot of value to present them to the user.

PR #3351, opened 2023-05-03
add variant to case message if there's a signal instance

wssheldon:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -35,6 +35,17 @@
...
+    fields = [
+        f"*Assignee* \n {case.assignee.individual.email}",
+        f"*Status* \n {case.status}",
+        f"*Severity* \n {case.case_severity.name}",
+        f"*Type* \n {case.case_type.name}",
+        f"*Priority* \n {case.case_priority.name}",
+    ]
+
+    if case.signal_instances:
+        fields.append(f"*Variant* \n {case.signal_instances[0].signal.variant}")
```
mvilanova, line 47: Thoughts on making the variant value a link to the signal definition external link if this exists?

mvilanova (approved):

PR #3365, opened 2023-05-05
Optimize ORM usage, db_session instantiation, and tuning

wssheldon:
This change introduces performance improvements for deduplicated Signal instances that are ingested. By optimizing the ORM usage across the flow (primarily by removing loads of unnecessary relationships, columns, and rows) this path is about 4x faster. To process 500 deduplicated instances, it takes roughly 10 seconds. 

The `create_signal_messages` had the most room for improvement, because it unnecessarily was loading all signal_instances and their associated raw data. It previously took about 3 seconds (and exponentially more as more and more signals are ingested). It is about 5-6x faster.

Before:
`DEBUG:function.elapsed.time.dispatch.plugins.dispatch_slack.case.messages.create_signal_messages: 3.1490371670006425`

After:
`DEBUG:function.elapsed.time.dispatch.plugins.dispatch_slack.case.messages.create_signal_messages: 0.056947083001432475:/Users/wshel/Projects/dispatch/src/dispatch/decorators.py:wrapper:185`

This implementation takes 10-15 seconds to process (deduplicated) 500 instances. Creating a new case from scratch (non-deduped) now takes about 6 seconds. It previously took roughly 30 seconds. 

`DEBUG:function.elapsed.time.dispatch.signal.scheduled.process_signal_instance: 29.435003541992046`

This is primarily due to the [removal of external resources](https://github.com/Netflix/dispatch/pull/3331) such as Google docs, drive, etc. 

The default dedupe filter is also first by fetching one row to determine if we dedupe instead of all of them in the time frame and only fetching the necessary column (case_id).

Based on these stats I think it makes sense to increase the duration the scheduler runs to every 20 seconds. This gives some buffer for net new case creation and makes the most efficient use of each run instead of waiting when it could be processing new signals. 

I also moved scoped database connection instantiation outside of the process signal function and into its own decorator, so that it's scoped on each scheduled task run, which saves us database connections.

On my M1 MBP I was able to send 25k signal instances across 10 threads and keep up with 25 max db connections.

Screenshot 2023-05-05 at 10 04 21 AM

**Testing**
`pytest -v`
```
174 passed, 3 skipped in 7.85s
```

### DIFF
--- a/src/dispatch/decorators.py
+++ b/src/dispatch/decorators.py
@@ -1,8 +1,12 @@
+from contextlib import _GeneratorContextManager, contextmanager
 from functools import wraps
-from typing import Any, List
+from typing import Any, Callable, List
 import inspect
 import logging
 import time
+
+from sqlalchemy.engine import Engine
+from sqlalchemy.orm import scoped_session, Session
 
 from dispatch.metrics import provider as metrics_provider
 from dispatch.organization import service as organization_service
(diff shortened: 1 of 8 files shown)


src/dispatch/decorators.py
```diff
@@ -19,39 +23,114 @@ def fullname(o):
...
+                    is_scoped=False,
+                ) as __session:
+                    __execute_task_within_session_context(__session)
+            else:
+                with _session(
+                    engine=schema_engine,
+                    is_scoped=True,
+                ) as __session:
+                    __execute_task_within_session_context(__session)
+
```
wssheldon, line 87: The `is_scoped` conditional block here is unnecessary. Pass `is_scoped` directly.

kevgliss:
I've modified the approach here a little bit. Instead of relying on the scheduler, we have one dedicated thread that processes all new signals in some reasonable order. 

With the other improvements around case creation, this is likely "good enough" for now and we can look into running multiple of these processes concurrently if required (instead of a scheduler doing it for us). 

I also added a`wrk` lua script to aid performance testing.

kevgliss (approved):

PR #3367, opened 2023-05-05
Add common modal for success that catches not_found exception

wssheldon:


src/dispatch/plugins/dispatch_slack/modals/common.py
```diff
@@ -0,0 +1,33 @@
...
+
+    try:
+        client.views_update(
+            view_id=view_id,
+            trigger_id=trigger_id if trigger_id else None,
+            view=modal,
+        )
+    except SlackApiError as e:
+        if e.response["error"] == "not_found":
+            e.add_note("This error usually indiciates that user closed the loading modal early.")
```
wssheldon, line 32:
```suggestion
            e.add_note("This error usually indicates that the user closed the loading modal early and is transparent.")
```

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -690,16 +686,11 @@ def _create_snooze_filter(
...
-                blocks=[Section(text="Adding Snooze... Success!")],
-            ).build()
-
-            client.views_update(
+            send_success_modal(
+                client=client,
                 view_id=body["view"]["id"],
-                view=modal,
+                title="Add Snooze",
+                message="Snooze Filter added succesfully.",
```
wssheldon, line 693:
```suggestion
                message="Snooze Filter added successfully.",
```

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -1670,14 +1654,11 @@ def handle_engagement_deny_submission_event(
...
-        close="Close",
-        blocks=[Section(text="Confirming event is not expected... success!")],
-    ).build()
-    client.views_update(
+    send_success_modal(
+        client=client,
         view_id=body["view"]["id"],
-        view=modal,
+        title="Confirm",
+        message="Confirmed event is not expected.",
```
kevgliss, line 1661:
```suggestion
        message="Event has been confirmed as not expected.",
```

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -44,11 +44,11 @@ def handle_slack_error(exception: SlackApiError, endpoint: str, kwargs: dict) ->
     )
     error = exception.response["error"]
 
-    if error in {"channel_not_found", "user_not_in_channel"}:
+    if error in {SlackAPIErrorCode.CHANNEL_NOT_FOUND, SlackAPIErrorCode.USER_NOT_IN_CHANNEL}:
```
kevgliss, line 47: Is there a reason to use `{}` here? Maybe we could use a `[]` or even better `()` (if we care about non-mutability)

kevgliss (approved):

PR #3369, opened 2023-05-07
Allow customizing notification templates per project

vindex10:
There is a way to have customizable templates for dispatch today, by mounting a volume with a customized templates dir as part of the docker compose:

```
  web:
    image: dispatch-local
    ...
    volumes:
      - "../dispatch-templates/messaging-email-templates:/usr/local/lib/python3.11/site-packages/dispatch/messaging/email/templates"
```

This approach does allow to customize the templates "deployment"-wide.

In this PR I suggest a small extension of this functionality. You can create a folder per project (identified by project id):

```
dispatch/messaging/email/templates/project_id/<project_id>/base.mjml
```

This will be used at the first place if exists, otherwise will gracefully fall back to the default template like "/templates/base.mjml".

I was a bit afraid to use "slug" of the project name, because this requires more careful work on escaping system paths and making sure we don't expose something by engineering a weird project name. This can be expanded  in future though, by introducing another subdir "templates/project_slug/<project_slug>".

### DIFF
--- a/src/dispatch/messaging/email/utils.py
+++ b/src/dispatch/messaging/email/utils.py
@@ -4,6 +4,7 @@ import subprocess
 import tempfile
 
 from dispatch.config import MJML_PATH
+from dispatch.project.models import Project
 
 
 from dispatch.messaging.strings import (
@@ -20,7 +21,7 @@ from .filters import env
 log = logging.getLogger(__name__)
 
 
-def get_template(message_type: MessageType):
+def get_template(message_type: MessageType, project: Project):
     """Fetches the correct template based on the message type."""
     template_map = {
         MessageType.incident_executive_report: ("executive_report.mjml", None),
@@ -45,19 +46,26 @@ def get_template(message_type: MessageType):
         ),
     }
 
-    template_path, description = template_map.get(message_type, (None, None))
+    template_key, description = template_map.get(message_type, (None, None))
 
-    if not template_path:
+    if not template_key:
         raise Exception(f"Unable to determine template. MessageType: {message_type}")
 
-    return env.get_template(os.path.join("templates", template_path)), description
+    template_path = os.path.join("templates", "project_id", f"{project.id}", template_key)
+    try:
+        template = env.get_template(template_path)
+    except FileNotFoundError:
+        template_path = os.path.join("templates", template_key)
+        template = env.get_template(template_path)
+
+    return template, description
 
 
 def create_multi_message_body(
-    message_template: dict, message_type: MessageType, items: list, **kwargs
+        message_template: dict, message_type: MessageType, items: list, project: Project, **kwargs
 ):
     """Creates a multi message message body based on message type."""
-    template, description = get_template(message_type)
+    template, description = get_template(message_type, project)
 
     master_map = []
     for item in items:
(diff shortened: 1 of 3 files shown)


kevgliss:
This looks great @vindex10! As this approach fallbacks gracefully, I see no reason why it can't be merged. Thanks for the contribution. 

As far as documentation, we don't currently have a good space for something like this. Maybe we could create new page for "Messaging" related settings here: https://netflix.github.io/dispatch/docs/administration/settings ?

vindex10:
Sounds good! Could also include information about the recent markdown enhancement in there :)

src/dispatch/messaging/email/utils.py
```diff
@@ -45,19 +46,26 @@ def get_template(message_type: MessageType):
...
+    if not template_key:
         raise Exception(f"Unable to determine template. MessageType: {message_type}")
 
-    return env.get_template(os.path.join("templates", template_path)), description
+    try:
+        template_path = os.path.join("templates", "project_id", f"{project_id}", template_key)
+        template = env.get_template(template_path)
+    except jinja2.exceptions.TemplateNotFound:
+        template_path = os.path.join("templates", template_key)
+        template = env.get_template(template_path)
```
mvilanova, line 59:
```suggestion
    try:
        template_path = os.path.join("templates", "project_id", f"{project_id}", template_key)
    except jinja2.exceptions.TemplateNotFound:
        template_path = os.path.join("templates", template_key)

template = env.get_template(template_path)
```

src/dispatch/messaging/email/utils.py
```diff
@@ -45,19 +46,26 @@ def get_template(message_type: MessageType):
...
-    if not template_path:
+    if not template_key:
         raise Exception(f"Unable to determine template. MessageType: {message_type}")
 
-    return env.get_template(os.path.join("templates", template_path)), description
+    try:
+        template_path = os.path.join("templates", "project_id", f"{project_id}", template_key)
+        template = env.get_template(template_path)
+    except jinja2.exceptions.TemplateNotFound:
+        template_path = os.path.join("templates", template_key)
```
mvilanova, line 58: We may want to add log a warning using `log.warn()`, if the template is not found and the default is used.

src/dispatch/messaging/email/utils.py
```diff
@@ -45,19 +46,26 @@ def get_template(message_type: MessageType):
...
+        template = env.get_template(template_path)
+
+    return template, description
 
 
 def create_multi_message_body(
-    message_template: dict, message_type: MessageType, items: list, **kwargs
+        message_template: dict, message_type: MessageType, items: list, project_id: int, **kwargs
```
mvilanova, line 65: Indentation seems to be wrong here.

mvilanova (changes requested):

mvilanova:
@vindex10 can you run `black` against the files you changed? It will fix the formatting issues the lint tests are complaining about.

src/dispatch/messaging/email/utils.py
vindex10: I ran through black

src/dispatch/messaging/email/utils.py
vindex10: we want to call "get template" to be able to catch TemplateNotFound :)

src/dispatch/messaging/email/utils.py
vindex10: But since this feature was not available before and nobody really requested it, sounds like most of the times the warning will be issued :)

do you think debug log will be more appropriate here? which will show what was the template path resolved?

vindex10: I added the debug for now, here is the example of how I tested it:

```
dispatch-web-1  | DEBUG:Resolved template path: templates/project_id/6/notification.mjml:/usr/local/lib/python3.11/site-packages/dispatch/messaging/email/utils.py:get_template:60
```

vindex10:
I definitely am, @kevgliss :) please do! Thank you!

kevgliss:
Thanks! This will be added in our next release.

PR #3395, opened 2023-05-17
Adds extact matching on name column when searching

kevgliss:
Allows for search to also return exact values in additional to fuzzy searching based on stemming for the name column.


metroid-samus (commented):
LGTM

PR #3398, opened 2023-05-17
Adds a new project settings section for managing search filters

mvilanova:
This PR adds a new section to the project settings for viewing and deleting search filters. Options for creating new and editing existing filters will be added in the future. We only allow users with admin, manager, and owner role to delete filters.

Screenshot 2023-05-17 at 5 52 32 PM

Screenshot 2023-05-17 at 5 54 37 PM


src/dispatch/static/dispatch/src/search/Table.vue
```diff
@@ -0,0 +1,166 @@
...
+  components: { DeleteDialog },
+
+  data() {
+    return {
+      headers: [
+        { text: "Name", value: "name", align: "left", width: "10%" },
+        { text: "Description", value: "description", sortable: false },
+        { text: "Individuals", value: "individuals" },
+        { text: "Teams", value: "teams" },
+        { text: "Services", value: "services" },
```
mvilanova, line 115: Do we need columns with counters for other data types, e.g. Signals?

kevgliss: We don't use `SearchFilters` for signals they have their own model `SignalFilters`.

kevgliss:
Do we want to add more permissions to these? I think they drive quite a lot of functionality, creating new filters is probably fine. But deleting/modifying/disabling existing filters is likely to be problematic.

mvilanova:
> Do we want to add more permissions to these? I think they drive quite a lot of functionality, creating new filters is probably fine. But deleting/modifying/disabling existing filters is likely to be problematic. 

Yeah, I can add permissions in a separate PR.

src/dispatch/static/dispatch/src/search/Table.vue
mvilanova: Looks like I missed Notification. I'll add it.

src/dispatch/static/dispatch/src/search/Table.vue
```diff
@@ -0,0 +1,170 @@
...
+            </template>
+            <template #[`item.teams`]="{ item }">
+              <v-chip small color="info" text-color="white">{{ item.teams.length }}</v-chip>
+            </template>
+            <template #[`item.services`]="{ item }">
+              <v-chip small color="info" text-color="white">{{ item.services.length }}</v-chip>
+            </template>
+            <template #[`item.notifications`]="{ item }">
+              <v-chip small color="info" text-color="white">{{ item.notifications.length }}</v-chip>
+            </template>
```
mvilanova, line 56: TODO: show list of individuals, teams, services, and notifications on click.

PR #3419, opened 2023-05-23
Updates documentation about prerequisites

metroid-samus:
Dispatch uses the `|` operator and requires python 3.10+.


kevgliss:
Thanks!

PR #3433, opened 2023-05-25
Example of purpose built dashboard queries

wssheldon:
Screenshot 2023-05-25 at 12 47 32 PM
Screenshot 2023-05-25 at 12 47 27 PM
Screenshot 2023-05-25 at 12 49 33 PM

Crude example, this approach helps w/ performance a lot, and has added benefit of each card loading async without waiting for main one. We'd need to add filtering support.


src/dispatch/case/service.py
```diff
@@ -80,6 +83,16 @@ def get_all_by_status(*, db_session, project_id: int, status: str) -> List[Optio
     )
 
 
+def get_count_of_case_types(*, db_session) -> list[tuple[str, int]]:
```
wssheldon, line 86: should be `case/type` dir

src/dispatch/case/views.py
```diff
@@ -246,3 +246,16 @@ def delete_case(
                 }
             ],
         ) from None
+
+
+@router.get(
```
wssheldon, line 251: should be in `case/type`

src/dispatch/static/dispatch/src/dashboard/case/CaseTypeBarChartCard.vue
```diff
@@ -40,17 +40,15 @@ export default {
...
   created: function () {
-    CaseTypeApi.getAll({ itemsPerPage: -1 }).then((response) => {
-      this.types = map(
-        filter(response.data.items, function (item) {
-          return !item.exclude_from_metrics
-        }),
-        "name"
-      )
+    CaseApi.getTypeCount().then((response) => {
+      console.log(response)
```
wssheldon, line 49:
```suggestion
```

src/dispatch/static/dispatch/src/dashboard/case/CaseTypeBarChartCard.vue
```diff
@@ -40,17 +40,15 @@ export default {
...
 
   created: function () {
-    CaseTypeApi.getAll({ itemsPerPage: -1 }).then((response) => {
-      this.types = map(
-        filter(response.data.items, function (item) {
-          return !item.exclude_from_metrics
-        }),
-        "name"
-      )
+    CaseApi.getTypeCount().then((response) => {
```
wssheldon, line 48:
```suggestion
    CaseTypeApi.getTypeCount().then((response) => {
```

would need to move method to apply

src/dispatch/case/service.py
mvilanova: Could we generalize and parameterize this function to accept the objects for which we want the counts, so we don't have to write one for each data type?

wssheldon: Yeah potentially, it may get tricky if we need to join across tables.

PR #3449, opened 2023-05-31
Updates Dispatch Conversation URLs to reflect complete workspace URL

metroid-samus:
The Conversation URL for incidents previously pointed to slack.com instead of  \<domain\>.slack.com, which may sometime redirection to different slack space depending on the workspaces a user is signed into in their browser.


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -281,13 +286,18 @@ def send_message(
...
+    )["team"]["domain"]
+
     if persist:
         add_pin(client, response["channel"], response["ts"])
 
     return {
         "id": response["channel"],
         "timestamp": response["ts"],
-        "weblink": f"https://slack.com/app_redirect?channel={response['id']}",  # TODO should we fetch the permalink?
+        "weblink": f"https://{domain}.slack.com/app_redirect?channel={response['id']}",  # TODO should we fetch the permalink?
```
mvilanova, line 300: I don't think fetching the domain is necessary for sending or updating messages. AFAIR we're not using the response for anything. If we did, we should use this endpoint to get the permalink to the message instead: https://api.slack.com/methods/chat.getPermalink

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -308,10 +318,15 @@ def update_message(
...
+    domain = make_call(
+        client,
+        SlackAPIGetEndpoints.team_info
+    )["team"]["domain"]
+
     return {
         "id": response["channel"],
         "timestamp": response["ts"],
-        "weblink": f"https://slack.com/app_redirect?channel={response['id']}",  # TODO should we fetch the permalink?
+        "weblink": f"https://{domain}.slack.com/app_redirect?channel={response['id']}",  # TODO should we fetch the permalink?
```
mvilanova, line 329: Same as `send_message`.

src/dispatch/static/dispatch/src/incident/ReportSubmissionCard.vue
```diff
@@ -216,6 +217,24 @@ export default {
...
+    if (this.query.tag) {
+      if (Array.isArray(this.query.tag)) {
+        this.tags = this.query.tag.map(function (t) {
+          return { name: t }
+        })
+      } else {
+        this.tags = [{ name: this.query.tag }]
+      }
+    }
+
```
mvilanova, line 237: Forgot to rebase the changes in master?

mvilanova (changes requested):

src/dispatch/plugins/dispatch_slack/service.py
metroid-samus: Done. Using chat.getPermalink instead.

src/dispatch/static/dispatch/src/incident/ReportSubmissionCard.vue
metroid-samus: Fixed.

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -281,13 +283,20 @@ def send_message(
         blocks=blocks,
     )
 
+    weblink = make_call(
+        client,
+        SlackAPIGetEndpoints.chat_permalink,
+        channel=response["channel"],
+        message_ts=response["ts"],
+    )
```
mvilanova, line 291: I'd put this in a function you can call, so you don't have the same exact code in two places.

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -185,10 +185,12 @@ def create_conversation(client: WebClient, name: str, is_private: bool = False)
         is_private=is_private,
     )["channel"]
 
+    domain = make_call(client, SlackAPIGetEndpoints.team_info)["team"]["domain"]
```
mvilanova, line 188: Same with this one if we needed to make the call in more than one place. Might be fine to leave it as is for now.

metroid-samus: Acknowledged. I'll keep it in for now.

src/dispatch/plugins/dispatch_slack/service.py
metroid-samus: Done.

src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -287,7 +298,7 @@ def send_message(
     return {
         "id": response["channel"],
         "timestamp": response["ts"],
-        "weblink": f"https://slack.com/app_redirect?channel={response['id']}",  # TODO should we fetch the permalink?
+        "weblink": get_message_permalink(client, response["channel"], response["ts"]),
```
wssheldon, line 301: it'd be good to cache `get_message_permalink` w/  the channed id if we'll call it on each `send_message`

mvilanova: @wssheldon I may be missing something, but why would we want to cache here, if each message will have a different permalink?

We could also not do the API call and include the permalink in the response, because we're not doing anything with it at the moment.

wssheldon: Ah yes, you're right. I'm surprised `chat_post_message` doesn't just return the permalink in the response. We should be able cache the domain we fetch on conversation creation though right?

`domain = make_call(client, SlackAPIGetEndpoints.team_info)["team"]["domain"]`

mvilanova: Yeah, that one we could cache.

metroid-samus: Done.

mvilanova (approved):

PR #3459, opened 2023-06-02
Fixes side bar loading

kevgliss:


wssheldon:
Will the filter for 5 per page in `getSignalData` be effectively the same as if we did `LIMIT 5` ? Just want to make sure we don't fetch every signal instance on component creation if we only need to show 5 of them.

kevgliss:
That's right, it eventually turns into a limit clause in the query.

PR #3462, opened 2023-06-02
Adds entity type management UI

kevgliss:
This is the initial work on creating a better management story for `EntityTypes` it adds a new section in the settings menu that allows for entity type-focused management (as opposed to signal definition focused). 

It also introduces the concept of an entity type 'scope'. The intent here is mostly around management, as the underlying logic does change much. Entities are either associated with a signal definition or they are not.

What defining a scope allows for is restricting the search space for assigned entities. The scopes roughly break down as:
 - `single`  -- Will only ever be associated with one definition and will not be a candidate to be associated with any other definition
 - `multiple` -- Can be assigned to one or more definitions, useful for entity types that may exist across some definitions
 - `all` -- This is a special case that will implicitly attach itself to all definitions (current and future) within the project. These are most useful for super common extraction that you'd like to do no matter what.
 
 
Screenshot 2023-06-02 at 4 06 04 PM


wssheldon:
How do you feel about the edit sheet being a modal instead of a drawer?

wssheldon:
Also, looks like you got the language server for JSON working? What was the fix?

kevgliss:
> How do you feel about the edit sheet being a modal instead of a drawer?

I thought about this, but it would be different than every other interaction new/edit component we had... so I wasn't sure it was really worth it. It is also slightly different enough and the existing modal/dialog that it wasn't going to be a drop-in replacement either.

src/dispatch/static/dispatch/src/entity_type/NewEditSheet.vue
```diff
@@ -22,164 +22,292 @@
...
-                >
-                </v-checkbox>
-              </v-flex>
-            </v-layout>
-          </v-container>
+          <ValidationProvider name="Name" rules="required" immediate>
+            <v-text-field
+              v-model="name"
+              label="Name"
+              hint="A name for your saved search."
```
wssheldon, line 34:
```suggestion
              hint="A name for your entity type."
```

src/dispatch/signal/flows.py
```diff
@@ -33,12 +34,15 @@ def signal_instance_create_flow(
     signal_instance = signal_service.get_signal_instance(
         db_session=db_session, signal_instance_id=signal_instance_id
     )
+    # fetch `all` entities that should be associated with all signal definitions
+    entity_types = entity_type_service.get_all(db_session=db_session, scope="all").all()
```
wssheldon, line 38:
```suggestion
    entity_types = entity_type_service.get_all(db_session=db_session, scope= EntityScopeEnum.all).all()
```

Could use the enum here

wssheldon (approved):

PR #3478, opened 2023-06-07
workflow related fixes

sara4dev:
Fix workflow params to parse the keys correctly
Send few incident metadata as part of the workflow params payload
Save the workflow run reasons

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/workflow.py
+++ b/src/dispatch/plugins/dispatch_slack/workflow.py
@@ -215,7 +215,7 @@ def handle_workflow_submission_event(
     named_params = []
     for i in form_data.keys():
         if i.startswith(RunWorkflowBlockIds.param_input):
-            key = i.split("-")[1]
+            key = i.split(RunWorkflowBlockIds.param_input + "-")[1]
             value = form_data[i]
             params.update({key: value})
             named_params.append({"key": key, "value": value})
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_slack/workflow.py
```diff
@@ -215,7 +215,7 @@ def handle_workflow_submission_event(
     named_params = []
     for i in form_data.keys():
         if i.startswith(RunWorkflowBlockIds.param_input):
-            key = i.split("-")[1]
+            key = i.split(RunWorkflowBlockIds.param_input + "-")[1]
```
sara4dev, line 218: without this fix, the workflow message is sent in the key called "workflow"
After this fix, the workflow message is sent in the key called "message"

src/dispatch/plugins/dispatch_slack/workflow.py
```diff
@@ -243,6 +243,10 @@ def handle_workflow_submission_event(
         {
             "externalRef": f"{DISPATCH_UI_URL}/{instance.incident.project.organization.name}/incidents/{instance.incident.name}?project={instance.incident.project.name}",
             "workflowInstanceId": instance.id,
+            "incident_name": instance.incident.name,
```
sara4dev, line 246: Having these metadata in payload is very handy, without making any extra api calls back to dispatch.

src/dispatch/workflow/service.py
```diff
@@ -154,6 +154,9 @@ def create_instance(
         incident = incident_service.get(db_session=db_session, incident_id=instance_in.incident.id)
         instance.incident = incident
 
+    if instance_in.run_reason:
```
sara4dev, line 157: workflow "run reason" was not saved in the DB before.

wssheldon:
@sara4dev Thanks for the contribution, this looks good. Out of curiosity, what are you using for your workflow plugin?

sara4dev:
@wssheldon - we integrated with the statuspage.io to post status page updates, by invoking `dispatch-run-workflows` command from the slack.

PR #3486, opened 2023-06-08
Always make sure there is a creator

kevgliss:


wssheldon (approved):
There *has* to be a creator for these i'm assuming

PR #3496, opened 2023-06-13
Addresses issues with cloud support in Jira plugin 

mvilanova:


src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -76,20 +76,23 @@ def get_user_field(client: JIRA, hosting_type: str, jira_username: str, user_ema
...
-                startAt=0,
-                maxResults=1,
-                params={"query": username},
-            )
-        )
-        return {"id": user.accountId}
+        return {"name": configuration.jira_username}
+    if configuration.hosting_type == "cloud":
+        import json
+        import requests
```
mvilanova, line 82: Using `requests` to workaround compatibility issues between Python client version and hosting types.

src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -76,20 +76,23 @@ def get_user_field(client: JIRA, hosting_type: str, jira_username: str, user_ema
...
+        from requests.auth import HTTPBasicAuth
+
+        endpoint = "groupuserpicker"
```
mvilanova, line 85: Golden endpoint to resolve user emails to ids.

src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -258,24 +258,27 @@ def create(
...
 
         project_id, issue_type_name = process_plugin_metadata(incident_type_plugin_metadata)
 
         if not project_id:
             project_id = self.configuration.default_project_id
 
+        # NOTE: to support issue creation by project id or key
+        project = {"id": project_id}
+        if not isinstance(project_id, int):
+            project = {"key": project_id}
```
mvilanova, line 272: This is to support both Jira project ids and keys in the configuration fields, as on-prem ids do not translate to cloud and it's easier to work with keys.

PR #3502, opened 2023-06-15
Adds a color emjoii to case priority

kevgliss:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -41,12 +41,22 @@
 
 
 def create_case_message(case: Case, channel_id: str) -> list[Block]:
+    # TODO we should probably restrict the possible colors to make this work
```
wssheldon, line 44: Or the names of the priorities and map those to colors.

kevgliss: The names are user-defined. So they could be anything :-/.

PR #3504, opened 2023-06-16
Fixes snoozing issue when signals have no entities

kevgliss:


src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -402,21 +402,28 @@ def snooze_button_click(
...
+        optional=True,
+    )
+
+    if entity_select_block:
+        blocks.append(entity_select_block)
+        blocks.append(
+            Context(
+                elements=[
+                    MarkdownText(
+                        text="Signal's that contain all selected entities will be snoozed for the configured timeframe."
```
whitdog47, line 421: No apostrophe here. Should be just "Signals"

kevgliss: For small changes feel free to use github's "Suggestion" feature.
```suggestion
                        text="Signals that contain all selected entities will be snoozed for the configured timeframe."
```

PR #3511, opened 2023-06-21
Adding tooltip with actual datetime to timeline tabs

whitdog47:


src/dispatch/static/dispatch/src/incident/TimelineTab.vue
```diff
@@ -76,6 +73,26 @@ export default {
       Util.exportCSV(items, this.name + "-timeline-export.csv")
       this.exportLoading = false
     },
+    formatToTimeZones(date) {
```
kevgliss, line 76: If we think that this will be used elsewhere we could also make this into a filter: https://github.com/Netflix/dispatch/blob/167442773ad7925792c27c2ea4ced59d67154cae/src/dispatch/static/dispatch/src/filters.js#L4

src/dispatch/static/dispatch/src/case/TimelineTab.vue
```diff
@@ -77,6 +74,26 @@ export default {
...
+          .format("YYYY-MM-DD HH:mm:ss")}`
+    },
   },
 }
 </script>
+
+<style scoped>
+.wavy-underline {
+  text-decoration: underline;
+  text-decoration-style: wavy;
```
wssheldon, line 94: I think this will look better, may need to up the darkness of the gray. I planned on changing this in the CasePage component as well. Let me know what you think.

```suggestion
  text-decoration-style: dotted;
```

src/dispatch/static/dispatch/src/case/TimelineTab.vue
```diff
@@ -2,23 +2,12 @@
   <v-container>
     <v-row justify="end">
       <v-switch v-model="showDetails" label="Show details" />
-      <v-btn
-        color="secondary"
-        class="ml-2 mr-2 mt-3"
-        @click="exportToCSV()"
-        :loading="exportLoading"
-      >
+      <v-btn color="secondary" class="ml-2 mr-2 mt-3" @click="exportToCSV()" :loading="exportLoading">
```
mvilanova, line 5: Are you using Prettier and ESLint with your IDE? I think one-liners are harder to read and maintain.

src/dispatch/static/dispatch/src/case/TimelineTab.vue
```diff
@@ -51,6 +47,7 @@
 <script>
 import { mapFields } from "vuex-map-fields"
 import Util from "@/util"
+import moment from "moment-timezone"
```
mvilanova, line 50: I don't think this import is needed anymore after switching to using a filter.

src/dispatch/static/dispatch/src/incident/TimelineTab.vue
```diff
@@ -51,6 +47,7 @@
 <script>
 import { mapFields } from "vuex-map-fields"
 import Util from "@/util"
+import moment from "moment-timezone"
```
mvilanova, line 50: Idem for this one.

src/dispatch/static/dispatch/src/case/TimelineTab.vue
```diff
@@ -80,3 +77,17 @@ export default {
...
+  text-decoration-style: wavy;
+  text-decoration-color: lightgray;
+  text-decoration-thickness: 1px;
+  text-underline-offset: 3px;
+}
+
+.pre-formatted {
+  white-space: pre;
+}
+</style>
```
mvilanova, line 103: I wonder if we could move this to a styles file we can import from so we don't have the same code in two places.

mvilanova (approved):
Nice work!

PR #3533, opened 2023-06-27
Deduplicates participants lists before they're used for adding users to conversations or threads

mvilanova:
Screenshot 2023-06-27 at 3 18 58 PM

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/plugin.py
+++ b/src/dispatch/plugins/dispatch_slack/plugin.py
@@ -238,7 +238,7 @@ class SlackConversationPlugin(ConversationPlugin):
     def add(self, conversation_id: str, participants: List[str]):
         """Adds users to conversation."""
         client = create_slack_client(self.configuration)
-        participants = [resolve_user(client, p)["id"] for p in participants]
+        participants = [resolve_user(client, p)["id"] for p in set(participants)]
 
         archived = conversation_archived(client, conversation_id)
         if not archived:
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_slack/service.py
```diff
@@ -246,7 +246,11 @@ def add_users_to_conversation_thread(
     client: WebClient, conversation_id: str, thread_id, user_ids: List[str]
 ) -> NoReturn:
     """Adds user to a threaded conversation."""
-    users = [f"<@{user_id}>" for user_id in user_ids]
```
wssheldon, line 249: If we don't care about the ordering can we call `set()` here to dedupe instead?

```python
users = [f"<@{user_id}>" for user_id in set(user_ids)]
```

mvilanova: Yes, I can use `set`. I just wanted to make it more explicit. I've actually id'ed other places where we want to dedupe as well.

src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -247,7 +247,7 @@ def add(self, conversation_id: str, participants: List[str]):
     def add_to_thread(self, conversation_id: str, thread_id: str, participants: List[str]):
         """Adds users to a thread conversation."""
         client = create_slack_client(self.configuration)
-        participants = [resolve_user(client, p)["id"] for p in participants]
+        participants = [resolve_user(client, p)["id"] for p in set(participants)]
```
mvilanova, line 250: Adding `set` here will save API calls to Slack.

PR #3535, opened 2023-06-29
Adding original values to timeline details for incidents and cases

whitdog47:


src/dispatch/case/service.py
```diff
@@ -169,6 +169,14 @@ def create(*, db_session, case_in: CaseCreate, current_user: DispatchUser = None
         db_session=db_session,
         source="Dispatch Core App",
         description="Case created",
+        details={
+            "title": case.title,
+            "description": case.description,
+            "type": case.case_type.name,
+            "severity": case.case_severity.name,
+            "priority": case.case_priority.name,
+            "visibility": str(case.visibility),
```
mvilanova, line 178: Hmm you shouldn't need to cast it to `str`, it should already be a string.

src/dispatch/incident/service.py
```diff
@@ -197,6 +197,14 @@ def create(*, db_session, incident_in: IncidentCreate) -> Incident:
         db_session=db_session,
         source="Dispatch Core App",
         description="Incident created",
+        details={
```
mvilanova, line 200: Both incidents and cases can be created with specific statuses via the Web UI. I wonder if should also capture the original status when the incident or case was created with.

mvilanova (approved):
LGTM

PR #3539, opened 2023-06-29
Send back a 409 for duplicate UUIDs

wssheldon:


src/dispatch/signal/views.py
```diff
@@ -104,12 +105,24 @@ def create_signal_instance(
...
-    signal_instance.signal = signal
-    db_session.commit()
-    return signal_instance
+    try:
+        signal_instance = signal_service.create_instance(
+            db_session=db_session, signal_instance_in=signal_instance_in
+        )
+    except IntegrityError as e:
+        if type(e.orig) is UniqueViolation:
+            msg = f"The id: {signal_instance_in.id} already exists in the database. Please provide a unique id."
```
mvilanova, line 114:
```suggestion
            msg = f"Signal instance id {signal_instance_in.id} already exists in the database. Please provide a unique id."
```

mvilanova: Will the info in this message be enough to troubleshoot on the other side?

mvilanova (approved):

src/dispatch/signal/views.py
wssheldon: Probably, it could be improved though. I'll add some additional information.

wssheldon:
Abandoning this PR in favor of #3564

PR #3545, opened 2023-07-03
Add's reporter to Case model and displays in UI

wssheldon:
Screenshot 2023-07-03 at 11 22 10 AM


src/dispatch/static/dispatch/src/case/ParticipantSelect.vue
```diff
@@ -1,36 +1,49 @@
 <template>
```
wssheldon, line 1: There's no change on this file it's just eslint formatting it

mvilanova (approved):

PR #3554, opened 2023-07-05
Autofill the tactical report's action section with opened tasks

metroid-samus:
This PR automatically adds the list of Open tasks to the "Actions" section in the tactical report, both through the web UI and on the Slack form.


src/dispatch/static/dispatch/src/incident/store.js
```diff
@@ -205,6 +205,13 @@ const actions = {
   showReportDialog({ commit }, incident) {
     commit("SET_DIALOG_REPORT", true)
     commit("SET_SELECTED", incident)
+
+    state.report.tactical.actions = incident.tasks.reduce((result, task) => {
+      if (task.status == "Resolved") {
+        return result
+      }
+      return (result ? result + "\n" : result) + "- " + task.description
```
whitdog47, line 213:
```suggestion
      return (result ? result + "\n" : "") + "- " + task.description
```

src/dispatch/static/dispatch/src/incident/store.js
```diff
@@ -205,6 +205,13 @@ const actions = {
   showReportDialog({ commit }, incident) {
     commit("SET_DIALOG_REPORT", true)
     commit("SET_SELECTED", incident)
+
+    state.report.tactical.actions = incident.tasks.reduce((result, task) => {
+      if (task.status == "Resolved") {
+        return result
+      }
+      return (result ? result + "\n" : "") + "- " + task.description
+    }, "")
```
mvilanova, line 214: Should we have this logic in `ReportDialog.vue` instead? We should load tasks in the actions text field in the Slack modal as well in `handle_report_tactical_command`.

src/dispatch/incident/models.py
```diff
@@ -300,6 +300,7 @@ class IncidentReadMinimal(IncidentBase):
     reporters_location: Optional[str]
     stable_at: Optional[datetime] = None
     tags: Optional[List[TagRead]] = []
+    tasks: Optional[List[TaskRead]] = []
```
mvilanova, line 303: We should define a new `TaskReadMinimal` Pydantic model with the minimum fields we need and use that instead. Otherwise, loading all the fields in `TaskRead` will degrade performance.

metroid-samus: Done.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1490,6 +1490,11 @@ def handle_report_tactical_command(
         conditions = tactical_report.details.get("conditions")
         actions = tactical_report.details.get("actions")
         needs = tactical_report.details.get("needs")
+    else:
```
mvilanova, line 1493: I think we should add open status tasks to the actions field regardless of whether we have a tactical report or not correct?

mvilanova: Also, we should make use of the `TaskStatus` enum `TaskStatus.resolved` instead of hardcoding `Resolved`

whitdog47 (approved):

PR #3564, opened 2023-07-10
Allows for signals to be updated

kevgliss:
This adds very basic signal update functionality (essentially making sure we are updating entities). Further improvements that could be made:

- A case timeline notification that a signal was updated
- A message in the case thread that a signal was updated
- Reopen a case if a signal is updated

I've left these off for simplicity, but also because I think they could very easily become confusing to whoever is triaging. I think this step allows us to capture the changed data while we think through what (if any) other behaviors we want to tie to a signal update.

### DIFF
--- a/src/dispatch/signal/flows.py
+++ b/src/dispatch/signal/flows.py
@@ -26,6 +26,32 @@ from dispatch.entity_type.models import EntityScopeEnum
 log = logging.getLogger(__name__)
 
 
+def signal_instance_update_flow(
+    signal_instance_id: int,
+    db_session: Session = None,
+):
+    """Create flow used by the API."""
+    signal_instance = signal_service.get_signal_instance(
+        db_session=db_session, signal_instance_id=signal_instance_id
+    )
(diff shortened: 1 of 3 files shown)


wssheldon (approved):

wssheldon:
@kevgliss We previously wanted to send back a 409 for duplicates, re: https://github.com/Netflix/dispatch/pull/3539/files

Should we abandon that approach in favor of this? I don't see how we can update signals if we implement the logic in #3539

kevgliss:
> @kevgliss We previously wanted to send back a 409 for duplicates, re: https://github.com/Netflix/dispatch/pull/3539/files
> 
> Should we abandon that approach in favor of the update logic in this pr (#3564)? I don't see how we can update signals if we implement the logic in #3539

Yes, I think the 429 PR was a smaller change so that was preferred if we needed immediate relief, but this PR is probably a better long-term approach. I do imagine there are probably multiple scenarios to allow the updating of signal data (e.g. GuardDuty, Uptycs) so we should probably support it (even if support is just minimal).

PR #3584, opened 2023-07-13
Fixes a bug with [] as a filter expression

kevgliss:


wssheldon:
cc @mvilanova -- Do you remember why the `[0]` list access was added (here's the pr #2052) ? I'm not familiar with the filter code at all and want to make sure nothing breaks here.

mvilanova:
> cc @mvilanova -- Do you remember why the `[0]` list access was added (here's the pr #2052) ? I'm not familiar with the filter code at all and want to make sure nothing breaks here.

I don't remember OTOMH. It's been a while since I cut that PR. I'll take a look today.

src/dispatch/database/service.py
```diff
@@ -355,7 +356,7 @@ def apply_filter_specific_joins(model: Base, filter_spec: dict, query: orm.query
         (SignalInstance, "EntityType"): (SignalInstance.entities, True),
     }
     filters = build_filters(filter_spec)
-    filter_models = get_named_models(filters)[0]
+    filter_models = get_named_models(filters)
```
mvilanova, line 359: @kevgliss @wssheldon IIRC it was a product of switching to using `SortedSet()` in ` get_named_models()` to respect the order in which we wanted the JOIN statements to happen. We could refactor `get_named_models()` so we don't have to do `[0]`

kevgliss: I think that fix is in place, we simply extend the `file_models` instead of creating a nested array (which we had previously just taken the first value of). 

I have tested incidents filtering with the change in place. lmk if you can think other areas that might need to be tested.

mvilanova (approved):

PR #3593, opened 2023-07-14
Do not raise an exception when the user cannot be found in MFA provider

wssheldon:


wssheldon:
@kevgliss I refactored this code to not use the private functions anymore, in case you want to take another look.

PR #3595, opened 2023-07-14
Adds new incident participant permission

whitdog47:
Allows incident participants to view the details of an incident even when the visibility of the incident is marked as restricted.

### DIFF
--- a/src/dispatch/auth/permissions.py
+++ b/src/dispatch/auth/permissions.py
@@ -342,6 +342,9 @@ class IncidentParticipantPermission(BasePermission):
         current_incident: Incident = incident_service.get(
             db_session=request.state.db, incident_id=pk.id
         )
+        if not current_incident:
+            return False
+
         participant_emails: list[str] = [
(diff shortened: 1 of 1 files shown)


whitdog47:
@mvilanova - this is modeled after the existing `CaseParticipantPermission` class in the same file.

src/dispatch/auth/permissions.py
```diff
@@ -330,6 +332,22 @@ def has_required_permissions(
...
+class IncidentParticipantPermission(BasePermission):
+    def has_required_permissions(
+        self,
+        request: Request,
+    ) -> bool:
+        current_user = get_current_user(request=request)
+        pk = PrimaryKeyModel(id=request.path_params["incident_id"])
+        current_incident: Incident = incident_service.get(
+            db_session=request.state.db, incident_id=pk.id
+        )
```
mvilanova, line 344: Can we check if `current_incident` exists?

mvilanova (approved):

PR #3597, opened 2023-07-14
lower all case priorities to standardize

wssheldon:
Screenshot 2023-07-14 at 10 18 35 AM

Case pri color appears to default to caps when setting in UI, somehow mine was lowered, and it appears to be the same in production. Standardizing to lower in the `create_case_message`.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/case/messages.py
+++ b/src/dispatch/plugins/dispatch_slack/case/messages.py
@@ -43,7 +43,7 @@ from dispatch.signal.enums import SignalEngagementStatus
 def create_case_message(case: Case, channel_id: str) -> list[Block]:
     # TODO we should probably restrict the possible colors to make this work
     priority_color_mapping = {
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -43,14 +43,14 @@
 def create_case_message(case: Case, channel_id: str) -> list[Block]:
     # TODO we should probably restrict the possible colors to make this work
     priority_color_mapping = {
-        "9E9E9E": "⚪",
-        "#8BC34A": "🟢",
-        "#FFEB3B": "🟡",
-        "#FF9800": "🟠",
-        "#F44336": "🔴",
-        "#9C27B0": "🟣",
+        "9e9e9e": "⚪",
```
wssheldon, line 46:
```suggestion
        "#9e9e9e": "⚪",
```

mvilanova (approved):

PR #3633, opened 2023-07-28
Allows users to conduct global search via the URL

whitdog47:
This PR adds the global search query to the URL, providing a way to copy & paste the search query, and allowing direct access to search results via URL. For instance, `https://dispatch.com/org/search/results?q=281` will show the search results for the query "281".

### DIFF
--- a/src/dispatch/static/dispatch/src/components/AppToolbar.vue
+++ b/src/dispatch/static/dispatch/src/components/AppToolbar.vue
@@ -182,9 +182,8 @@ export default {
     },
     performSearch() {
       let query = this.query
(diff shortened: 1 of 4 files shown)


kevgliss:
Thoughts on using a query param here? When I think of a search query I typically think of something like `/search?q=foo` mostly because I'd expect the result to be dynamic. Where as if it's directly in the URL id expect the results to be the same every time.

whitdog47:
Thanks for the suggestion, @kevgliss. I changed to a query parameter and improved further by updating the URL bar and search results when user types in the inputbox.

kevgliss (approved):

PR #3655, opened 2023-08-01
Adding ability to report incident from shortcut

whitdog47:
This PR adds a hook for a Report Incident Slack shortcut.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -1851,7 +1851,9 @@ def handle_update_incident_submission_event(
     )
 
 
-@app.shortcut(IncidentShortcutCallbacks.report, middleware=[db_middleware, shortcut_context_middleware])
+@app.shortcut(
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1849,6 +1851,54 @@ def handle_update_incident_submission_event(
...
+    initial_description = None
+    if body.get("message"):
+        permalink = (
+            client.chat_getPermalink(
+                channel=context["subject"].channel_id, message_ts=body["message"]["ts"]
+            )
+        )["permalink"]
+        initial_description = f"{body['message']['text']}\n\n{permalink}"
+
+    blocks = [
```
wssheldon, line 1873: Everything here and below is the same as in `handle_report_incident_command`, recommend moving into it's own function `draw_report_incident_modal` or similar to keep things consistent.

whitdog47: @wssheldon two differences are the `initial_value` for the `description_input` (line 1884) and the `trigger_id` (line 1901).

wssheldon (approved):

PR #3669, opened 2023-08-07
Putting a cap on the number of threads a scheduler can create

kevgliss:


wssheldon (approved):

src/dispatch/scheduler.py
```diff
@@ -7,28 +7,26 @@
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 .. moduleauthor:: Marc Vilanova <mvilanova@netflix.com>
 """
-import threading
 import logging
 import time
 
+from multiprocessing.pool import ThreadPool
```
forestmonster, line 13: Without getting too deep here, you prefer this over `concurrent.futures.ThreadPoolExecutor`?

src/dispatch/scheduler.py
```diff
@@ -7,28 +7,26 @@
...
-
 #  See: https://schedule.readthedocs.io/en/stable/ for documentation on job syntax
-class Scheduler(object):
+class Scheduler:
     """Simple scheduler class that holds all scheduled functions."""
 
     registered_tasks = []
     running = True
 
+    def __init__(self, num_workers=100):
```
forestmonster, line 27: Due to my lack of familiarity here... how does this not create a new bottleneck?

forestmonster (changes requested):
I'm curious about `concurrent.futures.ThreadPoolExecutor`, and if `num_workers=100` has the potential to create a new bottleneck?

forestmonster (approved):

PR #3699, opened 2023-08-17
Allows for filtering by participant and incident commander

whitdog47:
This PR adds the ability for the user to filter incidents by a given participant or a given incident commander. 

participant_filter


mvilanova:
Since the IC is just another participant, should we just keep it simple and filter by participant? This way we don't have to add checkboxes for scribe, observer, and liaison.

mvilanova (approved):
LGTM

PR #3711, opened 2023-08-21
Adds protections to the member_joined_channel Slack event handler

mvilanova:


whitdog47 (approved):
I like trying to retrieve the user from the participant table rather than depending on the email in the context to be valid.

PR #3721, opened 2023-08-23
Reverts unneeded changes to PagerDuty plugin

whitdog47:
This PR reverts some of the unneeded changes to the PagerDuty plugin which caused the `get()` to return a `dict` instead of the expected oncall email as a `str`.

### DIFF
--- a/src/dispatch/plugins/dispatch_pagerduty/plugin.py
+++ b/src/dispatch/plugins/dispatch_pagerduty/plugin.py
@@ -51,11 +51,11 @@ class PagerDutyOncallPlugin(OncallPlugin):
     def __init__(self):
         self.configuration_schema = PagerdutyConfiguration
 
-    def get(self, service_id: str, type: Literal["current", "next"] = "current", **kwargs) -> str:
-        """Gets the current or next oncall person's email."""
+    def get(self, service_id: str, **kwargs) -> str:
+        """Gets the current oncall person's email."""
         client = APISession(self.configuration.api_key.get_secret_value())
         client.url = self.configuration.pagerduty_api_url
-        return get_oncall(client=client, service_id=service_id, type=type)
+        return get_oncall(client=client, service_id=service_id)
 
     def page(
         self,
--- a/src/dispatch/plugins/dispatch_pagerduty/plugin.py
+++ b/src/dispatch/plugins/dispatch_pagerduty/plugin.py
@@ -14,7 +14,13 @@ from dispatch.decorators import apply, counter, timer
 from dispatch.plugins import dispatch_pagerduty as pagerduty_oncall_plugin
 from dispatch.plugins.bases import OncallPlugin
 
-from .service import get_oncall_email, page_oncall, oncall_shift_check, get_escalation_policy, get_service
+from .service import (
+    get_oncall_email,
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_pagerduty/plugin.py
```diff
@@ -51,11 +51,11 @@ class PagerDutyOncallPlugin(OncallPlugin):
     def __init__(self):
         self.configuration_schema = PagerdutyConfiguration
 
-    def get(self, service_id: str, type: Literal["current", "next"] = "current", **kwargs) -> str:
-        """Gets the current or next oncall person's email."""
+    def get(self, service_id: str, **kwargs) -> str:
```
mvilanova, line 54: We should double check but I think we can get rid of the `**kwargs` here

src/dispatch/plugins/dispatch_pagerduty/plugin.py
```diff
@@ -51,11 +51,11 @@ class PagerDutyOncallPlugin(OncallPlugin):
...
+        """Gets the current oncall person's email."""
         client = APISession(self.configuration.api_key.get_secret_value())
         client.url = self.configuration.pagerduty_api_url
-        return get_oncall(client=client, service_id=service_id, type=type)
+        return get_oncall(client=client, service_id=service_id)
```
mvilanova, line 58: Should we fold the `get_oncall` function into the `get_oncall_email` function in the service module and call `get_oncall_email` here instead?

src/dispatch/plugins/dispatch_pagerduty/service.py
```diff
@@ -79,11 +79,10 @@ def create_incident(client: APISession, headers: dict, data: dict) -> dict:
...
 
 
-def get_oncall_info(client: APISession, service: dict, type: Literal["current", "next"]) -> dict:
-    """Gets the current or next oncall info given a service."""
+def get_oncall_email(client, service: dict) -> str:
+    """Fetches the oncall's email for a given service."""
     escalation_policy_id = service["escalation_policy"]["id"]
-    escalation_policy = get_escalation_policy(client, escalation_policy_id)
-
+    escalation_policy = client.rget(f"/escalation_policies/{escalation_policy_id}")
```
mvilanova, line 85: Shouldn't we call `get_escalation_policy` instead, which has error handling?

src/dispatch/plugins/dispatch_pagerduty/service.py
```diff
@@ -95,44 +94,25 @@ def get_oncall_info(client: APISession, service: dict, type: Literal["current",
...
-            {
-                "name": user["name"],
-                "email": user["email"],
-                # "time_zone": user["time_zone"],
-                "start": oncall["start"],
-                "end": oncall["end"],
-            }
+            f"No users could be found for this pagerduty escalation policy ({escalation_policy_id}). Is there a schedule associated?"
         )
+    user = client.rget(f"/users/{user_id}")
```
mvilanova, line 107: Same here. Shouldn't we call `get_user` here?

mvilanova (changes requested):

mvilanova (approved):

PR #3726, opened 2023-08-25
Fixes bug in case escalate flow

mvilanova:


wssheldon (approved):

wssheldon:
Should these resources always be created on escalation? I'm assuming this bug is because we removed resource creation for signal instance cases.

mvilanova:
> Should these resources always be created on escalation? I'm assuming this bug is because we removed resource creation for signal instance cases.

Once we escalate the case, I don't think it makes sense to create a storage for it, as we can use the incident's one. Makes sense?

wssheldon:
> > Should these resources always be created on escalation? I'm assuming this bug is because we removed resource creation for signal instance cases.
> 
> Once we escalate the case, I don't think it makes sense to create a storage for it, as we can use the incident's one. Makes sense?

Makes sense, I forgot the resources aren't re-used for the incident.

PR #3734, opened 2023-08-29
Fixes issues with updating case reporter and assignee

mvilanova:


src/dispatch/case/flows.py
```diff
@@ -368,20 +368,28 @@ def case_update_flow(
     *,
     case_id: int,
     previous_case: CaseRead,
+    reporter_email: str,
     assignee_email: str,
-    user_email: str,
```
mvilanova, line 372: Removed it as it was not being used.

src/dispatch/case/flows.py
```diff
@@ -142,14 +142,14 @@ def case_add_or_reactivate_participant_flow(
             subject=case,
             group=case.tactical_group,
             group_action=GroupAction.add_member,
-            group_member=case.assignee.individual.email,
+            group_member=participant.individual.email,
```
mvilanova, line 145: We were always only adding the assignee.

src/dispatch/case/flows.py
```diff
@@ -403,22 +411,16 @@ def case_update_flow(
...
-                group_member=case.assignee.individual.email,
+                group_member=group_member,
                 db_session=db_session,
             )
-            event_service.log_case_event(
-                db_session=db_session,
-                source="Dispatch Core App",
-                description="Case group updated",
-                case_id=case_id,
-            )
```
mvilanova, line 421: Moved the logging to the case flows module.

metroid-samus (approved):

src/dispatch/group/flows.py
```diff
@@ -118,6 +120,21 @@ def update_group(
             log.exception(e)
             return
 
+        if subject_type == "case":
```
wssheldon, line 123: Unrelated to the pr, but maybe we could make a `log_subject_event` that takes a `subject_type` param to avoid the double function call where we emit an event.

src/dispatch/case/views.py
```diff
@@ -161,6 +161,22 @@ def update_case(
...
+        reporter_email = current_user.email
+
+    assignee_email = None
+    if case_in.assignee:
+        # we assign the case to the assignee provided
+        assignee_email = case_in.assignee.individual.email
+    elif current_user:
+        # we fall back to assign the case to the current user
+        assignee_email = current_user.email
+
```
wssheldon, line 179: I think it'd be a better UX if something can be assigned to no one, instead of defaulting to current user, as it could be confusing or not what someone is intending to do (un-assign something completely). 

Should we make reporter/assignee nullable?

src/dispatch/group/flows.py
mvilanova: Yeah, I think that's a great idea. There are too many places where we do the same now. We can tackle it in a separate PR.

src/dispatch/case/views.py
mvilanova: My thinking was that there shouldn't be any cases w/o a reporter or assignee, but in the event that we had any, then next time someone updated them, they would become the reporter and/ or assignee, so moving forward we would not have those fields empty. I can't think of any cases where we would like cases w/o a reporter and/ or assignee, as we loose context and accountability?

PR #3751, opened 2023-09-05
Allow a restricted incident priority for stable status

whitdog47:
This PR allows each project to select whether a certain incident priority should be selected for the "Stable" incident status. 

Screenshot 2023-09-04 at 3 10 12 PM

Screenshot 2023-09-04 at 3 10 48 PM


src/dispatch/project/models.py
```diff
@@ -65,10 +76,12 @@ class ProjectCreate(ProjectBase):
 
 class ProjectUpdate(ProjectBase):
     send_daily_reports: Optional[bool] = Field(True, nullable=True)
+    restrict_stable_to_id: Optional[int]
```
kevgliss, line 79: Why not just pass the whole object back? e.g. `IncidentPriorityRead` that's the pattern we've typically followed before.

src/dispatch/incident/models.py
```diff
@@ -239,6 +239,7 @@ class ProjectRead(DispatchBase):
     id: Optional[PrimaryKey]
     name: NameStr
     color: Optional[str]
+    restrict_stable_to: Optional[IncidentPriorityRead] = None
```
kevgliss, line 242: A minor nit, but I'm not sure `restrict_stable_to` (and it's other forms) really convey whats going on. To me, something like: `stable_priority` describes what it is `priority` and when it's set `stable`.

kevgliss (approved):

src/dispatch/project/models.py
whitdog47: Must be related to the quirkiness of the way I overcame the circular logic, but this was the only way I could get the UI to update that attribute.

src/dispatch/incident/models.py
whitdog47: Good suggestion. I like the way that reads better too. Fixed.

PR #3758, opened 2023-09-05
Record added and removed tags in timeline

whitdog47:
This PR records added and/or removed tags in the timeline. The description is just a summary of the number of added/removed tags and the details show the actual tags that were added/removed. 

Screenshot 2023-09-05 at 1 10 30 PM

### DIFF
--- a/src/dispatch/incident/flows.py
+++ b/src/dispatch/incident/flows.py
@@ -29,6 +29,7 @@ from dispatch.service import service as service_service
 from dispatch.storage import flows as storage_flows
 from dispatch.task.enums import TaskStatus
 from dispatch.ticket import flows as ticket_flows
+from dispatch.tag.flows import check_for_tag_change
 
 from .messaging import (
     # get_suggested_document_items,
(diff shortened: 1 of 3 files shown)


src/dispatch/incident/flows.py
```diff
@@ -437,6 +437,35 @@ def incident_closed_status_flow(incident: Incident, db_session=None):
     send_incident_rating_feedback_message(incident, db_session)
 
 
+def check_for_tag_change(
```
mvilanova, line 440: Should this logic live in tag/service.py or tag/flows.py?

src/dispatch/incident/flows.py
```diff
@@ -469,6 +498,19 @@ def conversation_topic_dispatcher(
...
+    )
+    if description:
+        event_service.log_incident_event(
+            db_session=db_session,
+            source="Incident Participant",
+            description=f"{individual.name} {description}",
+            details=details,
+            incident_id=incident.id,
+            individual_id=individual.id,
+        )
```
mvilanova, line 512: Should this logic live in the `incident_update_flow` function? This function is dedicated to determine if the conversation topic needs to be updated.

mvilanova (approved):

PR #3772, opened 2023-09-11
Adds initial canary support

kevgliss:


src/dispatch/signal/models.py
```diff
@@ -231,6 +231,7 @@ class SignalInstance(Base, TimeStampMixin, ProjectMixin):
     case_priority = relationship("CasePriority", backref="signal_instances")
```
captain-hoity-toity, line 231: This method might be better suited as a private utility.

PR #3774, opened 2023-09-12
Add support for SQLAlchemy pool pre ping

Cinojose:
Adding support for DB `pool_pre_ping ` to fix db disconnects. We've been getting quite a bit of `OperationalError ` lately and this seems to fix. 

We set the flag to True in env.

```bash
(psycopg2.OperationalError) SSL SYSCALL error: EOF detected
(Background on this error at: http://sqlalche.me/e/13/e3q8)
```
More reading 
https://docs.sqlalchemy.org/en/20/errors.html#operationalerror


whitdog47 (approved):
Thanks for the fix!

PR #3789, opened 2023-09-15
Rework the way slack api errors are propagated (to the caller)

kevgliss:
We really shouldn't be attempting to silence specific API errors in this function as it doesn't know if it's safe to discard them. Instead, we should propagate to the caller and let them decide how to handle the error. Afaict all the callers that care about `CHANNEL_NOT_FOUND`, `USER_NOT_IN_CHANNEL`, and `USERS_NOT_FOUND` already handle those errors themselves.


mvilanova (approved):

whitdog47 (approved):
Thanks for doing this.

PR #3791, opened 2023-09-19
Adds functionality to manually create missing incident and case resources.

metroid-samus:
A button appears on the incident or case resources tab if there are missing resources and if there exists an enabled plugin to create that resource. If there are no missing resources or there are no enabled plugins to create those resources, the button does not appear.

image

image


src/dispatch/static/dispatch/src/case/ResourcesTab.vue
```diff
@@ -65,23 +65,75 @@
...
+  async mounted() {
+    this.ticketPluginEnabled = await this.isPluginEnabled("ticket")
+    this.groupPluginEnabled = await this.isPluginEnabled("group")
+    this.conversationPluginEnabled =
+      (await this.isPluginEnabled("conversation")) &&
+      this.selected.case_type &&
+      this.selected.case_type.conversation_target
+    this.storagePluginEnabled = await this.isPluginEnabled("storage")
+    this.documentPluginEnabled = await this.isPluginEnabled("document")
+    console.log(this)
```
whitdog47, line 132: Would it be better to get all enabled plugins for this project once and then check the list to see if the type is on the list rather than making 5 separate API calls?

whitdog47, line 132:
```suggestion

```

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
```diff
@@ -80,5 +115,17 @@ export default {
       "selected.conversation",
     ]),
   },
+
+  async mounted() {
+    this.ticketPluginEnabled = await this.isPluginEnabled("ticket")
+    this.conferencePluginEnabled = await this.isPluginEnabled("conference")
+    this.conversationPluginEnabled = await this.isPluginEnabled("conversation")
+    this.storagePluginEnabled = await this.isPluginEnabled("storage")
+    this.documentPluginEnabled = await this.isPluginEnabled("document")
```
whitdog47, line 124: Same comment as above with cases.

src/dispatch/static/dispatch/src/case/ResourcesTab.vue
metroid-samus: I'm planning to add a feature that allows users to select which resource to add. 
For example, only creating a storage resource.

src/dispatch/incident/views.py
```diff
@@ -139,6 +140,21 @@ def create_incident(
...
+    response_model=IncidentUpdate,
+    summary="Creates resources for an existing incident.",
+)
+def create_incident_resources(
+    organization: OrganizationSlug,
+    incident_id: PrimaryKey,
+    current_incident: CurrentIncident,
+):
+    """Creates resources for an existing incident."""
+    incident_create_resources_flow(organization_slug=organization, incident_id=incident_id)
```
whitdog47, line 154: Should we do this as a background task with `background_tasks.add_task()`?

src/dispatch/case/views.py
```diff
@@ -145,6 +147,30 @@ def create_case(
     return case
 
 
+@router.post(
```
whitdog47, line 150: Should we do this as a background task with background_tasks.add_task()?

src/dispatch/incident/views.py
```diff
@@ -139,6 +140,21 @@ def create_incident(
     return incident
 
 
+@router.post(
+    "/{incident_id}/resources",
+    response_model=IncidentUpdate,
+    summary="Creates resources for an existing incident.",
+)
+def create_incident_resources(
```
whitdog47, line 148: Should we do this as a background task with background_tasks.add_task()?

src/dispatch/static/dispatch/src/case/ResourcesTab.vue
whitdog47: Still, one could retrieve the enabled plugin list once on load and check against the list as needed.

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
metroid-samus: I'm planning to add a feature that allows users to select which resource to add.
For example, only creating a storage resource.

src/dispatch/static/dispatch/src/case/ResourcesTab.vue
metroid-samus: Done.

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
metroid-samus: Done.

src/dispatch/incident/views.py
metroid-samus: Following up on this offline.

src/dispatch/incident/views.py
metroid-samus: ditto

src/dispatch/incident/views.py
metroid-samus: Done.

src/dispatch/incident/views.py
metroid-samus: Done.

src/dispatch/case/views.py
metroid-samus: Done.

src/dispatch/case/flows.py
```diff
@@ -174,6 +174,74 @@ def create_conversation(case: Case, conversation_target: str, db_session: Sessio
     return conversation
 
 
+def create_conversation_flow(
```
mvilanova, line 177: We have  a function for creating a conversation and one for adding participants to it in `conversation/flows.py`. We should refactor them to support case-specific conversations and delete `create_conversation_flow` and `create_conversation` above from this module, so we only have to maintain logic in one place.

src/dispatch/case/views.py
```diff
@@ -145,6 +147,32 @@ def create_case(
...
+    response_model=CaseRead,
+    summary="Creates resources for an existing case.",
+)
+def create_case_resources(
+    db_session: DbSession,
+    case_id: PrimaryKey,
+    current_case: CurrentCase,
+    background_tasks: BackgroundTasks,
+):
+    """Creates resources for an existing incident."""
```
mvilanova, line 161:
```suggestion
    """Creates resources for an existing case."""
```

src/dispatch/case/flows.py
```diff
@@ -684,7 +694,11 @@ def case_assign_role_flow(
 
 
 def case_create_resources_flow(
-    db_session: Session, case_id: int, individual_participants: list, team_participants: list
+    db_session: Session,
+    case_id: int,
+    individual_participants: list,
+    team_participants: list,
```
mvilanova, line 700: Typing should be ` List[str]`

src/dispatch/incident/flows.py
```diff
@@ -140,68 +140,74 @@ def inactivate_incident_participants(incident: Incident, db_session: Session):
...
+def incident_create_resources(*, incident: Incident, db_session=None) -> Incident:
+    """Creates all resources required for incidents."""
     # we create the incident ticket
-    ticket_flows.create_incident_ticket(incident=incident, db_session=db_session)
+    if not incident.ticket:
+        ticket_flows.create_incident_ticket(incident=incident, db_session=db_session)
 
     # we resolve individual and team participants
     individual_participants, team_participants = get_incident_participants(incident, db_session)
+    tactical_participant_emails = [i.email for i, _ in individual_participants]
```
mvilanova, line 151: nit: I'd move this line after the next `if` like you have it for the incident notifications group below

src/dispatch/incident/flows.py
```diff
@@ -265,6 +271,30 @@ def incident_create_flow(*, organization_slug: str, incident_id: int, db_session
...
         )
 
+    return incident
+
+
+@background_task
+def incident_create_resources_flow(
+    *, organization_slug: str, incident_id: int, db_session=None
+) -> Incident:
+    """Creates all resources required for new incidents and initiates incident response workflow."""
```
mvilanova, line 281: I think this docstring makes sense for the `incident_create_flow` function, but not this one. I'd change it to something like "Creates all resources for an existing incident" or something like that.

src/dispatch/incident/flows.py
```diff
@@ -265,6 +271,30 @@ def incident_create_flow(*, organization_slug: str, incident_id: int, db_session
...
+@background_task
+def incident_create_flow(*, organization_slug: str, incident_id: int, db_session=None) -> Incident:
+    """Creates all resources required for new incidents and initiates incident response workflow."""
+    # we get the incident
+    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+
+    # we create the incident resources
+    incident_create_resources(incident=incident, db_session=db_session)
+
     event_service.log_incident_event(
```
mvilanova, line 298: I wonder if this should live inside the new `incident_create_resources` function now.

mvilanova: Perhaps the `log_incident_event` for "Incident notifications sent" should live in `send_incident_created_notifications`

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
```diff
@@ -62,23 +62,83 @@
       </v-list-item>
       <v-divider />
     </span>
+    <span
+      v-if="
+        (!ticket && ticketPluginEnabled) ||
+        (!conference && conferencePluginEnabled) ||
+        (!conversation && conversationPluginEnabled) ||
+        (!storage && storagePluginEnabled) ||
+        (!documents && documentPluginEnabled)
```
mvilanova, line 71: I think we're missing groups in this component.

src/dispatch/static/dispatch/src/incident/store.js
```diff
@@ -347,6 +348,40 @@ const actions = {
...
+          commit("SET_SELECTED", response.data)
+          dispatch("getEnabledPlugins").then((enabledPlugins) => {
+            // Poll the server for resource creation updates.
+            var interval = setInterval(function () {
+              if (
+                state.selected.conversation ^ enabledPlugins.includes("conversation") ||
+                state.selected.documents ^ enabledPlugins.includes("document") ||
+                state.selected.storage ^ enabledPlugins.includes("storage") ||
+                state.selected.conference ^ enabledPlugins.includes("conference") ||
+                state.selected.ticket ^ enabledPlugins.includes("ticket")
```
mvilanova, line 365: groups is missing here too

mvilanova (dismissed):

src/dispatch/incident/flows.py
metroid-samus: I kept `tactical_participant_emails` here because the variable is used again for storage and conference creation.

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
metroid-samus: I chose not to include groups in this check since groups data was not rendered in this sheet.

src/dispatch/static/dispatch/src/incident/store.js
metroid-samus: ditto.

src/dispatch/incident/flows.py
metroid-samus: Done.

whitdog47 (approved):

src/dispatch/static/dispatch/src/incident/ResourcesTab.vue
metroid-samus: I'll resolve this in a separate PR.

src/dispatch/static/dispatch/src/incident/store.js
metroid-samus: I'll resolve this in a separate PR.

PR #3795, opened 2023-09-20
Timeline enhancements 

whitdog47:
This PR includes many enhancements to the incident timeline, including:

- Direct link to the timeline for an incident. Users can now browse to `.../{org_name}/incidents/{incident_name}/timeline` to open the incident view directly to the timeline.
- The incident details view pane is now expandable by a click and drag on the vertical divider line.
- Icons in the timeline bubbles quickly indicate the type of event.
- The incident timeline now has filtering to show/hide different kinds of events, with system events hidden by default.
- Users can now add custom events to the timeline directly in the UI by hovering between existing events in the timeline and click the "Add event" button.
- Custom events can be edited and/or deleted by the incident commander or reporter.
- Messages imported from Slack with the reaction event can now be edited and/or deleted by the incident commander or reporter.
<hr />

Screenshot 2023-09-20 at 9 27 25 AM
<hr />

Screenshot 2023-09-20 at 9 43 43 AM
<hr />

Screenshot 2023-09-20 at 9 42 30 AM
<hr />

Screenshot 2023-09-22 at 10 32 51 AM


src/dispatch/auth/permissions.py
```diff
@@ -332,6 +332,26 @@ def has_required_permissions(
...
+            return False
+
+        if current_incident.commander:
+            if current_incident.commander.individual.email == current_user.email:
+                return True
+
+        if current_incident.reporter:
+            if current_incident.reporter.individual.email == current_user.email:
+                return True
+
```
wssheldon, line 353:
```suggestion
if current_incident.commander and current_incident.commander.individual.email == current_user.email:
    return True

if current_incident.reporter and current_incident.reporter.individual.email == current_user.email:
    return True

return False
```

If none of the conditions meet the function will return `None` which I think is not expected.

src/dispatch/event/service.py
```diff
@@ -29,6 +30,11 @@ def get_by_incident_id(*, db_session, incident_id: int) -> List[Optional[Event]]
     return db_session.query(Event).filter(Event.incident_id == incident_id)
 
 
+def get_by_uuid(*, db_session, uuid: str) -> List[Optional[Event]]:
```
wssheldon, line 33: nit

```suggestion
def get_by_uuid(*, db_session, uuid: str) -> list[Event | None]:
```

py3.9+ has `list` type builtin now and `Optional` is deprecated for `| None`

src/dispatch/incident/views.py
```diff
@@ -300,6 +301,91 @@ def create_executive_report(
...
+    "/{incident_id}/event",
+    summary="Creates a custom event.",
+    dependencies=[Depends(PermissionsDependency([IncidentEditPermission]))],
+)
+def create_custom_event(
+    db_session: DbSession,
+    organization: OrganizationSlug,
+    incident_id: PrimaryKey,
+    current_incident: CurrentIncident,
+    event_in: dict,
```
wssheldon, line 314: Should we have a pydantic model for this? `EventCreate`

wssheldon:
One thing that stands out form a UX perspective in the timeline is user actions vs system actions look very similar. I think it'd help to differentiate them more. 

For example, the icon for the user actions could be the users profile picture and names could be bolded/clickable with the user popover component we have.

wssheldon:
A minor thing from last screenshot -- the delete / edit icons probably fit best positioned all the way right, under the date. that way they wont cover the text or interfere with the content if it is rich (hoverable/clickable) in the future

whitdog47:
> One thing that stands out form a UX perspective in the timeline is user actions vs system actions look very similar. I think it'd help to differentiate them more.
> 
> For example, the icon for the user actions could be the users profile picture and names could be bolded/clickable with the user popover component we have.

Will pull these ideas into a phase 2, which will align the entire Incident edit/view with the new Case UI.

jneelamekam (approved):
Looks good to me.

PR #3804, opened 2023-09-21
Adds null checks for document and case conversation flows.

metroid-samus:
Adds null checks for document and case conversation flows.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -176,9 +176,18 @@ def create_conversation(case: Case, conversation_target: str, db_session: Sessio
 
 def update_conversation(case: Case, db_session: SessionLocal):
     """Updates external communication conversation."""
+    if not case.conversation:
+        log.warning("Conversation not updated. No relationship between case and conversation.")
+        return
+
     plugin = plugin_service.get_active_instance(
         db_session=db_session, project_id=case.project.id, plugin_type="conversation"
     )
+
+    if not plugin:
+        log.warning("Document not updated. No conversation plugin enabled.")
+        return
+
     plugin.instance.update_thread(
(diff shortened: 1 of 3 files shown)


wssheldon:
Preferably we'd raise custom exceptions here instead of returning `None` for special cases. That way callers don't need to check the return value and can assume that if no exception is raised, the return value is valid.  If an exception is raised, it should be handled appropriately by the caller. Also, the `None` could be interpreted incorrectly in conditional statements.

Here's what it could look like:

```python
class ConversationUpdateException(Exception):
    """Base exception for all errors related to updating conversations."""
    pass

class NoConversationRelationError(ConversationUpdateException):
    def __init__(self, case_id):
        self.case_id = case_id  # custom property
        super().__init__(
            f"No relationship between case {case_id} and a conversation."
        )

def update_conversation(case: Case, db_session: SessionLocal) -> None:
    """Updates external communication conversation.
    
    Raises:
        NoConversationRelationError: If there is no relationship between the case and conversation.
        NoActivePluginError: If no conversation plugin is enabled.
    """
    if not case.conversation:
        log.exception("Conversation not updated. No relationship between case and conversation.")
        raise NoConversationRelationError("No relationship between case and conversation.")
    
    ....
```

src/dispatch/case/flows.py
```diff
@@ -176,9 +176,18 @@ def create_conversation(case: Case, conversation_target: str, db_session: Sessio
 
 def update_conversation(case: Case, db_session: SessionLocal):
```
wssheldon, line 177:
```suggestion
def update_conversation(case: Case, db_session: Session):
```

Maybe not in this PR but we should replace all of the `SessionLocal` with `Session` since `SessionLocal` just `Any`

metroid-samus: Ack.

PR #3811, opened 2023-09-26
Re-implement ParticipantSelect Component with v-autocomplete

wssheldon:
This allows searching by email and name. It is also faster in my experience.


src/dispatch/static/dispatch/src/incident/ParticipantSelect.vue
```diff
@@ -95,17 +96,20 @@ export default {
...
+      return name.indexOf(searchText) > -1 || email.indexOf(searchText) > -1
     },
     fetchData() {
       this.loading = "error"
       let filterOptions = {
         q: this.search,
         sortBy: ["name"],
         descending: [false],
-        itemsPerPage: this.numItems,
+        itemsPerPage: -1,
```
wssheldon, line 112: This is worth noting. We should check how many users are in production. It'll be faster if we can just cache them all in the browser so that's why I did this.

mvilanova:
There's also a `ParticipantSelect.vue` [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/static/dispatch/src/case/ParticipantSelect.vue). It would be great if we could combine the two if we can and move the resulting one to a new folder called `participant` and adjust imports.

wssheldon:
I will get this out first https://github.com/Netflix/dispatch/pull/3818 and add a flag here to stay safe

whitdog47 (approved):

src/dispatch/static/dispatch/src/incident/ParticipantSelect.vue
kevgliss: Unless we have some sort of global cache, e.g. in a vuex store. I think this might end up generating quite a bit of data downloading.

kevgliss:
What was the actual UX problem or annoyance? afict the current participant select already supports name and email filtering?
Screenshot 2023-10-02 at 8 19 29 PM

wssheldon:
> What was the actual UX problem or annoyance? afict the current participant select already supports name and email filtering? Screenshot 2023-10-02 at 8 19 29 PM

I noticed this as well when I tested earlier. I do think the autocomplete provides some niceties around finding values that aren't exact matches, so it's less strict on the search.

kevgliss:
Yeah, I think that's probably universal. I think search tends to err on stem matching when I guess we should just do %like% in most cases. I believe we have implemented it in a few places but we could circle back and do an _or(%like%) to all our search queries.

PR #3813, opened 2023-09-27
Process Signal SQS

kevgliss:
Adds a signal SQS plugin for signal consumption (in addition to the endpoint). Once completed, we can likely refactor the signal instance creation endpoint with the new `create_signal_instance` function. Existing functionality is left in place until this is merged and a refactoring can be done.

I have also explicitly left the `processing` command functionality in place. This could likely be refactored again as background tasks or within the signal consumption function itself.


src/dispatch/plugins/dispatch_aws/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+                            "sqs.signal.received", tags={"signalName": signal.name}
+                        )
+                        log.debug(f"Received signal: {signal}")
+                        entries.append(
+                            {"Id": message["MessageId"], "ReceiptHandle": message["ReceiptHandle"]}
+                        )
+                    except Exception as e:
+                        log.exception(e)
+
+            client.delete_message_batch(QueueUrl=sqs_queue_url, Entries=entries)
```
wssheldon, line 66: If there are no messages, entries will be undefined / `NameError`

wssheldon (approved):

PR #3846, opened 2023-10-05
Allows users to edit name and description of search filter and view membership

whitdog47:
This PR provides a view/edit dialog for search filters under Settings->Search filters that allows users to edit the name and description for the filter as well as view the current membership.

Screenshot 2023-10-05 at 5 01 10 PM


mvilanova:
How does it look? Mind adding a screenshot?

src/dispatch/static/dispatch/src/search/EditDialog.vue
```diff
@@ -0,0 +1,154 @@
...
+                  </ValidationProvider>
+                </v-flex>
+              </v-layout>
+              <div class="text-body-2 ml-1 mt-2">Individuals</div>
+              <v-divider class="mt-2" />
+              <v-list style="max-height: 500px" class="overflow-y-auto">
+                <template v-for="individual in individuals">
+                  <v-list-item :key="individual.id">
+                    <v-list-item-content>
+                      <v-list-item-title>{{ individual.name }}</v-list-item-title>
```
metroid-samus, line 70: Let's use components like `IncidentParticipant` instead of `individual.name` to increase clarity on which individual is subscribed. There may be multiple individuals with the same name.

whitdog47:
@mvilanova screenshot attached

src/dispatch/static/dispatch/src/search/EditDialog.vue
whitdog47: Good idea. Updated.

metroid-samus (approved):

mvilanova:
We should add support for editing the actual filter in the future.

mvilanova (approved):

PR #3855, opened 2023-10-06
Fix bug where entity types created via Modal don't appear in combobox

wssheldon:
The issue stemmed from us not returning entity types with the type `single`:

```vue
+ scope: ["multiple", "single"]
```

Which is the default type when created via the `+` -> Modal creation dialog (from within the definition edit sheet).


wssheldon:
@kevgliss FYI, in case this change doesn't track w/ what you expected for the `all/multiple/single` categories you added. This is just to resolve the immediate UX issue and I think it's okay for now. I need to think about it more.

mvilanova (approved):

PR #3857, opened 2023-10-09
Update to vue 3

KaelWD:
- vee-validate has been replaced with vuetify's built-in form validation
	- success-messages and immediate no longer exist, but this doesn't seem like it was working correctly to start with (hints hidden on required inputs). The current UX kinda sucks, I'd replace the disabled save button with on-demand validation at least.
- Date and time pickers are still in alpha, I've temporarily replaced them with native inputs as they'll most likely change pretty significantly.
- Material icons have been removed, now using mdi everywhere. Let me know if there's any missing icons, I probably didn't get them all.


wssheldon:
Some initial things stood out when I tested locally:

- Incident Drawer doesn't load for me
- I get recurring session resets (I have to re-login extremely fast) not sure if this is an artifact of my local build or something to do w/ these changes, though it only happens on this branch
- Charts in dashboard don't seem to work for me

kevgliss:
It also looks like it's failing to build:
```
npm ERR! Conflicting peer dependency: vuetify@3.3.21
npm ERR! node_modules/vuetify
npm ERR!   peer vuetify@"^3.0.0" from eslint-plugin-vuetify@2.0.5
npm ERR!   node_modules/eslint-plugin-vuetify
npm ERR!     dev eslint-plugin-vuetify@"^2.0.5" from the root project
```

whitdog47:
@KaelWD - commit [e09fb13](https://github.com/Netflix/dispatch/pull/3857/commits/e09fb138921ba63e502a2d1a8579b262c6cfc643) fixes the incident drawer by removing the slidable drawer border - is there an equivalent for Vue3?

A few other notes:

- Styling seems to be off. For instance, the font-family should be Roboto for most text.
- In the incident and case drawer's timeline, the v-cols don't seem to be honored.
- I can't seem to change the status of an incident using the "Status" selector in the Details tab of the incident drawer.
- The "Search Filters" setting screen under Project is broken and won't load (http://<root>/default/settings/projects/searchFilters?project=Test).
- When trying to Export incidents or cases from the table view, there's no scroll bar.
- Report Incident gives a 500 server error (but seems to record it in the database).
- All tables report the total count as "10" even if there are more (e.g., "1-10 of 10").
- Some latest UI functionality is missing (i.e., the "Oncall feedback" submenu under "Feedback").

KaelWD:
> Conflicting peer dependency

Fixed, I had to use a nightly build of vuetify for some fixes that hadn't been released yet. You can use  `--force` to bypass that, npm is dumb and doesn't resolve peers properly when they use an alias.

> Incident Drawer

`$refs.drawer.$el` isn't actually the element: https://github.com/vuetifyjs/vuetify/issues/16766#issuecomment-1675036625

> recurring session resets

How fast is "extremely fast"? I've had to log back on every day for both this branch and master.

> Charts in dashboard don't seem to work

Fixed.

> font-family should be Roboto

It is for me, do you have a specific example?

> timeline, the v-cols don't seem to be honored

Fixed.

> change the status of an incident using the "Status" selector

Works for me.

> Search Filters" setting screen under Project is broken

Fixed, this was due to a change on master.

> there's no scroll bar

Where should the scrollbar be? I don't see one on master either.

> Report Incident gives a 500 server error

I haven't touched the backend at all.

> All tables report the total count as "10"

Fixed.

> Some latest UI functionality is missing

I've merged master in again, we're going to have to figure something out while this is being finalised though as I can't keep reimplementing everything you're adding. There's probably more new changes with outdated syntax that I've missed.

mvilanova:
@KaelWD I did my best to resolve the conflicts in the `package-lock.json` file. It's possible I broke some stuff. Apologies in advance if that's the case.

mvilanova:
@KaelWD any suggestions on how to suppress [these](https://github.com/Netflix/dispatch/actions/runs/6712797651/job/18243107912) errors? I tried with the usual stuff (`eslint-disable-line`) with no success.

PR #3868, opened 2023-10-11
Prevents references to nonexistant converstion object.

metroid-samus:
Fixes bug where Dispatch may reference a non-existent conversation object.


wssheldon:
Since this PR touches the case creation hot path, could you locally test each case creation variation (via Slack, UI, and Signal Ingest) locally?

wssheldon:
I believe this PR resolve the issue, https://github.com/Netflix/dispatch/pull/4092/files

PR #3898, opened 2023-10-23
Enhances Signals table

mvilanova:
* Adds columns for Create Case and Canary values.
* Adds search support. Query string will be matched against signal definition name and description.


wssheldon (approved):
Thanks for adding this, I've wanted this exact thing a lot recently.

PR #3902, opened 2023-10-24
Fixing an issue where we aren't paging for critical

kevgliss:


kevgliss:
cc @mvilanova I had to remove the signalInstance vector code because it was causing the tests to fail. I don't believe you can create vectors based on column properties (or at least it's not straightforward). I think the better option is to join and then to a query search on the definition's vector.

PR #3917, opened 2023-11-01
Rollback session in case of exceptions

kevgliss:
Fixes an issue where if an exception is raised during ingestion it can invalidate the session and prevent further processing.


wssheldon:
I think if we used context manager version of the session it would do this by default, but this works as well

wssheldon (approved):

PR #3920, opened 2023-11-01
Adding padding for spacing on participants and resources tabs

whitdog47:
Fixing spacing on Resources and Participants tabs in both incidents and cases view/edit screens.


metroid-samus (commented):
There seems to be an error with v-stepper.

metroid-samus (commented):
There seems to be errors with v-stepper

PR #3944, opened 2023-11-06
Project autocomplete

kevgliss:
Moves the project to an autocomplete and uses composition api.


wssheldon:
It looks like there might be a bug where the Project Select autocomplete doesn't load the stored value. (e.g.

```
dispatch=# select project_id from dispatch_organization_default.case where id = 5310 limit 10;

 project_id
------------
          1

(1 row)
```

Screenshot 2023-11-06 at 9 52 50 AM

But the auto-complete loads empty.

wssheldon:
> It looks like there might be a bug where the Project Select autocomplete doesn't load the stored value. (e.g.
> 
> ```
> dispatch=# select project_id from dispatch_organization_default.case where id = 5310 limit 10;
> 
>  project_id
> ------------
>           1
> 
> (1 row)
> ```
> 
> Screenshot 2023-11-06 at 9 52 50 AM
> 
> But the auto-complete loads empty.

This appears to be fixed after recent changes? 🤷‍♀️

wssheldon (approved):

PR #3994, opened 2023-11-15
Declutters Slack case threads

mvilanova:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -214,7 +211,9 @@ def create_signal_messages(case_id: int, channel_id: str, db_session: Session) -
                 ),
             ]
         ),
-        Section(text=f"Total instances in this case: *{num_of_instances}*\n"),
+        Section(text="*Signal Instances*"),
+        Divider(),
+        Section(text=f"{num_of_instances} signal instances observed in this case."),
```
wssheldon, line 216:
```suggestion
        Section(text=f"{num_of_instances} alerts observed in this case."),
```

"Alerts" is more widely understood, in my opinion.

mvilanova: I already accepted it, but if we use alerts here, should we use it everywhere else too?

src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -214,7 +211,9 @@ def create_signal_messages(case_id: int, channel_id: str, db_session: Session) -
                 ),
             ]
         ),
-        Section(text=f"Total instances in this case: *{num_of_instances}*\n"),
+        Section(text="*Signal Instances*"),
```
wssheldon, line 214:
```suggestion
        Section(text="*Alerts*"),
```

src/dispatch/plugins/dispatch_slack/case/messages.py
wssheldon: In user facing UI, I would say so.

PR #3995, opened 2023-11-15
Uses entity type name in Slack snooze modal

mvilanova:


wssheldon (approved):
Assuming `str()` was redundant

mvilanova:
> Assuming `str()` was redundant

Yeah, the value was a string column in the model.

PR #4007, opened 2023-11-17
Handle invalid JSONPath with KeyError, Exceptions, and add test cases

ghost:
Bad `JSONPath` lookups, specifically ones that attempt to access a `list` index against a `dict` value, results in a `KeyError`

```bash
tests/entity/test_entity_service.py::test_get PASSED                                                                                                                                                                                                                                                                                                                                                        [  9%]
tests/entity/test_entity_service.py::test_get_all_by_signal PASSED                                                                                                                                                                                                                                                                                                                                          [ 18%]
tests/entity/test_entity_service.py::test_create PASSED                                                                                                                                                                                                                                                                                                                                                     [ 27%]
tests/entity/test_entity_service.py::test_update PASSED                                                                                                                                                                                                                                                                                                                                                     [ 36%]
tests/entity/test_entity_service.py::test_delete PASSED                                                                                                                                                                                                                                                                                                                                                     [ 45%]
tests/entity/test_entity_service.py::test_find_entities_with_field_and_regex PASSED                                                                                                                                                                                                                                                                                                                         [ 54%]
tests/entity/test_entity_service.py::test_find_entities_with_regex_only PASSED                                                                                                                                                                                                                                                                                                                              [ 63%]
tests/entity/test_entity_service.py::test_find_entities_with_field_only PASSED                                                                                                                                                                                                                                                                                                                              [ 72%]
tests/entity/test_entity_service.py::test_find_entities_with_no_regex_or_field PASSED                                                                                                                                                                                                                                                                                                                       [ 81%]
tests/entity/test_entity_service.py::test_find_entities_handles_key_error PASSED                                                                                                                                                                                                                                                                                                                            [ 90%]
tests/entity/test_entity_service.py::test_find_entities_multiple_entity_types PASSED
```

### DIFF
--- a/src/dispatch/entity/service.py
+++ b/src/dispatch/entity/service.py
@@ -365,7 +365,7 @@ def find_entities(
                                         project=signal_instance.project,
                                     )
                 except KeyError:
-                    log.exception(
+                    log.warning(
                         f"Unable to extract entity {str(jpath)} is not a valid JSONPath for Instance {signal_instance.id}."
                         f"A KeyError usually occurs when the JSONPath includes a list index lookup against a dictionary value."
                         f"  Example: dictionary[0].value"
--- a/src/dispatch/entity/service.py
+++ b/src/dispatch/entity/service.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import logging
 from typing import Generator, Optional, Sequence, Union, NewType, NamedTuple
 import re
 
@@ -14,6 +15,9 @@ from dispatch.entity.models import Entity, EntityCreate, EntityUpdate, EntityRea
 from dispatch.entity_type import service as entity_type_service
 from dispatch.entity_type.models import EntityType
 from dispatch.signal.models import Signal, SignalInstance
+
+
+log = logging.getLogger(__name__)
 
 
 def get(*, db_session: Session, entity_id: int) -> Optional[Entity]:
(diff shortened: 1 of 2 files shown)


wssheldon:
You can call `str` on the `JSONPath` object:

```bash
>>> from jsonpath_ng import jsonpath, parse
>>> jsonpath_expr = parse('foo[*].baz')
>>> str(jsonpath_expr)
'foo.[*].baz'
```

src/dispatch/entity/service.py
```diff
@@ -360,9 +364,18 @@ def _find_entities_by_regex_and_jsonpath_expression(
                                         entity_type=entity_type,
                                         project=signal_instance.project,
                                     )
-                except jsonpath_ng.JSONPathError:
-                    # field not found in signal_instance.raw
-                    pass
+                except KeyError:
+                    log.exception(
```
kevgliss, line 368: Should these be logged as exceptions or info? With global entities wouldn't this fire quite a bit, even though it's large expected? Looking for a path that doesn't exist?

wssheldon: I think it's an `exception`. If the key can't found be an empty list is returned.

wssheldon: Eh, I'll downgrade it to `warning`. I do think it's a failure case.

wssheldon: With the context, should it be `info`, or do you think `warning`?

kevgliss: I think an exception would indicate something needs to be fixed/handled if we don't expect to take action a warning / info is probably fine imo.

PR #4008, opened 2023-11-17
Do not send case traige reminders if there is no assignee.

metroid-samus:


src/dispatch/case/messaging.py
```diff
@@ -77,6 +77,10 @@ def send_case_triage_reminder(case: Case, db_session: SessionLocal):
         }
     ]
 
+    if not case.assignee:
```
wssheldon, line 80: nit, we can fail faster above `items`.

whitdog47 (approved):

PR #4012, opened 2023-11-20
Fixes signal instance deduplication check.

metroid-samus:
Deduplicating signal instances should only occur when those instances are associated with a case.

### DIFF
--- a/src/dispatch/signal/flows.py
+++ b/src/dispatch/signal/flows.py
@@ -291,6 +291,9 @@ def _should_update_signal_message(signal_instance: SignalInstance) -> bool:
     """
     global _last_nonupdated_signal_cache
 
+    if signal_instance.case_id:
+        return False
+
     case_id = str(signal_instance.case_id)
 
     if case_id not in _last_nonupdated_signal_cache:
--- a/src/dispatch/signal/flows.py
+++ b/src/dispatch/signal/flows.py
@@ -81,9 +81,9 @@ def signal_instance_create_flow(
         db_session=db_session,
         signal_instance=signal_instance,
     ):
-        # If the signal was deduplicated, we can assume a case exists,
-        # and we need to update the corresponding signal message
-        if _should_update_signal_message(signal_instance):
+        # If a case exists and the signal was deduplicated,
+        # we need to update the corresponding signal message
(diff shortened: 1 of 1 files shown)


wssheldon (commented):
I believe it should not be possible for a `SignalInstance` object without a corresponding `case_id` to enter this conditional block, so there is likely a deeper issue. 

Do you have more context on what Signal definition this instance was associated with? It looks like the root cause could stem from the snooze filter action case, where `True` is returned for filtered but no `case_id` is associated with the instance (in every other case, we do, which is where the assumption came from in the comment (But, it looks like that is not correct):


```python
            # an expression is not required for snoozing, if absent we snooze regardless of entity
            if f.expression:
                instances = query.filter(SignalInstance.id == signal_instance.id).all()

                if instances:
                    signal_instance.filter_action = SignalFilterAction.snooze
                    filtered = True
                    break
            else:
                signal_instance.filter_action = SignalFilterAction.snooze
                filtered = True
                break
```

The root cause may be in this bit above. Maybe we need to check the filter action before proceeding with the next `_should_update_message` conditional function. 

Am I missing something or does this track? I want to be confident we aren't burying the root cause on this one.

src/dispatch/signal/flows.py
```diff
@@ -291,6 +291,9 @@ def _should_update_signal_message(signal_instance: SignalInstance) -> bool:
     """
     global _last_nonupdated_signal_cache
 
+    if signal_instance.case_id:
```
wssheldon, line 294: Separately, this function should run for cases with a `case_id`, since it is responsible for updating the `Total instances in this case: X` text in the Case Slack thread. This will cause it to never update.

kevgliss:
I think my preference would be to fix the check here:

https://github.com/Netflix/dispatch/blob/master/src/dispatch/signal/flows.py#L86

The assumption that a case exists if a signal has been filtered is wrong (e.g. they can be filtered in other ways like snoozing). There are likely other actions we will likely want to take if a case exists, by moving the conditional higher in the stack we make it clear that updating a message requires a case and provide a reasonable place to add other case-related actions in the future.

src/dispatch/signal/flows.py
metroid-samus: Done.

kevgliss (approved):

PR #4044, opened 2023-11-29
Tries a different approach to signal instance table filtering

mvilanova:
Screenshot 2023-11-29 at 1 03 18 PM


src/dispatch/static/dispatch/src/signal/store.js
```diff
@@ -117,6 +117,14 @@ const actions = {
     return SignalApi.getAllInstances(params)
       .then((response) => {
         commit("SET_INSTANCE_TABLE_LOADING", false)
+
+        // We filter out instances based on the given query parameter provided
+        if (params.q) {
+          response.data.items = filter(response.data.items, function (item) {
+            return item.signal.name.includes(params.q) || item.signal.description.includes(params.q)
```
mvilanova, line 124: I only added signal name and description, but we could add more.

wssheldon (approved):

PR #4046, opened 2023-11-29
Setup `eslint` for TypeScript and resolve all errors

wssheldon:
You have to use `@typescript-eslint` and `@typescript-eslint/parser` plugins (which are NPM packages) to use TypeScript with `eslint`. 

https://stackoverflow.com/questions/70083042/eslint-parsing-error-unexpected-token


wssheldon:
`@` everyone for vis. Build should be clean at this point.

PR #4048, opened 2023-11-30
Set a saving indicator and save status to Case Page Header

ghost:
Screenshot 2023-11-29 at 5 40 03 PM

We update the time to the current time on any save and populate the value with `updated_at` from the database on page load. We could periodically refresh this value from db, but it's probably fine for now.


wssheldon:
There's probably a cleaner pattern for saving/updating the Case object. It's all over the place right now.

PR #4063, opened 2023-11-30
Renames Investigation Write-Up to Resolution and move button to top

wssheldon:
Screenshot 2023-11-30 at 2 47 08 PM


mvilanova:
Much better!

PR #4099, opened 2023-12-08
Do not try to render Slack Button with empty url in Signal message

wssheldon:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -204,12 +204,20 @@ def create_signal_messages(case_id: int, channel_id: str, db_session: Session) -
                     action_id=SignalNotificationActions.snooze,
                     value=button_metadata,
                 ),
+            ]
+            # The button URL must have at least one character.
+            # Otherwise, Slack will raise a Validation error.
+            # external_url is not a required field. If it's empty, an empty list is added,
```
mvilanova, line 210: Should we make it required in the UI instead, so we always have a link to a response plan for all detections?

wssheldon: Sorry for the delay, we could take that approach. I was worried about introducing friction. For example, creating new detections as part of an incident. 

I can make it a required field though if you feel strongly about it.

wssheldon: @mvilanova Going to merge this for now -- let me know and I can go the required route instead.

PR #4131, opened 2023-12-15
[Case] - Refactor table to use script setup v3 api

wssheldon:
In preparation for other improvements.


wssheldon:
This is ready for review. There's some issues I ran into, but they appear to be issue before the refactor:

- Sorting by any column returns: `Model not found. Check the name of your model.`

kevgliss (approved):

PR #4134, opened 2023-12-15
Fixing signal dedup and snooze conflict

kevgliss:


src/dispatch/signal/service.py
```diff
@@ -568,100 +568,135 @@ def update_instance(
...
-                    SignalInstance.signal_id == signal_instance.signal_id,
-                    SignalInstance.created_at >= default_dedup_window,
-                    SignalInstance.id != signal_instance.id,
-                    SignalInstance.case_id.isnot(None),  # noqa
-                )
-                .with_entities(SignalInstance.case_id)
-                .order_by(desc(SignalInstance.created_at))
-                .first()
+        query = apply_filter_specific_joins(SignalInstance, f.expression, query)
+        query = apply_filters(query, f.expression)
```
wssheldon, line 631: basically a nit, but i think we could make this more readable if we moved it into its own function to give a clear name as to what's going on here, and also we use the same code sequence in both functions so putting it into a function makes sense from that perspective as well

```python
def build_filter_query():
        query = db_session.query(SignalInstance).filter(
            SignalInstance.signal_id == signal_instance.signal_id
        )
        query = apply_filter_specific_joins(SignalInstance, f.expression, query)
        query = apply_filters(query, f.expression)
        return query
```

just an examlpe

tests/signal/test_signal_service.py
```diff
@@ -275,6 +275,67 @@ def test_filter_actions_deduplicate(session, signal, project):
...
+        expression=[{"or": [{"model": "Entity", "field": "id", "op": "==", "value": entity.id}]}],
+        action=SignalFilterAction.snooze,
+        expiration=datetime.now(tz=timezone.utc) + timedelta(minutes=5),
+        project=project,
+    )
+    signal.filters.append(signal_filter)
+
+    session.commit()
+    assert filter_signal(db_session=session, signal_instance=signal_instance_2)
+    assert signal_instance_2.filter_action == SignalFilterAction.snooze
```
wssheldon, line 336:
```suggestion
    # we assert the signal was snoozed, even though a deduplication filter was set, and set first
    assert signal_instance_2.filter_action == SignalFilterAction.snooze
```

theres some nuance thats helpful to explain here, i think

PR #4152, opened 2023-12-19
Email Search in `ParticipantSearchPopover`, `Participant` hover card component

wssheldon:
We handle the invalid `EmailStr` by providing a custom validator in `BaseContact` model, which will return a default value for invalid emails, instead of failing the entire request.

We add searching by emails in `ParticipantSearchPopover`

### DIFF
--- a/src/dispatch/models.py
+++ b/src/dispatch/models.py
@@ -1,9 +1,10 @@
-from typing import Optional
+from typing import Optional, Literal
 from datetime import datetime, timedelta
 
 from pydantic.fields import Field
 from pydantic.networks import EmailStr
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
+from pydantic.errors import EmailError
 from pydantic.types import conint, constr, SecretStr
 
 from sqlalchemy import Boolean, Column, DateTime, Integer, String, event, ForeignKey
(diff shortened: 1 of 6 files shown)


src/dispatch/models.py
```diff
@@ -140,11 +141,20 @@ class ResourceBase(DispatchBase):
 
 
 class ContactBase(DispatchBase):
-    email: EmailStr
+    email: str
```
wssheldon, line 144: @kevgliss figured u should see this change specifically

wssheldon: I did test this E2E by modifying an email to some junk value and it works as expected. I get the 422 before, and valid request + default value after this change.

kevgliss: Hmmm, this seems a bit counterintuitive to me? If I put in a bad email, I don't put in a valid email I get a 200 and the email is "unknown@email.com"?

wssheldon: agreed this is probably not ideal, it begs the question how we ended up with one in the first place. it might be hard to figure out at this point. 

without knowing for sure how it ended up in the database, does it make sense to have a dedicated view and model specifically for read operations, to have the robustness in the event that one somehow slips by (otherwise the entire component stops working, when we really just want the bad value to get handled in some way).

PR #4204, opened 2024-01-05
Pins Python protobuf and bumps base and dev dependencies

mvilanova:


whitdog47 (approved):

whitdog47 (commented):
Looks like something broke pytest

PR #4205, opened 2024-01-05
Logs exception when trying to add non-existing user to a case

whitdog47:
Logs exception when trying to add non-existing user to a case rather than throwing exception.


src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -782,16 +786,19 @@ def handle_new_participant_added(
...
 
+            participant = case_flows.case_add_or_reactivate_participant_flow(
+                case_id=context["subject"].id,
+                user_email=user_email,
+                db_session=db_session,
+                add_to_conversation=False,
+            )
+            participant.user_conversation_id = user_id
+        except Exception as e:
+            log.warn(f"Error adding participant {user_id} to case {context['subject'].id}: {e}")
```
metroid-samus, line 800:
```suggestion
            log.warn(f"Error adding participant {user_id} to Case {context['subject'].id}: {e}")
```

metroid-samus (approved):

PR #4218, opened 2024-01-08
Check for case assignee before trying to send reminder

jschroth:
Updated return types for definitions, changed None checks to use "is None" instead of "not" and added a None check for assignee when sending case close reminders

### DIFF
--- a/src/dispatch/case/messaging.py
+++ b/src/dispatch/case/messaging.py
@@ -20,7 +20,7 @@ from dispatch.plugin import service as plugin_service
 log = logging.getLogger(__name__)
 
 
-def send_case_close_reminder(case: Case, db_session: SessionLocal):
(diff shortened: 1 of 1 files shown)


whitdog47 (approved):

src/dispatch/case/messaging.py
```diff
@@ -30,9 +30,13 @@ def send_case_close_reminder(case: Case, db_session: SessionLocal):
     plugin = plugin_service.get_active_instance(
         db_session=db_session, project_id=case.project.id, plugin_type="conversation"
     )
-    if not plugin:
+    if plugin is None:
         log.warning("Case close reminder message not sent. No conversation plugin enabled.")
         return
+
```
whitdog47, line 36:
```suggestion

```

whitdog47: Looks like there was hidden whitespace the linter didn't like.

PR #4242, opened 2024-01-12
Ensure that next oncall observer is added to all workflows

whitdog47:


src/dispatch/incident/flows.py
```diff
@@ -156,6 +156,14 @@ def incident_create_resources(*, incident: Incident, db_session=None) -> Inciden
     individual_participants, team_participants = get_incident_participants(incident, db_session)
     tactical_participant_emails = [i.email for i, _ in individual_participants]
 
+    # we add any observer added in create (like new oncall participant)
+    participant_with_observer_role = participant_service.get_by_incident_id_and_role(
```
mvilanova, line 160: Is it only one or could it be more than one?

src/dispatch/incident/flows.py
```diff
@@ -156,6 +156,14 @@ def incident_create_resources(*, incident: Incident, db_session=None) -> Inciden
...
+        db_session=db_session, incident_id=incident.id, role=ParticipantRoleType.observer
+    )
+    if participant_with_observer_role:
+        # add to list
+        individual_participants.append(participant_with_observer_role)
```
mvilanova, line 165: You're mixing IndividualContacts with Participants here, which will make adding participants to the incident in `incident_add_or_reactivate_participant_flow()` down below fail. The plugin used in `get_incident_participants()` (`DispatchParticipantResolverPlugin`) above returns IndividualContacts. We should fix the definition of `get_incident_participants()`. 
```suggestion
        individual_participants.append(participant_with_observer_role.individual)
```

mvilanova (changes requested):

src/dispatch/incident/flows.py
whitdog47: Good catch - thanks! It seems the return type of `get_incident_participants` should be `tuple[list[IndividualContact | None], list[TeamContact | None]]` - does that seem right.

src/dispatch/incident/flows.py
mvilanova: I guess it won't be a problem because any incident participants with role `Observer` returned by `get_incident_participants(incident, db_session)` are not part of the incident yet.

mvilanova (approved):

PR #4243, opened 2024-01-13
New tag selector component

whitdog47:
Implements an improved tag selector allowing for search by tag name, collapsable/expandable tag type sections, and enforced exclusivity. Tags are shown as removable color chips with tag type icons.
Screenshot 2024-01-12 at 3 56 55 PM

New attributes for tag types include:
- color
- icon
- customizable discoverability by entity type (incidents, cases, queries, signals, and sources)
Screenshot 2024-01-12 at 3 59 58 PM


wssheldon (approved):
Looks good to me! I don't see any glaring issues. After we migrated to Vue 3, I have been trying to use the script setup composition API for new code, as it's the "new" way, and the options API is considered deprecated. 

Not saying we should change the new code in this PR but just as an FYI.

src/dispatch/static/dispatch/src/tag/TagPicker.vue
```diff
@@ -0,0 +1,349 @@
...
+      type: String,
+      default: null,
+    },
+    modelId: {
+      type: Number,
+      default: null,
+    },
+  },
+  directives: {
+    // add directive to close dropdown when clicked outside
```
wssheldon, line 170: Does it not do this by default?

PR #4287, opened 2024-01-24
Ensuring that the tag picker reloads on project selection

whitdog47:
The new tag picker component was not watching the project prop for changes and thus was not filtering properly.


src/dispatch/static/dispatch/src/incident_role/PolicyRoleBuilder.vue
```diff
@@ -57,7 +57,7 @@
               <v-list-item>
                 <tag-filter-auto-complete
                   label="Tags"
-                  :project="project"
+                  project="project"
```
metroid-samus, line 60:
```suggestion
                  :project="project"
```

metroid-samus (changes requested):

metroid-samus (approved):

PR #4289, opened 2024-01-24
Fixes bug in incident cost when stable_at is None

whitdog47:
This PR fixes an issue when the stable_at attribute is None

### DIFF
--- a/src/dispatch/incident_cost/scheduled.py
+++ b/src/dispatch/incident_cost/scheduled.py
@@ -52,7 +52,7 @@ def calculate_incidents_response_cost(db_session: SessionLocal, project: Project
             # after the incident was marked as stable
             if incident.status == IncidentStatus.closed:
                 if incident_response_cost:
-                    if incident_response_cost.updated_at > incident.stable_at:
+                    if incident.stable_at is None or incident_response_cost.updated_at > incident.stable_at:
                         continue
 
             if incident_response_cost is None:


src/dispatch/incident_cost/scheduled.py
```diff
@@ -52,7 +52,7 @@ def calculate_incidents_response_cost(db_session: SessionLocal, project: Project
             # after the incident was marked as stable
             if incident.status == IncidentStatus.closed:
                 if incident_response_cost:
-                    if incident_response_cost.updated_at > incident.stable_at:
+                    if incident.stable_at is None or incident_response_cost.updated_at > incident.stable_at:
```
mvilanova, line 55: If `incident.stable_at is None`, then it means that something went wrong when we were supposed to set the `stable_at` time and this won't alert us about it anymore. Why did we decide to add this check?

metroid-samus: This is to cover a race condition where an incident has not finished updating where the `status` is set but `stable_at` is not.

When an incident is updated, the status is [immediately set](https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident/views.py#L182). However, the status_at field is set in the `incident_update_flow` [background task](https://github.com/Netflix/dispatch/blob/master/src/dispatch/incident/views.py#L186).

In the event of this edge case, the response cost for following incidents is not updated.

whitdog47:
I'm closing this PR as no longer valid.

PR #4299, opened 2024-01-26
Add reporter to individual_participants list

wssheldon:
I walked back from `Adding the following individuals to help resolve this case` to resolve an issue where Case reporter is not added to the Case conversation. We pass some variant of `participant_emails` all the way down from `case_create_resource_flow`, where this list of emails is populated. It looks like we never add reporter here, if it exists, even though we do it for the assignee.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -622,6 +622,9 @@ def case_create_resources_flow(
     if case.assignee:
         individual_participants.append((case.assignee.individual, None))
 
+    if case.reporter:
+        individual_participants.append((case.reporter.individual, None))
+
     if create_all_resources:
         # we create the tactical group
         direct_participant_emails = [i.email for i, _ in individual_participants]


metroid-samus (approved):

src/dispatch/case/flows.py
```diff
@@ -622,6 +622,9 @@ def case_create_resources_flow(
     if case.assignee:
         individual_participants.append((case.assignee.individual, None))
 
+    if case.reporter:
+        individual_participants.append((case.reporter.individual, None))
```
wssheldon, line 626: It's used below in the function, like so:

```python
        # wait until all resources are created before adding suggested participants
        individual_participants = [x.email for x, _ in individual_participants]
        for email in individual_participants:
            # we don't rely on on this flow to add folks to the conversation because in this case
            # we want to do it in bulk
            case_add_or_reactivate_participant_flow(
                db_session=db_session,
                user_email=email,
                case_id=case.id,
                add_to_conversation=False,
            )
        # explicitly add the assignee to the conversation
        all_participants = individual_participants + [case.assignee.individual.email]
```

PR #4301, opened 2024-01-26
Bugfix/do not overwrite stable at

metroid-samus:
Currently, all non-active incidents have their `stable_at` field set to `datetime.now()`, even if the `stable_at` field is already set. This fix ensures that all non-active incidents with non null `stable_at` fields do not have those fields overwritten.


whitdog47 (changes requested):
Please deconflict this with PR 4300

metroid-samus:
> Please deconflict this with PR 4300

Done.

whitdog47 (approved):

PR #4307, opened 2024-01-30
Remove ability for any user to change password

whitdog47:
Only allows the current user to change their own password unless their role is Admin or above.


mvilanova (approved):
LGTM

PR #4335, opened 2024-02-01
Bugfix/null cost model select

metroid-samus:
Only add the cost model block to Slack modals when they exist. This is because block kit throws an error when trying to render a null block.

### DIFF
--- a/src/dispatch/incident/severity/service.py
+++ b/src/dispatch/incident/severity/service.py
@@ -96,7 +96,7 @@ def get_by_name_or_default(
     if incident_severity_in:
         if not incident_severity_in.project or incident_severity_in.project.id != project_id:
             log.warning(
-                "The selected incident severity is not associated with the same project as the incident. Searching for the nearest matching incident severity."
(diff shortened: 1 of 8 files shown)


src/dispatch/incident/severity/service.py
```diff
@@ -91,6 +94,10 @@ def get_by_name_or_default(
 ) -> IncidentSeverity:
     """Returns an incident severity based on a name or the default if not specified."""
     if incident_severity_in:
+        if not incident_severity_in.project or incident_severity_in.project.id != project_id:
```
kevgliss, line 97: If I'm understanding correctly, I think these should be exceptions as the caller has clearly become confused e.g. the `incident_in.project_id` differs with `project_id`.

metroid-samus: I went with the less strict approach due to the doc string. 
My interpretation of the doc string was that in the event multiple projects have severities with the same name, an incident's severity should resolve to the one within its project.

I'm pretty amenable to making these constraints more strict though. How strict do we want to enforce these project/incident/case-level separations moving forward?

PR #4340, opened 2024-02-02
Adds http url validation to all weblinks

kevgliss:


whitdog47 (approved):

whitdog47 (changes requested):
Let's see if we can get playwright fixed first

jschroth (approved):

whitdog47 (approved):

src/dispatch/individual/models.py
```diff
@@ -63,12 +63,21 @@ class IndividualContact(Base, ContactMixin, ProjectMixin):
...
     external_id: Optional[str] = Field(None, nullable=True)
 
+    @validator("weblink")
+    def weblink_validator(cls, v):
+        if v is None:
+            return v
+        if not isinstance(v, HttpUrl) and len(v) > 0:
+            raise ValueError(
+                "weblink is not an empty string and not a valid weblink")
+        return v
```
jschroth, line 79: this is a small nit, but I prefer something that is more explicit and a raise if not.
```suggestion
        if v is None or v == "" or isinstance(v, HttpUrl):
            return v

        raise ValueError("weblink is not an empty string or a valid weblink")
```

jschroth: I forget which order these need to be in so you could reverse the last two to `isinstance(v, HttpUrl) or v == ""`

whitdog47: I reversed the last two. Thanks for helping clean up the logic!

PR #4346, opened 2024-02-05
Do not allow user password changes via API

mvilanova:
* This change removes the ability to change a user's password via API, as user password changes are only supported via CLI and not UI.
* We still allow for changes to default projects, and roles if the user requesting the role change has `Owner` role.

### DIFF
--- a/src/dispatch/auth/views.py
+++ b/src/dispatch/auth/views.py
@@ -145,10 +145,13 @@ def update_user(
         )
 
     if user_in.role:
+        # New user role is provided
         user_organization_role = user.get_organization_role(organization)
         if user_organization_role != user_in.role:
+            # New user role provided is different than current user role
             current_user_organization_role = current_user.get_organization_role(organization)
             if current_user_organization_role != UserRoles.owner:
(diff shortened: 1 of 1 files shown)


src/dispatch/auth/views.py
```diff
@@ -163,6 +166,9 @@ def update_user(
         UserOrganization(role=user_in.role, organization=OrganizationRead(name=organization))
     ]
 
+    # we currently only allow user password changes via CLI, not UI/API.
+    user_in.password = None
```
mvilanova, line 170: This will make this [if](https://github.com/Netflix/dispatch/blob/master/src/dispatch/auth/service.py#L223) to evaluate as `False`, so password won't be updated via API, but still allow for password changes via CLI.

src/dispatch/auth/views.py
```diff
@@ -145,10 +145,13 @@ def update_user(
         )
 
     if user_in.role:
```
mvilanova, line 147: Permission check logic was moved to the function to only check for owner role when role change is requested and still allow for other changes like default projects.

whitdog47 (approved):

PR #4369, opened 2024-02-07
Fixing issue when event comes from an unrecognized sender

whitdog47:
If the `individual` is not found (for instance, if a Slack message from a bot is marked for timeline inclusion), then do not try to get the name.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -709,10 +709,12 @@ def handle_timeline_added_event(
         except Exception:
             individual = None
 
+        source = f"Slack message from {individual.name}" if individual else "Slack message"
+
         # we log the event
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -709,10 +709,12 @@ def handle_timeline_added_event(
         except Exception:
             individual = None
 
+        source = f"Slack message from {individual.name}" if individual else "Slack message"
```
mvilanova, line 712: Any chance we could include the name of the App/bot?

mvilanova (approved):

src/dispatch/plugins/dispatch_slack/incident/interactive.py
whitdog47: Good idea - done!

PR #4375, opened 2024-02-08
Replaces hyperlinks in a specified Google Doc.

metroid-samus:
If the url of a hyperlink contains a placeholder, it will be replaced with the corresponding value in a replacements list.

For example, if we aim to substitute all instances of `{{name}}` in a Google Doc with `netflix`, a hyperlink that originally directed to `http://www.{{name}}.com` will be transformed into `http://www.netflix.com`

### DIFF
--- a/src/dispatch/plugins/dispatch_google/docs/plugin.py
+++ b/src/dispatch/plugins/dispatch_google/docs/plugin.py
@@ -24,7 +24,7 @@ def remove_control_characters(s):
     return "".join(ch for ch in s if unicodedata.category(ch)[0] != "C")
 
 
-def find_links(obj: Any, find_key: str) -> Iterator[List[Any]]:
+def find_links(obj: dict | list, find_key: str) -> Iterator[list[Any]]:
     """Enumerate all the links found.
     Returns a path of object, from leaf to parents to root.
 
@@ -48,7 +48,7 @@ def find_links(obj: Any, find_key: str) -> Iterator[List[Any]]:
                 yield found
 
 
-def iter_links(document: dict) -> List[Tuple[str, str]]:
+def iter_links(document: dict) -> list[tuple[str, str]]:
     """Find all the links and return them.
 
     This method was originally implemented in the open source library `Beancount`.
--- a/src/dispatch/plugins/dispatch_google/docs/plugin.py
+++ b/src/dispatch/plugins/dispatch_google/docs/plugin.py
@@ -58,16 +58,16 @@ def find_links(obj: dict, find_key: str) -> iter(list[Any]):
                 yield found
 
 
-def iter_links(document_content: dict) -> list[tuple[str, str]]:
+def iter_links(document_content: list) -> list[tuple[str, str]]:
     """Find all the links and return them.
     Parameters:
-        document_content (dict): The contents of the body of a Google Doc (googleapi.discovery.Resource).
+        document_content (list): The contents of the body of a Google Doc (googleapi.discovery.Resource).
 
     Returns:
         list: A list of tuples containing the hyperlink url (str) and its corresponding hyperlink element (str).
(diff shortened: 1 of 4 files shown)


wssheldon:
It would be relatively easy to unit test the three functions involved here, we can mock response of content from google. I think this would be a good thing to add for this change.

src/dispatch/plugins/dispatch_google/docs/plugin.py
```diff
@@ -23,7 +24,80 @@ def remove_control_characters(s):
...
+                    found.append(obj)
+                    yield found
+    elif isinstance(obj, list):
+        for value in obj:
+            for found in find_links(value, find_key):
+                found.append(obj)
+                yield found
+
+
+def iter_links(document: dict) -> List[Tuple[str, str]]:
```
wssheldon, line 51:
```suggestion
def iter_links(document: dict) -> list[tuple[str, str]]:
```
you don't need to import these from typing

src/dispatch/plugins/dispatch_google/docs/plugin.py
```diff
@@ -23,7 +24,80 @@ def remove_control_characters(s):
...
+    """Find all the links and return them.
+
+    This method was originally implemented in the open source library `Beancount`.
+    The original source code can be found at
+    https://github.com/beancount/beancount/blob/master/tools/transform_links_in_docs.py.
+    BeanCount is licensed under the GNU GPLv2.0 license.
+    """
```
wssheldon, line 58: `Example:` in google style docstring would be a nice QOL improvement here as the code is quite obtuse if you don't know what the response of content looks like from Google

src/dispatch/plugins/dispatch_google/docs/plugin.py
```diff
@@ -23,7 +24,80 @@ def remove_control_characters(s):
     return "".join(ch for ch in s if unicodedata.category(ch)[0] != "C")
 
 
-def replace_text(client: Any, document_id: str, replacements: List[str]):
+def find_links(obj: Any, find_key: str) -> Iterator[List[Any]]:
```
wssheldon, line 27:
```suggestion
def find_links(obj: dict | list, find_key: str) -> Iterator[list[Any]]:
```

Iterator should be imported from `_collections_abc` i believe.

src/dispatch/plugins/dispatch_google/docs/plugin.py
```diff
@@ -23,7 +24,80 @@ def remove_control_characters(s):
...
+        for item in jpath:
+            if "textRun" in item:
+                link = item["textRun"]["textStyle"]["link"]
+                if "url" not in link:
+                    continue
+                url = link["url"]
+                yield (url, item)
+
+
+def replace_weblinks(client: Any, document_id: str, replacements: List[str]) -> None:
```
wssheldon, line 69:
```suggestion
def replace_weblinks(client: Any, document_id: str, replacements: list[str]) -> None:
```

We should use the actual type for `client` so that we can get LSP support on `client` methods.

src/dispatch/plugins/dispatch_google/docs/plugin.py
metroid-samus: Done.

src/dispatch/plugins/dispatch_google/docs/plugin.py
metroid-samus: Done.

metroid-samus:
> It would be relatively easy to unit test the three functions involved here, we can mock response of content from google. I think this would be a good thing to add for this change.

Done.

wssheldon (approved):
Looks good.

wssheldon:
For these external service API tests where we don't control the response we can consider using [vrcpy](https://vcrpy.readthedocs.io/en/latest/) in the future so they're more maintainable.

PR #4386, opened 2024-02-09
Fixes error from accessing a None type object.

metroid-samus:
Fixes error from accessing a None type object by moving outstanding_action initialization outside an if statement.


wssheldon:
Have you tested a tactical report in Slack where there the error condition exists?

wssheldon (approved):

metroid-samus:
> Have you tested a tactical report in Slack where the error condition exists?

Yes. I've replicated the issue and tested this bugfix

PR #4389, opened 2024-02-09
Moves a folder or file from a Google Drive to trash instead of immediately deleting it

metroid-samus:
This is a soft delete, the file will still exist in the trash.


jschroth:
either this or https://github.com/Netflix/dispatch/pull/4388 was fine with me.

PR #4396, opened 2024-02-13
Changes signals ingestion from opt in to opt out approach

mvilanova:
This PR changes the signal ingestion approach from opt-in to opt-out for signals submitted via API or consumed through a signal consumer plugin.

Screenshot 2024-02-15 at 1 43 02 PM


wssheldon:
It looks good. Have you tested ingesting a signal instance w/ and w/o a definition locally?

mvilanova:
> It looks good. Have you tested ingesting a signal instance w/ and w/o a definition locally?

Not yet. Dealing with an issue with the signal definition create form.

src/dispatch/signal/service.py
```diff
@@ -359,14 +359,14 @@ def create(*, db_session: Session, signal_in: SignalCreate) -> Signal:
...
             exclude={
-                "project",
+                "case_priority",
                 "case_type",
-                "case_prioity",
-                "source",
+                "entity_types",
                 "filters",
-                "tags",
-                "entty_types",
```
mvilanova: This typo prevented me from creating the default definition manually.

src/dispatch/signal/service.py
```diff
@@ -359,14 +359,14 @@ def create(*, db_session: Session, signal_in: SignalCreate) -> Signal:
     signal = Signal(
         **signal_in.dict(
             exclude={
-                "project",
+                "case_priority",
                 "case_type",
-                "case_prioity",
```
mvilanova: This typo prevented me from creating the default definition manually.

mvilanova:
@whitdog47 looks like your changes [here](https://github.com/Netflix/dispatch/pull/4336) broke the signal definition create form. I added support for checking via project name to fix it.

src/dispatch/signal/views.py
```diff
@@ -73,47 +74,40 @@ def create_signal_instance(
...
-            detail=[{"msg": msg}],
-        ) from None
+    signal_instance_in.signal = signal_definition
 
     try:
         signal_instance = signal_service.create_instance(
             db_session=db_session, signal_instance_in=signal_instance_in
         )
-        signal_instance.signal = signal
+        signal_instance.signal = signal_definition
```
mvilanova, line 110: I'm not sure this is needed if we are already doing the association [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/signal/service.py#L539)

mvilanova:
> It looks good. Have you tested ingesting a signal instance w/ and w/o a definition locally?

Tested the code with a signal with external id/variant and one without it.

src/dispatch/signal/views.py
wssheldon: Yeah, seems redundant -- also in `create_instance` we refetch it and I am not sure that's necessary either.

```python
signal = get(db_session=db_session, signal_id=signal_instance_in.signal.id)
```

src/dispatch/signal/views.py
```diff
@@ -73,47 +74,40 @@ def create_signal_instance(
...
-                status_code=status.HTTP_400_BAD_REQUEST,
-                detail=[{"msg": msg}],
-            ) from None
+    if not signal_definition:
+        # we get the default signal definition
+        signal_definition = signal_service.get_default(
+            db_session=db_session,
+            project_id=project.id,
+        )
+        msg = "Default signal definition used for signal instance with external id {external_id} or variant {variant}."
```
wssheldon, line 93:
```suggestion
        msg = f"Default signal definition used for signal instance with external id {external_id} or variant {variant}."
```

wssheldon (approved):
Just the `f` string

src/dispatch/signal/views.py
mvilanova: Yeah, we're already including signal in `signal_instance_in`. Also, the `db_session.commit()` [here](https://github.com/Netflix/dispatch/blob/master/src/dispatch/signal/views.py#L117) seems redundant too as we already do it at the end of `create_instance`

mvilanova: The get inside `create_instance` was needed. Otherwise, the app threw a `AttributeError: 'SignalRead' object has no attribute '_sa_instance_state'` exception

PR #4404, opened 2024-02-16
Makes the assignee also the reporter in cases created from signals

mvilanova:
This PR fixes an issue when trying to edit a case via Slack and the case does not have a reporter.


mvilanova:
Another option would be to use the `signal_instance.signal.owner` as reporter when creating the case, but we would need to ensure it's an email address and we could potentially run into issues when trying to resolve the email to a Slack user, etc.

wssheldon:
My preference would that reporter / assignee is allowed to be `None`. That feels more intuitive to me as a user than the reporter automatically getting set to the assignee.

PR #4430, opened 2024-02-22
Add required as a tag type attribute

whitdog47:
This PR adds a new tag type attribute called "required". When set, at least one tag of this tag type must be selected.

image


src/dispatch/static/dispatch/src/tag/TagPicker.vue
```diff
@@ -269,10 +283,26 @@ const emit = defineEmits(["update:modelValue"])
...
   if (all_tags_in_project) {
-    error.value = true
-    dummyText.value += " "
+    if (are_required_tags_selected(value)) {
+      error.value = true
+      dummyText.value += " "
+    } else {
+      error.value = "Please select at least one tag from each required category"
+      dummyText.value += " "
+    }
```
metroid-samus, line 305:
```suggested
    if (are_required_tags_selected(value)) {
      error.value = true
    } else {
      error.value = "Please select at least one tag from each required category"
    }
    dummyText.value += " "
```

src/dispatch/static/dispatch/src/tag_type/NewEditSheet.vue
```diff
@@ -126,6 +126,21 @@
...
+                  label="Required"
+                  hint="Is this tag type required for new entities of the discoverable types?"
+                />
+              </v-col>
+              <v-col cols="7">
+                <v-tooltip max-width="500px" open-delay="50" location="bottom">
+                  <template #activator="{ props }">
+                    <v-icon class="mt-4" v-bind="props">mdi-information</v-icon>
+                  </template>
+                  <span> If activated, only one tag of this type is allowed per incident. </span>
```
wssheldon, line 141: Is it "only one" or is it "at least one"?

src/dispatch/static/dispatch/src/tag/TagPicker.vue
whitdog47: Fixed

src/dispatch/static/dispatch/src/tag_type/NewEditSheet.vue
whitdog47: For exclusive, it is only one.

src/dispatch/static/dispatch/src/tag/TagPicker.vue
```diff
@@ -269,14 +283,28 @@ const emit = defineEmits(["update:modelValue"])
...
+    all_tags_in_project = value.every((tag) => tag.project?.name == project_name)
+  }
   if (all_tags_in_project) {
-    error.value = true
-    dummyText.value += " "
+    if (are_required_tags_selected(value)) {
+      error.value = true
+    } else {
+      error.value = "Please select at least one tag from each required category"
```
wssheldon, line 302: We can educate the user how to address the error here. Correct me if I am wrong, but right now it's not immediately obvious "which" categories are required.

whitdog47: I added a screenshot to the description showing how the Required label shows up in the selector. Let me know if you think of a better way to call this out.

wssheldon: I think we could include "The following categories are required: example, test, pizza" to the main error message.

whitdog47: Good idea! I've added a parenthetical list of required type(s) to the error message.

wssheldon (approved):

PR #4448, opened 2024-02-26
Add a Required column to the TagType table

wssheldon:
In https://github.com/Netflix/dispatch/pull/4430 / Add required as a tag type attribute, we added support for `Required` tag types, this PR adds a `Required` column that indicates whether a tag type is required in the backend tag type settings table.

After marking as required

Screenshot 2024-02-26 at 12 49 07 PM

Before marking as required

Screenshot 2024-02-26 at 12 48 58 PM


whitdog47 (approved):
thanks!

PR #4457, opened 2024-02-28
Only run member join or remove for incident channels

whitdog47:


jschroth (approved):

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -955,6 +955,10 @@ def handle_member_joined_channel(
             "Unable to handle member_joined_channel Slack event. Dispatch user unknown."
         )
 
+    if context["subject"].type != "incident":
```
mvilanova, line 958: Instead of hardcoding `incident` we could add a model called `IncidentSubjects` in [this module](https://github.com/Netflix/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/models.py) like the `CaseSubjects` one and do `IncidentSubjects.incident`

mvilanova (approved):

src/dispatch/plugins/dispatch_slack/incident/interactive.py
whitdog47: Great idea! I cleaned up all instances of hardcoded "Incident" and "case" in the dispatch_slack package and changed it to an enum.

PR #4502, opened 2024-03-12
Adds incident participant preview option during incident creation.

metroid-samus:
This PR enables admin users to preview the list of designated participants when creating a new incident via the Dispatch UI.
This allows to the user ensure only the intended individuals are involved when the incident is opened.

image

Screenshot 2024-03-12 at 12 10 49 PM


src/dispatch/participant/service.py
```diff
@@ -213,8 +218,68 @@ def get_or_create(
...
+        if individual_contact and contact_plugin:
+            # We get information about the individual
+            individual_info = contact_plugin.instance.get(
+                individual_contact.email, db_session=db_session
+            )
+
+        location = individual_info.get("location", "Unknown")
+        department = individual_info.get("department", "Unknown")
+        team = individual_contact.email.split("@")[1] if individual_contact else ""
+        team = individual_info.get("team", team)
```
whitdog47, line 251:
```suggestion
        domain = individual_contact.email.split("@")[1] if individual_contact else ""
        team = individual_info.get("team", domain)
```

whitdog47 (changes requested):
small nit for clarity

PR #4551, opened 2024-03-26
[Enhancement] - Cases, but they can be Slack channels

ghost:
#### Summary

A highly requested feature is for Cases to support dedicated channels, just like Incidents do today. This PR adds that functionality. We refactor some functions to use the `Subject` pattern, so that they work for both incidents, and cases. Notably, (Bookmarks and Announcement Messages).

When a dedicated channel Case is escalated to an Incident, we reuse the `Conversation` object for the new `Incident`. There was already a conditional in `create_incident_resources` that only created a conversation if one did not exist on the incident object, so we leverage this, and it worked seamlessly. Case channels are prepended with `case-` so that they are not ambiguous. The channel is renamed on escalation to the new `Incident` name and the `case-` is naturally removed at this point.

We add some Slack `/slash` commands to make the `Case` dedicate channel experience better. Allowing you to update cases and escalate them directly in Slack.

##### Miscellaneous Changes

- We remove the participant announcement messages on `Incident` creation and replace them with a "bulk" version, that sends one message, with the number of participants added, and their cards.
- We refactor the `triage_flow` for cases to use a `match` statement to improvement readability.
- We add tooltips to the `Case` page for all resource icons, the `private` icon, and the `escalate` icon.
- We truncate the description on the `Case Type` dropdown on the Case report form.

Screenshot 2024-03-26 at 3 03 44 PM


wssheldon:
@mattsta Why the confused emoji?

src/dispatch/conversation/flows.py
```diff
@@ -38,11 +40,18 @@ def create_case_conversation(case: Case, conversation_target: str, db_session: S
 
     conversation = None
 
+    use_channel = True
```
wssheldon, line 43:
```suggestion
```

Debugging line.

src/dispatch/group/flows.py
```diff
@@ -32,6 +32,9 @@ def create_group(
     if group_type == GroupType.notifications:
         group_name = f"{subject.name}-{GroupType.notifications}"
 
+    # TODO(wshel): LOCAL UNBLOCKER: Avoid Duplicates on Group creation
+    group_name = f"{group_name}_test"
+
```
wssheldon, line 37:
```suggestion

```

Debugging line.

src/dispatch/incident/flows.py
```diff
@@ -310,7 +310,6 @@ def incident_create_resources_flow(
     return incident_create_resources(incident=incident, db_session=db_session)
 
 
-@background_task
```
wssheldon, line 313: Encountered race condition where `commit()` was called before the `incident` object existed in the database, `None` is called on `commit()`

mattsta:
oh, sorry, I didn't realize homepage reactions are actually issue reactions. didn't expect anybody to see it outside of the promo box itself (I always thought the home feed reactions were in their own isolated universe somehow?).

Screenshot 2024-03-26 at 18 24 58

initial reaction was more "cases are sad to have, wish everything just worked." i'd also disagree with it being "confused" emoji as they have it labeled? never noticed that was its official title before. always thought it was more "sad / sorry / condolences" but who knows.

I think per-case channels are a great idea though. Much better than the CEO dropping into #engineering with WHY IS THE WEBSITE DOWN then starting a secret video chat without any historical trace of troubleshooting progress and participants when new contacts/departments/helpers need to be brought in mid-event. 👍

src/dispatch/case/flows.py
```diff
@@ -529,10 +531,38 @@ def case_to_incident_endpoint_escalate_flow(
     case.escalated_at = datetime.utcnow()
     case.status = CaseStatus.escalated
 
+    incident = incident_service.get(db_session=db_session, incident_id=incident_id)
+
+    channel_case = True if case.conversation.thread_id else False
```
mvilanova, line 536: You return True if it's a thread based conversation, which I find it a little bit confusing when reading the name of the variable `channel_case`
```suggestion
    channel_case = False if case.conversation.thread_id else True
```
or
```
channel_case = True if not case.conversation.thread_id else False
```

src/dispatch/case/flows.py
```diff
@@ -529,10 +531,38 @@ def case_to_incident_endpoint_escalate_flow(
...
+    # This is a channel based Case, so we reuse the case conversation for the incident
+    if not channel_case:
```
mvilanova, line 538:
```suggestion
    if channel_case:
```

src/dispatch/case/flows.py
```diff
@@ -529,10 +531,38 @@ def case_to_incident_endpoint_escalate_flow(
...
+        incident.conversation = case.conversation
+        db_session.add(incident)
+        db_session.commit()
+
     # we run the incident create flow
     incident = incident_flows.incident_create_flow(
         incident_id=incident_id, organization_slug=organization_slug, db_session=db_session
     )
+
+    if not channel_case:
```
mvilanova, line 548:
```suggestion
    if channel_case:
```

src/dispatch/case/flows.py
wssheldon: Yeah sorry, it's not ready for review yet -- locally I change this to a property on the Case model so it's more intuitive:

```python
    @property
    def has_channel(self) -> bool:
        return True if not self.conversation.thread_id else False

    @property
    def has_thread(self) -> bool:
        return True if self.conversation.thread_id else False
```

mvilanova: Got it. I'll wait until it's ready then :)

wssheldon:
@mvilanova @whitdog47 @metroid-samus Ready for review now if you'd like to take a look.

src/dispatch/conversation/flows.py
```diff
@@ -53,11 +71,12 @@ def create_case_conversation(case: Case, conversation_target: str, db_session: S
 
     conversation.update({"resource_type": plugin.plugin.slug, "resource_id": conversation["id"]})
 
+    print(f"got convo: {conversation}")
```
wssheldon, line 74:
```suggestion
```

src/dispatch/database/revisions/tenant/versions/2024-04-09_3a33bc153e7e.py
```diff
@@ -0,0 +1,29 @@
+"""Creates a column for `dedicated_channel` on the Case model to support Cases
+in dedicated coversation channel (as opposed to only threads).
+
+Revision ID: 3a33bc153e7e
+Revises: 71cb25c06fa0
```
whitdog47, line 5:
```suggestion
Revises: b07bb852fd67
```

src/dispatch/database/revisions/tenant/versions/2024-04-09_3a33bc153e7e.py
```diff
@@ -0,0 +1,29 @@
...
+Create Date: 2024-04-09 16:28:06.148971
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "3a33bc153e7e"
+down_revision = "71cb25c06fa0"
```
whitdog47, line 15:
```suggestion
down_revision = "b07bb852fd67"
```

PR #4555, opened 2024-03-27
Validating uuid before storing

whitdog47:


src/dispatch/signal/service.py
```diff
@@ -584,6 +594,14 @@ def create_instance(
         )
         signal_instance.case_type = case_type
 
+    if not is_valid_uuid(signal_instance.id):
```
wssheldon, line 597: I would check even sooner, right below:

```python
    # we round trip the raw data to json-ify date strings
    signal_instance = SignalInstance(
        **signal_instance_in.dict(
            exclude={
                "case",
                "case_priority",
                "case_type",
                "entities",
                "external_id",
                "project",
                "raw",
                "signal",
            }
        ),
        raw=json.loads(json.dumps(signal_instance_in.raw)),
        project=project,
        signal=signal,
    )
```

Since do all the extra stuff below that is a waste anyway.

whitdog47: moved up to after the check in `raw` just in case

wssheldon (approved):

PR #4565, opened 2024-03-29
github_utils: Allowing long return window and adding 'improvement' category

whitdog47:


utils/github_utils.py
```diff
@@ -97,6 +97,7 @@ def release_notes(pull_request_number: int) -> NoReturn:
         "feature": "",
         "techdebt": "",
         "tests": "",
+        "improvement": "",
```
mvilanova, line 100: Isn't it very similar to enhancement?

whitdog47: Yes, I don't know what was marked as improvement, but without it in the map, it breaks the util.

PR #4580, opened 2024-04-03
[Bug] Adding the message template in SlackAPIError when channel is archived

jschroth:
Something is causing the an attempt to do something with an archived incident channel and incident. In this case, in the incident commander is removed from the channel because the slack account is deactivated and it is unclear why there is a job trying to add the person back. Trying to figure out where this is happening and the only thing, aside from updating all the calls with different notifications types, was to include the template. Then when/if it happens again, we can trace where it is coming from.


jschroth:
I believe the error I am trying to find is in `send_incident_commander_readded_notification` - just decline this if you do not think this will be useful for debugging other issues

metroid-samus (approved):

PR #4586, opened 2024-04-04
Send ephemeral message when participant is added to an incident due to an assigned task

whitdog47:
Also fixes error on Tasks tab when the user tries to filter by more than one category and sorting was not working correctly.


src/dispatch/database/service.py
```diff
@@ -377,11 +377,14 @@ def apply_filter_specific_joins(model: Base, filter_spec: dict, query: orm.query
...
     for filter_model in filter_models:
         if model_map.get((model, filter_model)):
             joined_model, is_outer = model_map[(model, filter_model)]
             try:
-                query = query.join(joined_model, isouter=is_outer)
+                if joined_model not in joined_models:
+                    query = query.join(joined_model, isouter=is_outer)
+                    joined_models.append(joined_model)
             except Exception as e:
                 log.debug(str(e))
```
kevgliss, line 389: I wonder if this shouldn't just be `log.exception`?

src/dispatch/database/service.py
```diff
@@ -437,6 +440,12 @@ def create_sort_spec(model, sort_by, descending):
             direction = "desc" if direction else "asc"
 
             # we have a complex field, we may need to join
+            try:
+                new_field = json.loads(field)
+                field = new_field.get("key", "")
+            except json.JSONDecodeError:
+                pass
```
kevgliss, line 447: I think I understand this to mean that sometimes the field will be json but other times not and that is _expected_. It might be good to add a comment to that effect and then explain the two scenarios with an example.

src/dispatch/task/service.py
```diff
@@ -73,6 +75,19 @@ def create(*, db_session, task_in: TaskCreate) -> Task:
 
     assignees = []
     for i in task_in.assignees:
+        participant = participant_service.get_by_incident_id_and_email(
+            db_session=db_session, incident_id=incident.id, email=i.individual.email
+        )
+
+        if not participant or not participant.active_roles:
+            # send emphemeral message to user
```
kevgliss, line 83: I'd either delete this or add more context to the above line stating that we want to give users context as to why they are being added to an incident.

kevgliss (approved):

PR #4587, opened 2024-04-04
Fixing date in date picker from modelValue

whitdog47:
Fixes time zone issue with converting date from URL (since default is UTC) by breaking down parts of date string.

### DIFF
--- a/src/dispatch/static/dispatch/src/components/DateWindowInput.vue
+++ b/src/dispatch/static/dispatch/src/components/DateWindowInput.vue
@@ -52,6 +52,15 @@ let today = function () {
 
 const DATE_FORMAT = "yyyy-MM-dd"
 
+function convertStringToDate(value) {
+  let parts = value.split("-")
+  let year = parseInt(parts[0], 10)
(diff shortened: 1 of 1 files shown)


src/dispatch/static/dispatch/src/components/DateWindowInput.vue
```diff
@@ -52,6 +52,15 @@ let today = function () {
 
 const DATE_FORMAT = "yyyy-MM-dd"
 
+function convertStringToDate(value) {
+  let parts = value.split("-")
+  let year = parseInt(parts[0], 10)
+  let month = parseInt(parts[1], 10) - 1
```
jschroth, line 58: subtract 1 from the month? So if the month is 4, it will always make it 3? Then the next time it runs, this.modelValue.end will have month 3 so then it will be 2?

whitdog47: Stupidly, JavaScript months are 0-based.

jschroth (approved):

PR #4651, opened 2024-04-23
Fixing Report Case card to ensure case type reflows on project change

whitdog47:
Ensure that the case type will repopulate with correct types from project when user changes the project.

image

### DIFF
--- a/src/dispatch/static/dispatch/src/case/type/CaseTypeSelect.vue
+++ b/src/dispatch/static/dispatch/src/case/type/CaseTypeSelect.vue
@@ -149,13 +149,6 @@ export default {
       CaseTypeApi.getAll(filterOptions).then((response) => {
         this.items = response.data.items
 
-        if (this.case_type) {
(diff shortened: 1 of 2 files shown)


src/dispatch/static/dispatch/src/case/type/CaseTypeSelect.vue
```diff
@@ -129,31 +129,33 @@ export default {
...
-        })
-      }
-      filterOptions.itemsPerPage = this.numItems
       CaseTypeApi.getAll(filterOptions).then((response) => {
+        this.items = response.data.items
+
+        if (this.case_type) {
+          // check to see if the current selection is available in the list and if not we add it
+          if (!this.items.find((match) => match.id === this.case_type.id)) {
+            this.items = [this.case_type].concat(this.items)
```
metroid-samus, line 155: We might not want to add `this.case_type` to the `items`. When `this.case_type.project != this.project`, `this.case_type` still shows up in the items.

PR #4655, opened 2024-04-24
bug: Accurately captures severity level at time of incident report

Meandmybadself:
# Issue
At present, when reporting a bug, the severity, no matter what is input in the dialog will default to the project's default severity.

# Resolution
Instead of depending on the default severity, this updates the incident service to use the incident severity service's `get_by_name_or_default` method just like the priority uses above.


wssheldon (approved):
Thank you! ⭐

PR #4661, opened 2024-04-26
Don't send a conversation update if no thread

whitdog47:


src/dispatch/case/flows.py
```diff
@@ -127,6 +127,11 @@ def case_add_or_reactivate_participant_flow(
 
 def update_conversation(case: Case, db_session: SessionLocal):
```
wssheldon, line 128:
```suggestion
def update_conversation(case: Case, db_session: Session) -> None:
```
scouts code

PR #4664, opened 2024-04-27
feature: Allows fallback to Basic Auth in PKCE mode

Meandmybadself:
## Feature
When using PKCE to facilitate authentication, owners and admins can still create username/password-based users if they choose to enable `DISPATCH_AUTH_REGISTRATION_ENABLED`.  These types of users make it much simpler to introduce API-based interactions.

Presently, if a username/password-based user attempts to authenticate in a PKCE auth strategy, their supplied auth token lacks the necessary properties to fulfill the token & erroneously fails.

## Changes
This introduces a simple fallback check to see if the supplied token fulfills PKCE needs.  If it doesn't, it falls back to Basic Auth.


Meandmybadself:
Requesting a code review / pointers to documentation for requesting a code review – @whitdog47 @kevgliss @milavent

PR #4703, opened 2024-05-09
Moves all calls to .all() within the */service.py files

metroid-samus:
This PR moves all session queries with `all.()` calls to `*/service.py` files for consistency.


wssheldon:
Nice! I think another good addition here would be to fix all the return types. That way, if you try to call `.all()` on one of these service functions you will get a type error w/ `pyright`.

`list[Whatever] | None` in most cases, I believe.

src/dispatch/tag/scheduled.py
```diff
@@ -54,7 +55,7 @@ def sync_tags(db_session: SessionLocal, project: Project) -> NoReturn:
 @scheduled_project_task
 def build_tag_models(db_session: SessionLocal, project: Project) -> NoReturn:
```
wssheldon, line 56:
```suggestion
def build_tag_models(db_session: SessionLocal, project: Project) -> None:
```

src/dispatch/case/service.py
```diff
@@ -69,7 +69,7 @@ def get_by_name_or_raise(*, db_session, project_id: int, case_in: CaseRead) -> C
 
 def get_all(*, db_session, project_id: int) -> List[Optional[Case]]:
```
wssheldon, line 70: `List` and `Optional` are both deprecated. `List` -> `list` (no import required) and `Optional` becomes `Case | None` (no import required)

recommend we use this syntax going forward for consistency.

metroid-samus: Acknowledged. I've created a separate PR to handle the List/Optional/Union changes.
https://github.com/Netflix/dispatch/pull/4744

PR #4706, opened 2024-05-09
bug: Fixes tags not showing up during incident report / update

Meandmybadself:
# Issue
When attempting to define tags during incident reporting / updating in Slack, we were not seeing any tags autocomplete even though we have a large number associated w/ the selected project.

Screenshot 2024-05-08 at 12 02 58 PM

# Root Cause
[handle_tag_search_action](https://github.com/Meandmybadself/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/incident/interactive.py#L233) uses `project_id` as defined in the BoltContext to query for tags associated with a project.  
This value is initially set in [subject_middleware](https://github.com/Meandmybadself/dispatch/blob/master/src/dispatch/plugins/dispatch_slack/middleware.py#L397) at the start of the interaction & never updated.

If the selected project is not the default project, this will return an empty set of tags.

# Proposed Solution
When a project is defined/updated in the modal, we should reflect that value in the context so it is available in subsequent interactions.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -281,6 +281,8 @@ def handle_update_incident_project_select_action(
         "selected_option"
     ]["value"]
 
+    context["subject"].project_id = project_id
+
     project = project_service.get(
         db_session=db_session,
         project_id=project_id,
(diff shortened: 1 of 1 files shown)


whitdog47:
Thanks for submitting the fix!

PR #4708, opened 2024-05-09
[Enhancement] - Case participant slack functions, ephemeral message, and escalate bug fix

wssheldon:
### Summary

Adds case participant functionality one would expect in incidents to dedicated channel cases. Such as, adding users as a participant on channel join, invalidaitng the participant, and sending them a welcome message. Fixes a bug in the escalate case command where form_data was not being passed and so priority/type options were not being respected.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/middleware.py
+++ b/src/dispatch/plugins/dispatch_slack/middleware.py
@@ -141,6 +141,7 @@ def engagement_button_context_middleware(
 def action_context_middleware(body: dict, context: BoltContext, next: Callable) -> None:
     """Attempt to determine the current context of the event."""
     private_metadata = json.loads(body["view"]["private_metadata"])
+    print(f"METADATA {private_metadata=}")
     if private_metadata.get("form_data"):
         context.update({"subject": FormMetadata(**private_metadata)})
     else:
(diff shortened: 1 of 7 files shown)


src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -141,6 +141,7 @@ def engagement_button_context_middleware(
 def action_context_middleware(body: dict, context: BoltContext, next: Callable) -> None:
     """Attempt to determine the current context of the event."""
     private_metadata = json.loads(body["view"]["private_metadata"])
+    print(f"METADATA {private_metadata=}")
```
wssheldon, line 144:
```suggestion
```

src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -208,13 +208,6 @@ def is_bot(request: BoltRequest) -> bool:
     if user == "USLACKBOT":
         return True
 
-    authorizations = body.get("authorizations")
```
wssheldon, line 211: This always returned `True`, at least on channel join events.

whitdog47 (approved):

PR #4726, opened 2024-05-13
Change topic in case channel on update

whitdog47:
This PR ensures that the channel topic for dedicated channel cases is updated when details about the case change.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -361,6 +361,16 @@ def case_update_flow(
         # we send the case updated notification
         update_conversation(case, db_session)
 
+    if case.has_channel and case.status != CaseStatus.closed:
+        # determine if case channel topic needs to be updated
+        if (
+            case.case_type.name != previous_case.case_type.name
+            or case.case_severity.name != previous_case.case_severity.name
+            or case.case_priority.name != previous_case.case_priority.name
+            or case.status != previous_case.status
+        ):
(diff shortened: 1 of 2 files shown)


src/dispatch/case/flows.py
```diff
@@ -361,6 +361,16 @@ def case_update_flow(
...
         update_conversation(case, db_session)
 
+    if case.has_channel and case.status != CaseStatus.closed:
+        # determine if case channel topic needs to be updated
+        if (
+            case.case_type.name != previous_case.case_type.name
+            or case.case_severity.name != previous_case.case_severity.name
+            or case.case_priority.name != previous_case.case_priority.name
+            or case.status != previous_case.status
+        ):
```
wssheldon, line 371: I think this comparison is lengthy enough that it would be worth encapsulating it, purely for readability reasons. `e.g.`

```python
def case_details_changed(case: Case, previous_case: Case) -> bool:
    return (
            case.case_type.name != previous_case.case_type.name
            or case.case_severity.name != previous_case.case_severity.name
            or case.case_priority.name != previous_case.case_priority.name
            or case.status != previous_case.status
        )
```

Just a nit though.

wssheldon (approved):
Looks good!

src/dispatch/case/flows.py
whitdog47: Good idea!

PR #4729, opened 2024-05-14
[Escalate] - Do not override user selected inc type with case type mapping

ghost:
Do we actually want to override selected incident type on case escalation with this mapping?

```python
incident_type = case.case_type.incident_type if case.case_type.incident_type else incident_type
```

I don't have context on why this was added so not aware of what the use case was here. I selected Malware as my Incident Type but got Customer Data -- I'd expect it to respect whatever I selected.

Removed this mapping, `incident_type` is required, and what the user selected should be what is used on escalation.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -692,7 +692,7 @@ def case_to_incident_escalate_flow(
     case: Case,
     organization_slug: OrganizationSlug,
     db_session: Session,
-    incident_priority: IncidentType | None,
+    incident_priority: IncidentType,
     incident_type: IncidentPriority | None,
 ):
     if case.incidents:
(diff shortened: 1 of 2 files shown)


whitdog47 (approved):

src/dispatch/case/flows.py
```diff
@@ -692,7 +692,7 @@ def case_to_incident_escalate_flow(
     case: Case,
     organization_slug: OrganizationSlug,
     db_session: Session,
-    incident_priority: IncidentType | None,
+    incident_priority: IncidentType,
     incident_type: IncidentPriority | None,
```
metroid-samus, line 696:
```suggestion
    incident_priority: IncidentPriority | None,
    incident_type: IncidentType,
```

wssheldon: Thanks good catch!

metroid-samus (approved):

PR #4744, opened 2024-05-17
PEP-585 compliance

metroid-samus:
Incremental compliance with PEP-585

- Removed references to typing.Optional, typing.Union, and typing.List
- Former `Optional[List]]` objects default to `[]`


metroid-samus:
Removing lines is causing the code coverage to fall below the accepted threshold.
[PR 4742](https://github.com/Netflix/dispatch/pull/4745) will increase the code coverage.

wssheldon:
```python
class UserBase(DispatchBase):
    email: EmailStr
    - projects: Optional[List[UserProject]] = []
    - organizations: Optional[List[UserOrganization]] = []
    + projects: list[UserProject] = []
    + organizations: list[UserOrganization] = []
```

`Optional[List[UserProject]]` would equate to `list[UserProject] | None` (as opposed to just `list[UserProject]`. I think this will matter especially on the Pydantic models we may pass in `None` and then that is now a runtime error. 

Also, I know this was how it worked before in these models, but default to `None` would be 'more correct' here than the empty list I think (since it's more explicit / semantically correct). I worry that may cause bugs though if we change that in mass and I'm not sure how many places we've crafted business logic around the assumption of empty lists instead of `None`.

PR #4752, opened 2024-05-20
bugfix: Bolt event subscriptions get hijacked by one of the funcs registered

wssheldon:
AFAICT, the bolt SDK will only execute one of the registered functions for a given event subscription? case version was hitting for me locally. This immediately returns if you are not in a case context and thus the incident version is never invoked. This behavior seems kind of strange to me, you would think multiple functions could be registered to the same event, and I'd be curious to see if we can find any mention of this in the bolt documentation.

Either way, we can create a single function for both contexts. I kept it in incident for now since we don't have a spot for these shared functions at this very moment. We could add shared/subject folder or similar.

**Incident Participant Added**

Screenshot 2024-05-20 at 4 36 34 PM

**Case Participant Added**

Screenshot 2024-05-20 at 4 39 39 PM

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/incident/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/incident/interactive.py
@@ -28,6 +28,8 @@ from slack_sdk.web.client import WebClient
 from sqlalchemy.orm import Session
 
 from dispatch.auth.models import DispatchUser
+from dispatch.case import service as case_service
+from dispatch.case import flows as case_flows
 from dispatch.config import DISPATCH_UI_URL
 from dispatch.database.service import search_filter_sort_paginate
 from dispatch.enums import Visibility, EventType, SubjectNames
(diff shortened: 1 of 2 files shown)


src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -994,20 +996,81 @@ def handle_member_joined_channel(
...
-        # only run this workflow for incidents
-        return
+        incident_flows.incident_remove_participant_flow(
+            user.email, context["subject"].id, db_session=db_session
+        )
 
-    incident_flows.incident_remove_participant_flow(
-        user.email, context["subject"].id, db_session=db_session
-    )
+    if context["subject"].type != CaseSubjects.case:
```
whitdog47:
```suggestion
    if context["subject"].type == CaseSubjects.case:
```

metroid-samus (approved):

whitdog47 (changes requested):

whitdog47 (approved):

whitdog47:
Thank you for jumping on this so quickly!

PR #4793, opened 2024-06-03
Adds fractional seconds and Unix timestamp pasting to advanced date picker

whitdog47:
This PR allows for entering seconds and fractional seconds in date-time pickers. It also allows pasting a Unix timestamp.

image


src/dispatch/static/dispatch/src/components/PreciseDateTimePicker.vue
```diff
@@ -0,0 +1,334 @@
...
+    },
+    handlePaste(event) {
+      event.preventDefault()
+      const clipboardData = event.clipboardData || window.clipboardData
+      let pastedData = clipboardData.getData("Text")
+      if (!pastedData.endsWith("Z")) {
+        pastedData += "Z"
+      }
+      // check to see if pastedData is a valid Unix timestamp
+      if (isIsoDate(pastedData)) {
```
jschroth, line 203: When I think of Unix Timestamp, I think of something like `1717457347.123` and it has no timezone - it is based on unix epoch (number of seconds since 01-Jan-1970). 

It seems like this is expecting one to use the ISO string format which would have a timezone as part of the string, but this only handles the UTC string of `Z` and not `+|-##:##`. If we were to support the ISO format, I am fine with that, but I would lean towards being expected to provide the timezone and have Dispatch parse it and interpret it correctly.

src/dispatch/static/dispatch/src/components/PreciseDateTimePicker.vue
```diff
@@ -0,0 +1,334 @@
...
+            />
+          </v-col>
+        </v-row>
+        <div class="text-caption font-weight-light">{{ timezone || "UTC" }}</div>
+      </v-container>
+      <div class="mt-3 mb-3">Or paste a Unix timestamp:</div>
+      <v-text-field
+        class="paste-only"
+        v-model="unixTimestamp"
+        label="Unix Timestamp (UTC)"
```
jschroth, line 103: This looks like it is a field for ISO Timestamp, not the unix timestamp. If it is for ISO we should not only accept and parse `Z` formats.

src/dispatch/static/dispatch/src/components/PreciseDateTimePicker.vue
```diff
@@ -0,0 +1,334 @@
...
+          <span class="slash-space">,&nbsp;</span>
+          <v-col cols="1" class="pa-0">
+            <v-text-field
+              v-model="hour"
+              class="m-0"
+              variant="plain"
+              density="compact"
+              hide-details="true"
+              maxlength="2"
+              @update:model-value="filterNumeric('year', $event)"
```
jschroth, line 60: is this right? `filterNumeric('year'` for the `hour` field?

src/dispatch/static/dispatch/src/components/PreciseDateTimePicker.vue
```diff
@@ -0,0 +1,334 @@
...
+          <span class="colon-space">:</span>
+          <v-col cols="1" class="pa-0">
+            <v-text-field
+              v-model="minutes"
+              class="m-0"
+              variant="plain"
+              density="compact"
+              hide-details="true"
+              maxlength="2"
+              @update:model-value="filterNumeric('year', $event)"
```
jschroth, line 75: is this right?

src/dispatch/static/dispatch/src/components/PreciseDateTimePicker.vue
```diff
@@ -0,0 +1,334 @@
...
+          <span class="colon-space">:</span>
+          <v-col cols="3" class="pa-0">
+            <v-text-field
+              v-model="seconds"
+              class="m-0"
+              variant="plain"
+              density="compact"
+              hide-details="true"
+              maxlength="6"
+              @update:model-value="filterNumeric('year', $event)"
```
jschroth, line 90: is this right?

jschroth (approved):
Thanks!

PR #4826, opened 2024-06-12
Fixes participant announcement message

mvilanova:
The message always shows `Incident Role(s)` even if it's a case participant. This PR changes it to make it dependent on the subject type.
Screenshot 2024-06-12 at 11 59 47 AM


whitdog47 (approved):
Thanks!

PR #4849, opened 2024-06-18
Set owner column as optional when exporting timeline

whitdog47:
This PR allows the Owner field in the timeline export to be optional. Column widths in the export are adjusted accordingly.

image


jneelamekam (approved):
Looks good to me, thank you for working on this.

PR #4866, opened 2024-06-21
bugfix: reporter is optional on update flow, reporter can be None

ghost:
### Resolves:

```bash
AttributeError
"'NoneType' object has no attribute 'individual'"
```

### From:

```python
    case_flows.case_update_flow(
        case_id=case.id,
        previous_case=previous_case,
        db_session=db_session,
        reporter_email=case.reporter.individual.email, <-------
        assignee_email=assignee_email,
        organization_slug=context["subject"].organization_slug,
    )
```

### Solution:

Reporter is optional on cases, it can be `None`.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -360,12 +360,20 @@ def case_update_flow(
 
     if case.tactical_group:
         # we update the tactical group
-        for group_member in [reporter_email, assignee_email]:
+        if reporter_email:
             group_flows.update_group(
                 subject=case,
                 group=case.tactical_group,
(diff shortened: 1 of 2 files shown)


wssheldon:
May need to be handled here too:

```python
    if case.tactical_group:
        # we update the tactical group
        for group_member in [reporter_email, assignee_email]:
            group_flows.update_group(
                subject=case,
                group=case.tactical_group,
                group_action=GroupAction.add_member,
                group_member=group_member,
                db_session=db_session,
            )
```

whitdog47 (approved):

mvilanova (approved):

PR #4874, opened 2024-06-24
Adding linting to migrations

kevgliss:
This PR adds a new type of linting for our alembic migration files.  It dumps all migrations to .sql files and then runs a subset of linting rules that looks for problematic locks and other issues. Currently, it only attempts to look at the last added revision, as we generally only have one migration per pr; I think this is probably reasonable. If we find value we can extend this a bit. 

Documentation reference: https://kaveland.no/eugene/introduction.html

Additionally, I have added several improvements to the actual migrations themselves based on the advice here: 
https://squawkhq.com/docs/web-frameworks/#alembic-and-sqlalchemy. This should help prevent migrations from locking up the database if we accidentally introduce a change that requires exclusivity.

### DIFF
--- a/.github/workflows/lint-migrations.yml
+++ b/.github/workflows/lint-migrations.yml
@@ -37,6 +37,7 @@ jobs:
           chmod +x eugene
       - name: Generate sql files
         run: |
+          export DISPATCH_ENCRYPTION_KEY=supersecret
           dispatch database upgrade --revision-type tenant  --sql
       - name: Lint files
         run: ./eugene lint alembic_output.sql -f md --accept-failures > lint.md
--- a/.github/workflows/lint-migrations.yml
+++ b/.github/workflows/lint-migrations.yml
@@ -15,10 +15,10 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python 3.11
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: 3.11.2
-      - uses: actions/cache@v2
+      - uses: actions/cache@v4
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-migrations.yml
```diff
@@ -0,0 +1,73 @@
...
+
+jobs:
+  lint-migration:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v2
```
wssheldon, line 18:
```suggestion
        uses: actions/setup-python@v5
```

Is the latest version

.github/workflows/lint-migrations.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          python-version: 3.11.2
+      - uses: actions/cache@v2
```
wssheldon, line 21:
```suggestion
      - uses: actions/cache@v4
```

Latest

src/dispatch/cli.py
```diff
@@ -405,32 +407,41 @@ def upgrade_database(tag, sql, revision, revision_type):
     from .database.manage import init_database
 
     alembic_cfg = AlembicConfig(config.ALEMBIC_INI_PATH)
+    path = None
```
whitdog47, line 410: It looks like the playwright tests are failing since the `path` is still `None` and line `set_main_option` in line 418 requires a string.

whitdog47 (changes requested):

src/dispatch/cli.py
```diff
@@ -405,32 +407,41 @@ def upgrade_database(tag, sql, revision, revision_type):
...
-        if "dispatch_core" not in schema_names:
-            click.secho("Detected single tenant database, converting to multi-tenant...")
-            conn.execute(sqlalchemy.text(open(config.ALEMBIC_MULTI_TENANT_MIGRATION_PATH).read()))
+        elif revision_type == "tenant":
+            path = config.ALEMBIC_TENANT_REVISION_PATH
 
-        if revision_type:
-            if revision_type == "core":
-                path = config.ALEMBIC_CORE_REVISION_PATH
+    alembic_cfg.set_main_option("script_location", path)
```
metroid-samus, line 418:
```suggestion
        alembic_cfg.set_main_option("script_location", path)

PR #4884, opened 2024-06-26
Adds additional bot check.

metroid-samus:
Adds additional check for is_bot flag in user info

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/middleware.py
+++ b/src/dispatch/plugins/dispatch_slack/middleware.py
@@ -232,7 +232,7 @@ def user_middleware(
     payload: dict,
 ) -> None:
     """Attempts to determine the user making the request."""
-    if is_bot(request) or client.users_info(user=user_id)["user"]["is_bot"]:
+    if is_bot(request):
         return context.ack()
 
     user_id = None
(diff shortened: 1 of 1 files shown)


src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -284,7 +284,12 @@ def user_middleware(
             user_in=UserRegister(email=participant.individual.email),
         )
     else:
-        email = client.users_info(user=user_id)["user"]["profile"]["email"]
+        user_info = client.users_info(user=user_id)["user"]
```
jschroth, line 287:
```suggestion
        user_info = client.users_info(user=user_id).get("user", {})
```

src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -284,7 +284,12 @@ def user_middleware(
...
+
+        if user_info["is_bot"]:
```
jschroth, line 289:
```suggestion
        if user_info.get("is_bot", False):
```

src/dispatch/plugins/dispatch_slack/middleware.py
```diff
@@ -284,7 +284,12 @@ def user_middleware(
...
+            return context.ack()
+
+        email = user_info["profile"]["email"]
```
jschroth, line 292:
```suggestion
        email = user_info.get("profile", {}).get("email")
```

jschroth (approved):
nit: trying to prevent errors when the response is not formatted the way we expect. It will still error, but it will error on the ContextError and not the None does not have...

PR #4899, opened 2024-06-28
Create case costs, case cost types, and case cost models

metroid-samus:
Creates models associated with case costs.


src/dispatch/database/revisions/tenant/versions/2024-06-28_15a8d3228123.py
```diff
@@ -0,0 +1,128 @@
...
+        "entity",
+        ["search_vector"],
+        unique=False,
+        postgresql_using="gin",
+    )
+    op.alter_column("entity_type", "jpath", existing_type=sa.VARCHAR(), nullable=True)
+    op.add_column("participant_activity", sa.Column("case_id", sa.Integer(), nullable=True))
+    op.create_foreign_key(None, "participant_activity", "case", ["case_id"], ["id"])
+    op.drop_column("plugin_instance", "configuration")
+    op.drop_constraint("project_stable_priority_id_fkey", "project", type_="foreignkey")
```
wssheldon, line 88: Are all these schema changes correct / related to the current changes? Some that look out of place are:

```python
    op.alter_column(
        "entity", "source", existing_type=sa.BOOLEAN(), type_=sa.String(), existing_nullable=True
    )
    op.drop_index("ix_entity_search_vector", table_name="entity", postgresql_using="gin")
    op.create_index(
        "entity_search_vector_idx",
        "entity",
        ["search_vector"],
        unique=False,
        postgresql_using="gin",
    )
    op.alter_column("entity_type", "jpath", existing_type=sa.VARCHAR(), nullable=True)
```

```python
    op.drop_column("plugin_instance", "configuration")
    op.drop_constraint("project_stable_priority_id_fkey", "project", type_="foreignkey")
```

wssheldon (changes requested):
Requesting change on the alembic migration question, just to make sure those are expected

src/dispatch/case/type/service.py
```diff
@@ -148,6 +160,13 @@ def create(*, db_session, case_type_in: CaseTypeCreate) -> CaseType:
 
 def update(*, db_session, case_type: CaseType, case_type_in: CaseTypeUpdate) -> CaseType:
     """Updates a case type."""
+    cost_model = None
+    if case_type_in.cost_model:
+        cost_model = cost_model_service.get_cost_model_by_id(
+            db_session=db_session, cost_model_id=case_type_in.cost_model.id
+        )
+    case_type.cost_model = cost_model
+
```
whitdog47, line 169: incident/type/service.py also includes a check for `should_update_incident_cost` - should we do that here too?

metroid-samus: Oh. it looks like i accidentally added the service.py file to this PR.
It's meant to be in https://github.com/Netflix/dispatch/pull/4900

whitdog47 (approved):

wssheldon (approved):

PR #4900, opened 2024-06-28
Enable backend functionality of calculating case costs.

metroid-samus:
Adds backend functionality for cases to opt-in to a case cost calculation feature. If no cost models are associated with this type of case, no costs are calculated.

Depends on `enhancement/case-cost-model`


src/dispatch/case/service.py
```diff
@@ -72,6 +73,16 @@ def get_all(*, db_session, project_id: int) -> List[Optional[Case]]:
     return db_session.query(Case).filter(Case.project_id == project_id)
 
 
+def get_all_open_by_case_type(*, db_session, case_type_id: int) -> List[Optional[Case]]:
+    """Returns all non-closed cases based on the given case type."""
+    return (
+        db_session.query(Case)
+        .filter(Case.status != CaseStatus.closed)
```
whitdog47, line 80: Add escalated here as well

src/dispatch/case_cost/service.py
```diff
@@ -0,0 +1,416 @@
...
+
+    # we calculate the number of hours the participant has spent in the case role
+    participant_role_time_hours = participant_role_time.total_seconds() / SECONDS_IN_HOUR
+
+    # we make the assumption that participants only spend 8 hours a day working on the case,
+    # if the case goes past 24hrs
+    # TODO(mvilanova): adjust based on case priority
+    if participant_role_time_hours > HOURS_IN_DAY:
+        days, hours = divmod(participant_role_time_hours, HOURS_IN_DAY)
+        participant_role_time_hours = math.ceil(((days * HOURS_IN_DAY) / 3) + hours)
```
metroid-samus, line 332: Delete this. Unneeded since we are calculating based on activity

whitdog47 (changes requested):

src/dispatch/case_cost/scheduled.py
```diff
@@ -0,0 +1,68 @@
...
+            # we don't need to update the cost of closed cases if they already have a response cost and this was updated after the case was closed
+            if case.status == CaseStatus.closed:
+                if case_response_cost:
+                    if case_response_cost.updated_at > case.closed_at:
+                        continue
+            # we don't need to update the cost of escalated cases if they already have a response cost and this was updated after the case was escalated
+            if case.status == CaseStatus.escalated:
+                if case_response_cost:
+                    if case_response_cost.updated_at > case.escalated_at:
+                        continue
```
wssheldon, line 53: Is there a way we can pull less cases here as opposed to all of them to optimize performance? I may be wrong here but couldn't we move these conditions for `closed` and `escalated` to the database query above instead of `get_all`?

src/dispatch/participant_activity/service.py
```diff
@@ -66,6 +70,16 @@ def get_last_participant_activity(
     )
 
 
+def get_all_case_participant_activities_for_case(
+    db_session: SessionLocal,
```
wssheldon, line 74:
```suggestion
    db_session: Session,
```

`SessionLocal` is just `Any` AFAIK.

src/dispatch/case_cost/scheduled.py
metroid-samus: This is to cover cases that were marked `closed` or `escalated` in the time since the last cost update.

wssheldon: Another optimization I think we can make here is selecting only the columns we need to compute the cost as it essentially does all of them currently, I believe.

metroid-samus: [Done](https://github.com/Netflix/dispatch/pull/4900/commits/be30c8d4501537ead81c22b992af05f8afd2fe91#diff-85562e526dc8615124aead4d1b1dc02d9fbfdbc211396e4c880f3e63f354fb00R6)

wssheldon: Should we need to use one of these service functions (`get_all_by_status`) instead of `get_all` above?

metroid-samus: Good point. [Fixed it here](https://github.com/Netflix/dispatch/pull/4900/commits/128a9db2628764a60122d360efc79fd561cab221).

wssheldon: The function call is missing the `status` parameter and my guess is we'll need to pull multiple different status so may need an update to accept a list of statuses.

metroid-samus: Done.

whitdog47 (approved):

PR #4907, opened 2024-07-02
feat: add self-join button and api to cases

wssheldon:
Screenshot 2024-07-02 at 10 28 45 AM

You can join dedicated case channels, if it is not closed. You cannot join non-dedicated case channels and there is no concept of 'subscribe'

### DIFF
--- a/src/dispatch/auth/permissions.py
+++ b/src/dispatch/auth/permissions.py
@@ -444,9 +444,9 @@ class CaseJoinPermission(BasePermission):
         request: Request,
     ) -> bool:
         pk = PrimaryKeyModel(id=request.path_params["case_id"])
-        current_incident = incident_service.get(db_session=request.state.db, incident_id=pk.id)
+        current_case = case_service.get(db_session=request.state.db, incident_id=pk.id)
 
-        if current_incident.visibility == Visibility.restricted:
+        if current_case.visibility == Visibility.restricted:
             return OrganizationAdminPermission(request=request)
 
         return True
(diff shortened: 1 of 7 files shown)


src/dispatch/auth/permissions.py
```diff
@@ -438,6 +438,20 @@ def has_required_permissions(
         return current_user.email in participant_emails
 
 
+class CaseJoinPermission(BasePermission):
+    def has_required_permissions(
+        self,
+        request: Request,
+    ) -> bool:
+        pk = PrimaryKeyModel(id=request.path_params["case_id"])
+        current_incident = incident_service.get(db_session=request.state.db, incident_id=pk.id)
```
mvilanova, line 447: Shouldn't we fetch the case not the incident here?

src/dispatch/case/views.py
```diff
@@ -338,3 +339,25 @@ def delete_case(
                 }
             ],
         ) from None
+
+
+@router.post(
+    "/{case_id}/join",
+    summary="Adds an individual to an case.",
```
mvilanova, line 346:
```suggestion
    summary="Adds an individual to a case.",
```

src/dispatch/case/views.py
```diff
@@ -338,3 +339,25 @@ def delete_case(
...
+    dependencies=[Depends(PermissionsDependency([CaseJoinPermission]))],
+)
+def join_case(
```
mvilanova, line 349: Lots of references to incident in this block

mvilanova (changes requested):

wssheldon:
@mvilanova Ready for review now if you'd like to take another look, I took a pass.

mvilanova (approved):

PR #4908, opened 2024-07-02
Provides option of an alternative folder structure

whitdog47:
This PR allows for an alternative folder structure when creating the incident storage, with options defined per project. The following options are added:
(1) Allows for setting the names of the two storage folders created (current defaults are "Logs" and "Screengrabs")
(2) Allows the main storage folder of the incident and review documents to be put under one of these folders
(3) Allows the title of the storage folder to be named with the name of the incident instead of the incident number
(4) Allows for the main storage folder location to be defined by the uri of the tag set when the incident is created
(5) Allows the incident and review document templates to contain tag type placeholders in the form {{tag_tagType}}

image


wssheldon:
He wont stop shipping!!!

metroid-samus:
After enabling the alternative folder structure, I discovered the following:

Create a new Incident without tags:
- In the incident's `Resources` tab, the `Storage` link takes the user to the Primary folder. I was expecting it to take me to the parent folder that shares the name of the incident. If this is the expected behavior, I would recommend also adding the Secondary folder to the `Resources` tab.
- The incident document is created, but it is not listed under the `Resources` tab.

Create a new Incident with tags:
- It is not intuitive that the tag's URI should be a Google Drive folder ID and not a URL.

PR #4913, opened 2024-07-03
Skip retries for defined SlackAPIErrorCode

jschroth:
Does not retry for any item in the SlackAPIErrorCode enum

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/service.py
+++ b/src/dispatch/plugins/dispatch_slack/service.py
@@ -103,11 +103,11 @@ def should_retry(exception: Exception) -> bool:
     """
     match exception:
         case SlackApiError():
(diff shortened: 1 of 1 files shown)


whitdog47 (approved):
Thanks!

wssheldon:
You should be able to simplify this down to:

```python
def should_retry(exception: Exception) -> bool:
    match exception:
        case SlackApiError():
            return exception.response["error"] not in SlackAPIErrorCode.__members__.values()
        case TimeoutError() | Timeout():
            return True
        case _:
            return False
```

wssheldon:
We could also implement `__contains__` on `DispatchEnum` to allow `exception.response["error"] not in SlackAPIErrorCode`

PR #4940, opened 2024-07-10
feat(signals): FIFO queue on signal processing, refactor into individual functions, other minor improvements

ghost:
A similar implantation to what was thoroughly tested here. https://github.com/Netflix/dispatch/pull/3337


whitdog47 (approved):
So much cleaner and easier to read

PR #4949, opened 2024-07-11
Do not overwrite description field

metroid-samus:
This PR fixes the following bug:
* When escalating a case via Slack, the user’s incident description is overwritten with the case description

How to test:
* In Slack, Report a case
* In the resulting case channel, type `/dispatch-escalate-case'
* Modify the description field and submit
* The new incident description should have the updated information as a prefix."


wssheldon (approved):
Looks good.

PR #4958, opened 2024-07-16
Adds form scoring based on scoring schema

whitdog47:
This PR uses the form type scoring schema to calculate a risk score based on the completed form data.

image


src/dispatch/forms/scoring.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for s in scoring:
+        # get the value of the form data based on the key
+        var = form_vals.get(s.get("var"))
+        if var:
+            includes = s.get("includes")
+            if includes:
+                # if any element in the form data is in the includes list, add the score
+                if any(v in includes for v in var):
+                    score += s.get("score", 0)
```
jschroth, line 34: nit
```suggestion
        if (var := form_vals.get(s.get("var"))) and (includes := s.get("includes")):
            # if any element in the form data is in the includes list, add the score
            if any(v in includes for v in var):
                score += s.get("score", 0)
```

jschroth (approved):

PR #4998, opened 2024-07-22
Allows tagging of runbooks

whitdog47:
This PR allows users to tag runbooks. Note that desired tag types must have the "Document" discoverability checked.

image


wssheldon:
```vue
              <v-col cols="12">
                <tag-picker
                  label="Tags"
                  v-model="tags"
                  :project="project"
                  model="document"
                  :model-id="id"
                  show-copy
                />
              </v-col>
```

Does it only show valid elements (e.g. `Note that desired tag types must have the "Document" discoverability checked`)? -- Best case, I think, we only show valid elements, and there's an `info` tooltip that says what elements are shown.

whitdog47:
@wssheldon yes, the `model` prop here tells the tag picker which tag type discoverability model to use.

jschroth (approved):

PR #5001, opened 2024-07-23
Fixes issue with join case button

whitdog47:


src/dispatch/messaging/strings.py
```diff
@@ -415,7 +453,7 @@ class MessageType(DispatchEnum):
     "buttons": [
         {
             "button_text": "Join",
-            "button_value": "{{organization_slug}}-{{incident_id}}",
```
whitdog47, line 418: @jschroth here is the real fix - the rest were autoformatting

src/dispatch/messaging/strings.py
```diff
@@ -415,7 +453,7 @@ class MessageType(DispatchEnum):
...
+            "button_value": "{{organization_slug}}-{{case_id}}",
```
whitdog47, line 456: @jschroth and this one

src/dispatch/messaging/strings.py
```diff
@@ -428,7 +466,7 @@ class MessageType(DispatchEnum):
...
```
whitdog47, line 469: @jschroth and this one

jschroth (approved):

PR #5004, opened 2024-07-24
Update endpoints.py

shubham-1997:


mvilanova:
@shubham-1997 can you share context on what the goal of this pull request is?

PR #5021, opened 2024-07-26
bugfix(case/ui): some escalate dialog resources never exit load state

ghost:
Screenshot 2024-07-26 at 11 12 08 AM


jschroth (approved):
approveprs

PR #5032, opened 2024-08-01
Add logging for case status transitions

jschroth:
Adding logging to debug when some cases do not get a triaged_at time.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -542,7 +542,7 @@ def case_status_transition_flow_dispatcher(
     db_session: Session,
 ):
     """Runs the correct flows based on the current and previous status of the case."""
-    log.info(f"Transitioning Status for Case {case.id} from {previous_status} to {current_status}")
+    log.info(f"Transitioning Status for Case #{case.id} from {previous_status} to {current_status}")
     match (previous_status, current_status):
         case (CaseStatus.closed, CaseStatus.new):
             # Closed -> New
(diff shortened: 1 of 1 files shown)


metroid-samus (approved):

src/dispatch/case/flows.py
```diff
@@ -542,6 +542,7 @@ def case_status_transition_flow_dispatcher(
     db_session: Session,
 ):
     """Runs the correct flows based on the current and previous status of the case."""
+    log.info(f"Transitioning Status for Case {case.id} from {previous_status} to {current_status}")
```
jschroth, line 545:
```suggestion
    log.info(f"Transitioning Status for Case #{case.id} from {previous_status} to {current_status}")
```

src/dispatch/case/flows.py
```diff
@@ -568,6 +569,7 @@ def case_status_transition_flow_dispatcher(
 
         case (_, CaseStatus.triage):
             # Any -> Triage/
+            log.warning(f"Unexpected Status Transition for Case #{case.id} from {previous_status} to {current_status}")
             pass
```
jschroth, line 574:
```suggestion

```
don't need to `pass` now

src/dispatch/case/flows.py
```diff
@@ -568,8 +569,7 @@ def case_status_transition_flow_dispatcher(
 
         case (_, CaseStatus.triage):
             # Any -> Triage/
-            pass
-
+            log.warning(f"Unexpected Status Transition for Case #{case.id} from {previous_status} to {current_status}")
```
jschroth, line 572:
```suggestion
            log.warning(f"Unexpected Status Transition for Case #{case.id} from {previous_status} to {current_status}")

```

PR #5037, opened 2024-08-05
Adds detection lifecyle to signal model

mvilanova:


src/dispatch/signal/models.py
```diff
@@ -311,6 +312,7 @@ class SignalBase(DispatchBase):
     project: ProjectRead
     source: Optional[SourceBase]
     variant: Optional[str]
+    lifecycle: Optional[str]
```
kevgliss, line 315: Should we set a default lifecycle here? Are we hoping to restrict the possible lifecycles? If so an `enum` might be warranted.

mvilanova: Since this will be propagated from Geiger where we already have an enum to restrict possible values and the value in Dispatch should be readonly I'm not sure it's necessary.

src/dispatch/static/dispatch/src/signal/NewEditDialog.vue
```diff
@@ -100,6 +100,16 @@
...
                   />
                 </v-col>
+                <v-col cols="12">
+                  <v-text-field
+                    v-model="lifecycle"
+                    label="Lifecycle"
+                    hint="The lifecycle stage of the signal."
+                    persistent-hint
+                    name="lifecycle"
+                    :rules="[rules.required]"
```
wssheldon, line 110: I don't think it should be required if it's nullable, right?

wssheldon: (I don't actually know) -- Just worried about the bug where you can't save the edit sheet because of the required rule.

mvilanova: This is expected to be populated by Geiger, so it should always have a value, and it shouldn't be edited manually. I could remove `:rules="[rules.required]"` and add `readonly `.

wssheldon (approved):

src/dispatch/static/dispatch/src/signal/NewEditDialog.vue
mvilanova: I added `readonly` and kept `:rules="[rules.required]"`.

mvilanova: I did it locally but didn't push the changes upstream lol. I'll add the changes in a separate PR.

PR #5046, opened 2024-08-08
Moving Slack button update to after case escalation flow

whitdog47:
When a case escalation is in progress, it's possible that a user could click on "Join Incident" in the updated conversation thread before the incident is created, causing an issue when trying to retrieve the associated incident. This PR fixes that issue by moving the conversation update code block to after incident creation.


wssheldon (approved):

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -240,8 +240,14 @@ def handle_update_case_command(
...
         resolution_input(initial_value=case.resolution),
         assignee_select(initial_user=assignee_initial_user),
-        case_status_select(initial_option={"text": case.status, "value": case.status}, statuses=statuses),
-        Context(elements=["Cases cannot be escalated here. Please use the `/dispatch-escalate-case` slash command."]),
+        case_status_select(
+            initial_option={"text": case.status, "value": case.status}, statuses=statuses
+        ),
+        Context(
+            elements=[
+                "Cases cannot be escalated here. Please use the `/dispatch-escalate-case` slash command."
```
mvilanova, line 248: I wonder if we could do `SlackConversationConfiguration().slack_command_escalate_case` instead of hardcoding the name of the command.

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -1250,6 +1242,20 @@ def handle_escalation_submission_event(
...
+    blocks = create_case_message(case=case, channel_id=context["subject"].channel_id)
+    if case.has_thread:
+        client.chat_update(
+            blocks=blocks,
+            ts=case.conversation.thread_id,
+            channel=case.conversation.channel_id,
+        )
+
+    client.chat_postMessage(
+        text="This case has been escalated to an incident. All further triage work will take place in the incident channel.",
```
mvilanova, line 1254: We may want to include the name of the incident channel in the message.

PR #5074, opened 2024-08-13
Create AI-generated weekly summary

whitdog47:
This PR adds a weekly AI-generated summary of closed incidents. Requires an enabled `artificial-intelligence` plugin.


src/dispatch/incident/scheduled.py
```diff
@@ -222,3 +226,105 @@ def incident_close_reminder(db_session: SessionLocal, project: Project):
...
+    )
+
+    # no incidents closed in the last week
+    if not incidents:
+        return
+
+    storage_plugin = plugin_service.get_active_instance(
+        db_session=db_session, plugin_type="storage", project_id=project.id
+    )
+
```
mvilanova, line 264: You may want to check that you have a storage_plugin here

mvilanova (approved):
Nice seeing the ai plugin being used! I think it would be great if we could generate summaries for cases as well in the future.

src/dispatch/incident/scheduled.py
whitdog47: Good thinking. Fixed.

mvilanova (approved):

PR #5093, opened 2024-08-16
Truncate the case resolution reason text due to Slack limit

whitdog47:
Slack has a 3000 character limit on section text, so we're truncating the resolution text in case it is longer.

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/case/messages.py
+++ b/src/dispatch/plugins/dispatch_slack/case/messages.py
@@ -38,6 +38,7 @@ from dispatch.signal.models import (
     assoc_signal_instance_entities,
 )
 from dispatch.signal.enums import SignalEngagementStatus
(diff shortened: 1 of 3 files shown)


wssheldon:
It looks like you can set a `max_length` on the plain text input field, which I assume will pop an error if you exceed the limit, and prevent you from submitting. [0]

Do you think that UX would be preferable? 

- [0] [Plain-text input element ](https://api.slack.com/reference/block-kit/block-elements#input:~:text=The%20maximum%20length%20of%20input%20that%20the%20user%20can%20provide.%20If%20the%20user%20provides%20more%2C%20they%20will%20receive%20an%20error.%20Acceptable%20values%20for%20this%20field%20are%20between%201%20and%203000%2C%20inclusive.)

whitdog47:
@wssheldon porque no los dos?

mvilanova:
LGTM. My suggestion would be store the 3000 in a constant to prevent things from getting out of sync across the codebase.

wssheldon (approved):

PR #5113, opened 2024-08-21
Adding some better error handling to sqs consume

kevgliss:


wssheldon (approved):
We may want to log on any exception as well

PR #5128, opened 2024-08-26
fix(plugin): delete non-conforming messages from sqs queue

ghost:
Messages in the queue that do not conform should not stay in the queue

### DIFF
--- a/src/dispatch/plugins/dispatch_aws/plugin.py
+++ b/src/dispatch/plugins/dispatch_aws/plugin.py
@@ -73,6 +73,10 @@ class AWSSQSSignalConsumerPlugin(SignalConsumerPlugin):
                     log.exception(
                         f"Received signal instance that does not conform to `SignalInstanceCreate` structure, skipping creation: {e}"
                     )
+                    # Add the message to be deleted even if it doesn't conform
+                    entries.append(
+                        {"Id": message["MessageId"], "ReceiptHandle": message["ReceiptHandle"]}
+                    )
                     continue
 
                 try:
--- a/src/dispatch/plugins/dispatch_aws/plugin.py
+++ b/src/dispatch/plugins/dispatch_aws/plugin.py
@@ -70,7 +70,7 @@ class AWSSQSSignalConsumerPlugin(SignalConsumerPlugin):
                         project=project, raw=signal_data, **signal_data
                     )
                 except ValidationError as e:
-                    log.exception(
+                    log.warning(
                         f"Received signal instance that does not conform to `SignalInstanceCreate` structure, skipping creation: {e}"
                     )
                     # Add the message to be deleted even if it doesn't conform


kevgliss:
Is there a reason why you'd rather do this instead of letting them fail a few times and using the DLQ functionality?

wssheldon:
> Is there a reason why you'd rather do this instead of letting them fail a few times and using the DLQ functionality?

My thinking is that non-conforming instances shouldn't be retried since we are certain they will not work and can safely be immediately deleted. We would not want to re-process them either.

src/dispatch/plugins/dispatch_aws/plugin.py
```diff
@@ -73,6 +73,10 @@ def consume(self, db_session: Session, project: Project) -> None:
                     log.exception(
```
whitdog47, line 73: Should we change this to a warning then?

kevgliss:
> > Is there a reason why you'd rather do this instead of letting them fail a few times and using the DLQ functionality?
> 
> My thinking is that non-conforming instances shouldn't be retried since we are certain they will not work and can safely be immediately deleted. We would not want to re-process them either.

I guess it depends on how non-conforming they are. The nice thing about the DLQ is that if the error is on the Dispatch side, we can push a change and then re-drive all of the messages.

src/dispatch/plugins/dispatch_aws/plugin.py
wssheldon: True, changed that.

wssheldon:
> > > Is there a reason why you'd rather do this instead of letting them fail a few times and using the DLQ functionality?
> > 
> > 
> > My thinking is that non-conforming instances shouldn't be retried since we are certain they will not work and can safely be immediately deleted. We would not want to re-process them either.
> 
> I guess it depends on how non-conforming they are. The nice thing about the DLQ is that if the error is on the Dispatch side, we can push a change and then re-drive all of the messages.

Would it make sense to only re-drive on general exceptions in the `create_signal_instance` function?

```python
                try:
                    signal_instance = signal_service.create_signal_instance(
                        db_session=db_session,
                        signal_instance_in=signal_instance_in,
                    )
                except Exception as e:
                    log.exception(f"Unable to create signal instance: {e}")
                    continue
```

The schema is the schema is my thinking and any Validation errors we can safely discard. That's where my head's at but happy to go either direction if you think it makes sense.

kevgliss:
> Would it make sense to only re-drive on general exceptions in the `create_signal_instance` function?
> 
> ```python
>                 try:
>                     signal_instance = signal_service.create_signal_instance(
>                         db_session=db_session,
>                         signal_instance_in=signal_instance_in,
>                     )
>                 except Exception as e:
>                     log.exception(f"Unable to create signal instance: {e}")
>                     continue
> ```
> 
> The schema is the schema is my thinking and any Validation errors we can safely discard. That's where my head's at but happy to go either direction if you think it makes sense.

Yeah, the failure mode I was thinking about was things like the underlying database not having the correct migrations. This would cause `create` to fail, but the messages should still be re-processed.

src/dispatch/plugins/dispatch_aws/plugin.py
```diff
@@ -70,9 +70,13 @@ def consume(self, db_session: Session, project: Project) -> None:
...
+                    log.warning(
                         f"Received signal instance that does not conform to `SignalInstanceCreate` structure, skipping creation: {e}"
                     )
+                    # Add the message to be deleted even if it doesn't conform
+                    entries.append(
+                        {"Id": message["MessageId"], "ReceiptHandle": message["ReceiptHandle"]}
+                    )
                     continue
 
                 try:
```
wssheldon, line 82: @kevgliss I think we may need to call rollback on the `db_session` in these exception cases as well?

wssheldon:
@kevgliss I agree this is no longer needed with a DQL setup.

PR #5140, opened 2024-08-28
Case cost model - frontend

metroid-samus:
This PR adds the ability to create case costs by assigning a cost model to case types. All cases of the same type use the same cost model to calculate ongoing response costs.

The final cost of a case is an accumulation of one or multiple cost types.
Screenshot 2024-08-28 at 8 07 30 PM 
Screenshot 2024-08-28 at 8 17 28 PM

Each case type can be assigned a cost model, which are used to calculate ongoing response costs. If no cost model is assigned to a case type, all cases of that same type do not update their response costs.
Screenshot 2024-08-28 at 8 10 44 PM

### DIFF
--- a/src/dispatch/case_cost/views.py
+++ b/src/dispatch/case_cost/views.py
@@ -20,7 +20,6 @@ router = APIRouter()
 @router.get("", response_model=CaseCostPagination)
 def get_case_costs(common: CommonParameters):
     """Get all case costs, or only those matching a given search term."""
-    print("get case costs")
     return search_filter_sort_paginate(model="CaseCost", **common)
 
 
--- a/src/dispatch/case_cost/views.py
+++ b/src/dispatch/case_cost/views.py
@@ -25,7 +25,7 @@ def get_case_costs(common: CommonParameters):
 
 @router.get("/{case_cost_id}", response_model=CaseCostRead)
 def get_case_cost(db_session: DbSession, case_cost_id: PrimaryKey):
-    """Get an case cost by its id."""
+    """Get a case cost by its id."""
     case_cost = get(db_session=db_session, case_cost_id=case_cost_id)
     if not case_cost:
         raise HTTPException(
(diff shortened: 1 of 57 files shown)


src/dispatch/case_cost/views.py
```diff
@@ -0,0 +1,88 @@
...
+from .service import create, delete, get, update
+
+
+router = APIRouter()
+
+
+@router.get("", response_model=CaseCostPagination)
+def get_case_costs(common: CommonParameters):
+    """Get all case costs, or only those matching a given search term."""
+    print("get case costs")
```
wssheldon, line 23:
```suggestion
```

wssheldon (approved):

src/dispatch/case_cost/views.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@router.get("/{case_cost_id}", response_model=CaseCostRead)
+def get_case_cost(db_session: DbSession, case_cost_id: PrimaryKey):
+    """Get an case cost by its id."""
+    case_cost = get(db_session=db_session, case_cost_id=case_cost_id)
+    if not case_cost:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=[{"msg": "An case cost with this id does not exist."}],
```
mvilanova, line 34:
```suggestion
            detail=[{"msg": "A case cost with this id does not exist."}],
```

mvilanova: There are other docstrings with the same issue.

src/dispatch/case_cost_type/views.py
```diff
@@ -0,0 +1,105 @@
...
+
+@router.get("", response_model=CaseCostTypePagination)
+def get_case_cost_types(common: CommonParameters):
+    """Get all case cost types, or only those matching a given search term."""
+    return search_filter_sort_paginate(model="CaseCostType", **common)
+
+
+@router.get("/{case_cost_type_id}", response_model=CaseCostTypeRead)
+def get_case_cost_type(db_session: DbSession, case_cost_type_id: PrimaryKey):
+    """Get an case cost type by its id."""
```
mvilanova, line 28:
```suggestion
    """Get a case cost type by its id."""
```

mvilanova: May want to do a search and replace to fix all of them at once.

metroid-samus: Good catch. Done.

PR #5152, opened 2024-08-30
Check role before reassigning

whitdog47:
This PR makes sure we don't assign a role twice when escalating a case.

### DIFF
--- a/src/dispatch/case/flows.py
+++ b/src/dispatch/case/flows.py
@@ -706,6 +706,14 @@ def common_escalate_flow(
             # Map the case role to an incident role
             incident_role = ParticipantRoleType.map_case_role_to_incident_role(case_roles)
 
+            # If role already assigned, assign as participant
+            if participant_service.get_by_incident_id_and_role(
+                db_session=db_session,
+                incident_id=incident.id,
+                role=incident_role,
+            ):
+                incident_role = ParticipantRoleType.participant
+
             participant_flows.add_participant(
                 participant.individual.email,
                 incident,


metroid-samus (commented):
This would convert all case observers to incident participants which may skew metrics.

src/dispatch/case/flows.py
```diff
@@ -706,6 +706,14 @@ def common_escalate_flow(
             # Map the case role to an incident role
             incident_role = ParticipantRoleType.map_case_role_to_incident_role(case_roles)
```
jschroth, line 707: It seems like the issue is actually here. one can have multiple roles, but this assumes one role. I am also unclear why we pass a list of roles to map and only return one role. Is that expected? It seems like if someone is the reporter, they might get that role here if it is first in the list of case_roles.

Also, the reporter in this case would move to be just a participant.

jschroth: `map_case_role_to_incident_role` is only used here so this seems like an opportunity to improve. Either move the role to a participant (runs the risk of not having an IC if there is no on call service) or If the role comes back as IC, then there can be an additional check if there is already an IC or an IC will be assigned based on an oncall, then remove the IC role from the returned set.

src/dispatch/case/flows.py
```diff
@@ -669,6 +669,29 @@ def send_escalation_messages_for_channel_case(
...
+            if participant_service.get_by_incident_id_and_role(
+                db_session=db_session,
+                incident_id=incident.id,
+                role=ParticipantRoleType.incident_commander,
+            ):
+                incident_roles.add(ParticipantRoleType.participant)
+            else:
+                incident_roles.add(ParticipantRoleType.incident_commander)
+        else:
+            incident_roles.add(ParticipantRoleType.participant)
```
jschroth, line 691: should we default to observer?

whitdog47: I'm not sure. These are folks that were working on the case, so it's less likely they were just acting as observers. Maybe the right way is to add folks as observers and then change to participant once they respond for the first time. I'll write a new ticket for that.

forestmonster: > Add folks as observers and then change to participant once they respond for the first time

Definitely sounds like the right way to do it. :+1:

jschroth (approved):
We can always update the default to observer later, but I’ll let you decide

whitdog47:
Good point @metroid-samus, I changed it to add to the incident the same role type as in the case (participant/observer)

PR #5175, opened 2024-09-06
feat(mfa/plugin): add dispatch mfa plugin

ghost:


src/dispatch/static/dispatch/src/auth/Mfa.vue
```diff
@@ -0,0 +1,106 @@
...
+    </v-row>
+  </v-container>
+</template>
+
+<script setup lang="ts">
+import { ref, computed } from "vue"
+import { useRoute } from "vue-router"
+import authApi from "@/auth/api"
+
+/* eslint-disable no-unused-vars */
```
wssheldon, line 34: Was getting an `unused` here, but it definitely is used? Not sure what's up here. I added it to a type file and exported as const enum and got the same error.

src/dispatch/plugins/dispatch_core/plugin.py
```diff
@@ -273,6 +288,115 @@ def get(
...
+        challenge: MfaChallenge | None = (
+            db_session.query(MfaChallenge)
+            .filter_by(challenge_id=payload.challenge_id)
+            .one_or_none()
+        )
+
+        if not challenge:
+            raise InvalidChallengeError("Invalid challenge ID")
+        if challenge.dispatch_user_id != current_user.id:
+            raise UserMismatchError("Challenge does not belong to the current user")
```
mvilanova, line 377:
```suggestion
            raise UserMismatchError(f"Challenge does not belong to the current user: {current_user.email}")
```

mvilanova (approved):

PR #5180, opened 2024-09-10
Do not round up incident cost

metroid-samus:


whitdog47:
@metroid-samus should we also do a db migration? Right now the `amount` column in the `incident_cost` and `case_cost` tables is set to 10 digits of accuracy with only 2 decimals places, i.e.,
```sa.Column("amount", sa.Numeric(precision=10, scale=2), nullable=True),```
_(from [this file](https://github.com/Netflix/dispatch/blob/master/src/dispatch/database/revisions/tenant/versions/2024-06-28_15a8d3228123.py#L48))_

mvilanova (approved):

PR #5185, opened 2024-09-10
Restrict select incident severities to active incidents only

whitdog47:
This PR allows select incident severities to only be allowed for Active incidents. For instance, "Undetermined" could be a severity that is only available while the incident is Active.

image


metroid-samus (approved):
Thanks for putting this together!

iamdancao:
No more "undetermined" incidents at Stable or Closed makes Dan a happy JackieDan

PR #5197, opened 2024-09-12
ux(slack): improve case report modal type to assignee and oncall resolution

ghost:
### Problem

Users may accidentally select the incorrect case type when reporting a case. Case types determine who the assignee of a case will be based on the configured on-call schedule for that case type. 

### Solution

This change improves the user experience of reporting a case in Slack by surfacing *who* the assignee of a case will be before they report it. Additionally, we provide additional context as 'tips' in the modal to inform the user that they choice they make when selecting a case type does determine backend behavior. Finally, we extend the PagerDuty on-call plugin to support fetching the on-call service URL to further enrich the modal with a hyperlink to on-call schedule that's associated with the case type they selected. 

### Notes

The backend will continue to function as expected. This is a no-op from a case creation perspective. We simply surface additional context to the user. The logic that determines assignee of a case can be seen below:

```python
    assignee_email = None
    if case_in.assignee:
        # we assign the case to the assignee provided
        assignee_email = case_in.assignee.individual.email
    elif case_type.oncall_service:
        # we assign the case to the oncall person for the given case type
        assignee_email = service_flows.resolve_oncall(
            service=case_type.oncall_service, db_session=db_session
        )
    elif current_user:
        # we fall back to assign the case to the current user
        assignee_email = current_user.email
```

### Screencast

https://github.com/user-attachments/assets/cd1af99f-a6ab-4538-a351-9de11fcbcd5e


wssheldon:
Note, I believe the override may not be working as expected -- digging in on that.

wssheldon:
> Note, I believe the override may not be working as expected -- digging in on that.

Fixed in https://github.com/Netflix/dispatch/pull/5197/commits/5c377c11300bc9f13e0cf0ab6c6648491df210b5

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -1748,6 +1927,7 @@ def handle_report_submission_event(
         case_type=case_type,
         dedicated_channel=True,
         reporter=ParticipantUpdate(individual=IndividualContactRead(email=user.email)),
+        assignee=ParticipantUpdate(individual=IndividualContactRead(email=assignee_email)),
```
whitdog47, line 1930: @wssheldon if `assignee_email` is `None`, will this throw an validation error?

wssheldon: It does:

```bash
    assignee=ParticipantUpdate(individual=IndividualContactRead(email=None)),
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for IndividualContactRead
```

I don't know that it *can be* `None` though. You must select an Assignee to proceed no matter what. In the event that it is, what do you think we should default to (assuming it gets that far)? Having a hard time coming up w/ that myself.

whitdog47: Line 1916 sets `assignee_email` to `None`. If we're sure that assignee will be selected, could we remove the `if` statement from line 1917?

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -1740,6 +1913,12 @@ def handle_report_submission_event(
     if form_data.get(DefaultBlockIds.case_type_select):
         case_type = {"name": form_data[DefaultBlockIds.case_type_select]["name"]}
 
+    assignee_email = None
+    if form_data.get(DefaultBlockIds.case_type_select):
+        assignee_email = client.users_info(
+            user=form_data[DefaultBlockIds.case_assignee_select]["value"]
+        )["user"]["profile"]["email"]
```
wssheldon, line 1920:
```suggestion
    assignee_email = client.users_info(
        user=form_data[DefaultBlockIds.case_assignee_select]["value"]
    )["user"]["profile"]["email"]
```

src/dispatch/plugins/dispatch_slack/case/interactive.py
wssheldon: Makes sense, made the change [here](https://github.com/Netflix/dispatch/pull/5197/files#r1759887243)

jschroth (approved):

PR #5204, opened 2024-09-13
Adding additional oncall metrics

whitdog47:
This PR adds additional metrics to the oncall feedback table, including number of incidents, cases, and total number of participants across all incidents and dedicated case channel cases during that person's shift.

image


src/dispatch/incident/service.py
```diff
@@ -125,6 +125,14 @@ def get_all_by_status(*, db_session, status: str, project_id: int) -> List[Optio
     )
 
 
+def get_all_last_x_hours(*, db_session, hours: int) -> List[Optional[Incident]]:
+    """Returns all incidents in the last x hours."""
```
forestmonster, line 129: `Returns all incidents in the last 'hours' hours.`

forestmonster: If you think that makes sense...

forestmonster (approved):
Looks great.

PR #5237, opened 2024-09-23
feat(genai): add llm case summary and historical summary to slack plugin

ghost:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -298,6 +305,116 @@ def create_signal_messages(case_id: int, channel_id: str, db_session: Session) -
...
+        db_session.query(SignalInstance, Signal)
+        .join(Signal)
+        .with_entities(SignalInstance.id, Signal)
+        .filter(SignalInstance.case_id == case.id)
+        .order_by(SignalInstance.created_at)
+    )
+
+    (first_instance_id, first_instance_signal) = signal_instances_query.first()
+
+    # Fetch up to 10 recent related cases
```
mvilanova, line 327: I'd suggest we move fetching the cases and related historical context outside of this function, so it can be reused for other case related use cases.

src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -298,6 +305,116 @@ def create_signal_messages(case_id: int, channel_id: str, db_session: Session) -
...
+    )
+
+    genai_plugin = plugin_service.get_active_instance(
+        db_session=db_session, project_id=case.project.id, plugin_type="artificial-intelligence"
+    )
+    if not genai_plugin:
+        return signal_metadata_blocks
+
+    response = genai_plugin.instance.chat_completion(
+        prompt=f"""
```
mvilanova, line 375: You should be able to fetch the prompt with the exception of current case, ... from the signal definition with the changes I made.

mvilanova (approved):

PR #5240, opened 2024-09-23
Enable users to create tasks from Slack

metroid-samus:
Users can now create tasks from Slack with the command `/dispatch-create-task`
Screenshot 2024-09-30 at 4 12 17 PM

The channel is notified when a task is created via Slack. The task can be resolved and reopened within Slack.
Screenshot 2024-09-30 at 3 56 23 PM


src/dispatch/database/revisions/tenant/versions/2024-09-23_32652e0360dd.py
```diff
@@ -0,0 +1,105 @@
+"""empty message
```
wssheldon, line 1: Could you add a quick description for this migration? It's not clear to me what it's about.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1514,6 +1517,108 @@ def handle_assign_role_submission_event(
...
+    body,
+    dict,
+    client: WebClient,
+    context: BoltContext,
+    db_session: Session,
+) -> None:
+    """Displays a modal for task creation."""
+    ack()
+    if context["subject"].type == CaseSubjects.case:
+        raise CommandError("Command is not currently available for cases.")
```
wssheldon, line 1531: suggestion: we should return this message as an 'error modal' as this will cause an 'unexpected error' state.

I believe there's a pattern for this already somewhere in the `interactive.py` files.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1514,6 +1517,108 @@ def handle_assign_role_submission_event(
...
+
+    for participant in active_participants:
+        participant_email = participant.individual.email
+        participant_info = contact_plugin.instance.get(participant_email, db_session=db_session)
+        participant_name = participant_info.get("fullname", participant.individual.email)
+        participant_list.append({"text": participant_name, "value": participant_email})
+
+    blocks = [
+        static_select_block(
+            label="Participant",
```
wssheldon, line 1558: suggestion: maybe `Assignee` is more descriptive?

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1514,6 +1517,108 @@ def handle_assign_role_submission_event(
...
+    )
+    incident = incident_service.get(db_session=db_session, incident_id=context["subject"].id)
+
+    task_in = TaskCreate(
+        assignees=[ParticipantUpdate.from_orm(owner)],
+        creator={"individual": {"email": user.email}},
+        description=form_data.get(CreateTaskBlockIds.description, ""),
+        incident=IncidentRead.from_orm(incident),
+    )
+    task_service.create(db_session=db_session, task_in=task_in)
```
wssheldon, line 1619: suggestion: i believe there's some opportunity to improve the UX at this stage of the flow and I'd recommend we create some connections here. 

after the task is created, where does it go? how do I find it in the UI? is there a direct link to it? how can i see my current tasks or all tasks for the incident.

i'm worried right now the tasks will get created and then become invisible basically.

whitdog47: I also had similar feedback that after submitting, it would be great to have some kind of confirmation. Maybe even let the assignee know in the channel that the task was created? (As an ephemeral message or not.)

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -1514,6 +1517,108 @@ def handle_assign_role_submission_event(
...
+
+    modal = Modal(
+        title="Create Task",
+        blocks=blocks,
+        submit="Create",
+        close="Close",
+        callback_id=CreateTaskActionIds.submit,
+        private_metadata=context["subject"].json(),
+    ).build()
+    client.views_open(trigger_id=body["trigger_id"], view=modal)
```
whitdog47, line 1580: The first time this works fine but trying to issue the command again I get 
```
slack_sdk.errors.SlackApiError: The request to the Slack API failed. (url: https://slack.com/api/views.open)
The server responded with: {'ok': False, 'error': 'expired_trigger_id'}
File "/Users/dwhittaker/dev/dispatch/src/dispatch/plugins/dispatch_slack/incident/interactive.py", line 1580, in handle_create_task_command
    client.views_open(trigger_id=body["trigger_id"], view=modal)
    ```

src/dispatch/plugins/dispatch_slack/incident/interactive.py
metroid-samus: Done.

src/dispatch/database/revisions/tenant/versions/2024-09-23_32652e0360dd.py
metroid-samus: Done.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
metroid-samus: Done.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
metroid-samus: I'm not able to replicate this error.

src/dispatch/plugins/dispatch_slack/incident/interactive.py
metroid-samus: Added a message to the channel when a task is created.  This message also includes a button that allows the users to resolve/reopen the task.
Screenshot 2024-09-30 at 3 56 23 PM

whitdog47 (approved):

PR #5286, opened 2024-10-03
If an error is already handled and logged, do not re-log that same error.

metroid-samus:
The `create_genai_signal_analysis_message()` function manages and logs edge cases where a signal summary isn't generated. When no summary is produced, the function returns an empty list. The result from `create_genai_signal_analysis_message()` is posted to a Slack channel. If the result is empty, Slack will generate an error. However, since this error has already been handled and logged, there is no need to log it again.


mvilanova (approved):
Thanks!

PR #5287, opened 2024-10-03
Show GenAI signal analysis errors in Slack message.

metroid-samus:


src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -287,6 +287,14 @@ def create_action_buttons_message(
     return Message(blocks=signal_metadata_blocks).build()["blocks"]
 
 
+def create_signal_metadata_blocks(signal_metadata_blocks: list[Block], message: str) -> list[Block]:
```
mvilanova, line 290:
```suggestion
def create_genai_signal_message_metadata_blocks(signal_metadata_blocks: list[Block], message: str) -> list[Block]:
```

mvilanova: I'd make the function name more specific to genai as we include `:magic_wand: *GenAI Alert Analysis` in the section text.

src/dispatch/plugins/dispatch_slack/case/messages.py
```diff
@@ -316,8 +324,8 @@ def create_genai_signal_analysis_message(
     ).first()
 
     if not first_instance_id or not first_instance_signal:
-        log.warning("Unable to generate GenAI signal analysis. No signal instances found.")
```
mvilanova, line 319: I'd keep the warnings so we can measure how often this happens.

metroid-samus: Done.

mvilanova (approved):

PR #5295, opened 2024-10-07
feat(slack): Expose create case slash command

bashbreakpoint:
In the slack conversation plugin, there are slash commands to be used on cases after they are created, but there is no slash command for creating a case. 
As I looked into this issue, I found that the full setup for allowing the command is present, but the final wire-up of the command was not. 
I added that final wire-up. 
I looked for tests for any functionality like this and found none. I can create it, but I have no model to go on.
I have looked at the Contributing docs on the dispatch web page, but I'm not very sure of the standards here and am fully open to changes and feedback.


bashbreakpoint:
If anyone is looking at this, I'm having trouble creating the alembic file I would need for this type migration. I am following the steps I find in #5240, merged last week. I've never worked with alembic and the contributing docs don't mention how to do this outside the scope of creating a new type. Extending a new type isn't transparent to me, and I'd love to do this the right way.

metroid-samus (approved):

PR #5299, opened 2024-10-08
Does not execute jpath if it is not sanitized.

metroid-samus:


wssheldon:
I recommend we validate `jpath` on the backend to prevent creation of unsantized ones. If our version of Pydantic supports Annotated types, we could create one for `jpath` with a `StringConstraint` and use that as the type for the model values.

src/dispatch/entity_type/service.py
```diff
@@ -105,6 +117,12 @@ def update(
 
     entity_type.signals = signals
 
+    try:
+        json.loads(entity_type_in.jpath)
```
wssheldon, line 121: I believe this will always raise an exception, as `JSONPath` is not valid `JSON`, e.g. `$.user.hostname` would be a valid `JSONPath`

metroid-samus: Fixed.

wssheldon: Thanks, that looks good. The only thing remaining I think would be valuable to add is a unit test that ensures this library is not vulnerable to the same issue in the javascript version and does indeed resolve the issue. Might want to move the parsing to a dedicated function to make testing easier.

Apologies for not calling all this out up front.

metroid-samus: Done.

wssheldon (approved):

PR #5317, opened 2024-10-09
fix(scheduler): weekly reports should run weekly

whitdog47:
This PR fixes the weekly incident report to run weekly instead of daily. Also ensured that an empty report wasn't sent in the case of all restricted incidents for that week, instead a notice is sent that indicates there were no incidents to report on.

### DIFF
--- a/src/dispatch/incident/scheduled.py
+++ b/src/dispatch/incident/scheduled.py
@@ -319,6 +319,10 @@ def incident_report_weekly(db_session: Session, project: Project):
         except Exception as e:
             log.exception(e)
 
+    # it's possible all incidents were restricted, so exit instead of sending an empty report
+    if not items_grouped:
+        return
+
     notification_kwargs = {
         "items_grouped": items_grouped,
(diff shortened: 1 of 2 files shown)


src/dispatch/incident/scheduled.py
```diff
@@ -319,6 +319,10 @@ def incident_report_weekly(db_session: Session, project: Project):
         except Exception as e:
             log.exception(e)
 
+    # it's possible all incidents were restricted, so exit instead of sending an empty report
+    if not items_grouped:
+        return
```
mvilanova, line 324: Should we still send a report, but say there are no incidents to report? If we don't send it, then one could wonder if something went wrong when we tried to generate it. At the very least, perhaps log an informational message that we didn't have incidents to report.

mvilanova (approved):

src/dispatch/incident/scheduled.py
whitdog47: I like the idea of sending the notification indicating there were no incidents to report.

PR #5325, opened 2024-10-10
add comment to docker yaml about default credentials

jschroth:


jschroth:
handled in https://github.com/Netflix/dispatch/pull/5324

PR #5332, opened 2024-10-12
Fixing case/incident report selects.

kevgliss:
This seems better; I was able to replicate the lock-up that we experienced locally. But am unable to now... 🤞🏻


src/dispatch/static/dispatch/src/case/type/CaseTypeSelect.vue
```diff
@@ -1,47 +1,42 @@
 <template>
   <v-select
-    v-model="case_type"
+    v-model="selectedIncidentType"
```
whitdog47, line 3: Is this right to use incident type in the `CaseTypeSelect.vue`?

whitdog47 (approved):

PR #5374, opened 2024-10-22
fix(feedback): store service id rather than schedule id

whitdog47:
This PR ensures that the service id is stored for each oncall service feedback item rather than the schedule id. Also, modifying the service filter so that it only shows the services that collect health metrics and providing pagination. Fixes DISDEV-368.


src/dispatch/feedback/service/scheduled.py
```diff
@@ -133,15 +135,15 @@ def count_active_participants(participants):
     send_oncall_shift_feedback_message(
         project=project,
         individual=individual,
-        schedule_id=schedule_id,
+        schedule_id=service_id,
```
mvilanova, line 138: Can we rename the parameter to service_id to avoid confusion?

PR #5375, opened 2024-10-22
bug(slack/api): Make slack /event respond to challenge

bashbreakpoint:
When configuring event listening for a slack app, your endpoint is required to respond correctly to the slack challenge that is sent. If your endpoint does not respond correctly, slack will not let you use it. Currently, the events endpoint does not respond correctly to the challenge because it is using a background thread to process all incoming events and responding with a static response. In order to use the events endpoint correctly (including using an @ in the incident channel to add an observer), that endpoint has to respond. I altered the event post to check the body for a "url_verification" type (which is how slack does the verification) and, if it exists, process the event synchronously in the same way the `actions` endpoint does. Otherwise, process async like it always has. I don't see any tests for slack endpoints. I can add one. I have deployed this to our own dev environment and verified it manually. I checked the following:

* Configuring the slack bot Event Subscription to point to our deployment at the `/api/v1/<organization>/events/slack/event` endpoint as described in the [documentation](/api/v1/<organization>/events/slack/event). This was previously failing
* adding someone to a channel using `@` results in them being added to the incident as an observer
* someone leaving the channel removes them from the incident
* using threads in a channel when "ban threads" is enabled results in dispatch whispering to the person not to use threads.

As far as I can see, this is the expected behavior of the event listening. I can add more if needed.


mvilanova (approved):
Thanks for the bugfix!

PR #5379, opened 2024-10-22
bug(docker): Revert python version bump

bashbreakpoint:
The move to python 3.14 breaks the docker build. The srsly package doesn't have a compatible wheel with anything above 3.12 right now ([docs](https://github.com/explosion/srsly/releases/tag/v2.4.8)), and it's using deprecated methods, so it can't be built by 3.14 yet. Because of this, `docker build .` is broken on master.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13.0-slim-bullseye as sdist
+FROM python:3.11.4-slim-bullseye as sdist
 
 LABEL maintainer="oss@netflix.com"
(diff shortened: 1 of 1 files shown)


mvilanova (approved):

bashbreakpoint:
I'm sorry, I have no idea how to add a label. I suspect I lack permissions to do so (which is fine).

mvilanova:
> I'm sorry, I have no idea how to add a label. I suspect I lack permissions to do so (which is fine).

No worries. I'll check if we can give perms to collaborators.

PR #5385, opened 2024-10-23
fix(ui): fixes details tab selects

whitdog47:
This PR locks the project field in the Details tab for both incidents and cases and fixes the type and priority dropdowns.


kevgliss:
So, this is fine... but you might run into issues in the case/incident report forms where we do allow for the project to change.

kevgliss (approved):

PR #5394, opened 2024-10-25
chore(node): Upgrade to node lts

bashbreakpoint:
Upgrading the docker build and nvm to node 20.18, which is the latest supported, stable LTS version. Node 12 and 16 have been EoL since April, 2022 and August, 2023, respectively. In addition, a transitive dependency of mjml, minimatch, is not compatible with anything less than node 14 at this point, so email generation from templates was broken. With the move to newer node, I had to deal with a new node/docker restriction which won't allow you to `npm install` in the root directory. I wanted to maintain backwards compatiblity with current installs and their MJML_PATH parameter, so I worked around the problem by installing mjml in another directory and then moving node_modules to the root directory. It's not pretty, but it worked. I'm trying to run tests locally and I get one failure, but I'm not sure if it's my setup or something I need to correct. I've manually tested the setup, gone into every page, and not experienced problems.


mvilanova (approved):
Thanks again for giving this file some love!

PR #5431, opened 2024-11-05
fix(db): cleaning up alembic migration files

whitdog47:
This PR syncs the database migration with actual values in the database. Further commands to autogenerate migration files now show as:
```INFO:No changes found skipping revision creation.```
Fixes DISDEV-377

### CODE REVIEW

File: src/dispatch/signal/models.py [hunk 1]
```diff
@@ -48,19 +48,6 @@ class RuleMode(DispatchEnum):
     inactive = "Inactive"
 
 
-assoc_signal_instance_tags = Table(
```
mvilanova, line 51: This is not used and getting populated in prod, correct?

whitdog47: No. It appears that it was dropped in [this migration](https://github.com/Netflix/dispatch/blob/master/src/dispatch/database/revisions/tenant/versions/2023-02-13_b168b50764c7.py#L71) but the code was never removed.

image

mvilanova (approved):

PR #5434, opened 2024-11-05
Database/update database

whitdog47:
This PR updates the sample database to the latest migration and adds case types and priorities in preparation for playwright tests against cases.


metroid-samus (approved):
Thanks! This is super helpful!

PR #5447, opened 2024-11-08
fix(docker-image): downgrade Python base image and remove pkg_resources

rialg:
* Downgrade to docker base image to Python 3.11.10, or else replace distutils module ([removed in version 3.12](https://docs.python.org/3/whatsnew/3.12.html))
* Replace setuptools' `pkg_resources` modules, which was deprecated, [with importlib](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources).
* Upgrade `lxml` module


rialg:
Hi @mvilanova , could you take a look at this changes? :pray: 
Also, do you know when the next release will be available? Is there anything missing that I could help with?

mvilanova (approved):

mvilanova:
Can you fix this?
```
src/dispatch/common/utils/cli.py:3:46: F401 [*] `importlib.metadata.EntryPoint` imported but unused
  |
1 | import traceback
2 | import logging
3 | from importlib.metadata import entry_points, EntryPoint
  |                                              ^^^^^^^^^^ F401
4 | from sqlalchemy.exc import SQLAlchemyError
  |
  = help: Remove unused import: `importlib.metadata.EntryPoint`
```

PR #5449, opened 2024-11-08
chore(deps): bumping dependencies

whitdog47:


requirements-dev.txt
```diff
@@ -4,6 +4,8 @@
 #
 #    pip-compile requirements-dev.in
 #
+--index-url https://pypi.netflix.net/simple
+
```
wssheldon, line 8:
```suggestion
```

setup.cfg
```diff
@@ -7,7 +7,7 @@ selenium_driver = chrome
 self-contained-html = true
 
 [bdist_wheel]
-python-tag = py38
+python_tag = py38
```
kevgliss, line 10: is this even needed? I don't think we support 3.8 (or building wheels for that matter).

whitdog47: I don't think so either. Removing.

mvilanova (approved):

PR #5458, opened 2024-11-13
fix(ui/ux): show zero escalated cases if there are none

mvilanova:
Screenshot 2024-11-12 at 4 43 03 PM


src/dispatch/static/dispatch/src/dashboard/case/CaseOverview.vue
```diff
@@ -216,13 +199,15 @@ export default {
         if (item.triage_at) {
           return 1
         }
+        return 0
```
mvilanova, line 202: This is the change.

src/dispatch/static/dispatch/src/dashboard/case/CaseOverview.vue
```diff
@@ -216,13 +199,15 @@ export default {
...
       })
     },
     totalCasesEscalated() {
       return sumBy(this.items, function (item) {
         if (item.escalated_at && item.incidents.length > 0) {
           return 1
         }
+        return 0
```
mvilanova, line 210: This is the change.

wssheldon (approved):

PR #5459, opened 2024-11-13
feat(ui): add dedicated status tab for can report in incident edit sheet

ghost:
Here nor there on the name or position of the tab. We could call it `CAN`, but I doubt that's widely understood.

Screenshot 2024-11-12 at 4 42 51 PM

```bash
 ✓ src/tests/TimelineReportTab.spec.js (7)
   ✓ mounts correctly and displays tactical reports
   ✓ filters out non-tactical report events
   ✓ displays event details correctly
   ✓ sorts events chronologically
   ✓ displays UTC time notice
   ✓ handles empty events array
   ✓ displays correct icon for tactical reports

 Test Files  1 passed (1)
      Tests  7 passed (7)
   Start at  16:57:50
   Duration  471ms
```

### DIFF
--- a/src/dispatch/static/dispatch/src/incident/EditSheet.vue
+++ b/src/dispatch/static/dispatch/src/incident/EditSheet.vue
@@ -34,7 +34,7 @@
       </template>
       <v-tabs color="primary" fixed-tabs v-model="tab" :disabled="id == null">
         <v-tab value="details"> Details </v-tab>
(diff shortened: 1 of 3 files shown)


mvilanova:
I'd call it Reports, as that's what they're.

mvilanova (approved):

whitdog47 (approved):

PR #5462, opened 2024-11-13
fix(case/ui): fixes case page to show escalation

whitdog47:
This pull request addresses an issue on the case page where the escalation status was not being displayed correctly after the Escalate Case dialog completed. It also ensures that the user cannot try to escalate a case that has already been escalated.

#### Key Changes
1. **Conditional Dialog Display**:
   - Updated the `CaseStatusSelectGroup.vue` component to ensure the escalation dialog is only opened if the case is not already in an "Escalated" status.

2. **Nullish Coalescing Operator**:
   - Modified the `EscalateDialog.vue` component to use the nullish coalescing operator (`?.`) for safer access to `incident_type`, preventing potential errors when `caseType` is undefined.

3. **Page Reload for Status Update**:
   - Added a page reload in the `store.js` actions to ensure the case page reflects the updated status after an escalation, providing immediate feedback to the user.

#### Files Changed
- **`src/dispatch/static/dispatch/src/case/CaseStatusSelectGroup.vue`**: 
  - Adjusted the condition for opening the escalation dialog.

- **`src/dispatch/static/dispatch/src/case/EscalateDialog.vue`**: 
  - Implemented the nullish coalescing operator for safer property access.

- **`src/dispatch/static/dispatch/src/case/store.js`**: 
  - Added a window reload to update the case status on the page.

Fixes DISDEV-317


wssheldon:
Couple things I think are worth updating here before shipping:

- We should provide some feedback to the user when they try to escalate an already escalated case. Right now (AFACIT) the button will just appear to not work. My preference would be to put the button in some kind of disabled state with a tooltip that explains why you can't click the button.

- Reloading the page feels a bit heavy, is it necessary? I think we should be able to do an optimistic update for the case status.

A meta note, I think a screencast would come in really handy here, to show the before / after for these changes. I've been using: https://screen.studio/

wssheldon:
I would also be curious if we could have caught this issue with a more solid set of component level tests. We have one for this component actually in `tests/` directory: `CaseStatusSelectGroup.spec`

wssheldon (approved):
I don't think anything I mentioned is blocking. But, I do think we should do them, the optimistic update and button feedback, specifically.

mvilanova (approved):

PR #5507, opened 2024-11-21
fix(ui): fixes edit pane on case type

whitdog47:
This pull request addresses an issue with the edit pane in the case type UI. The changes ensure that the project and incident project are correctly fetched and assigned, and updates the way project IDs are handled across different components. 

#### Changes
The main changes include:
1. **`NewEditSheet.vue`:**
   - Enhanced the logic for setting `project` and `incidentProject` within the `created` lifecycle hook. The code now correctly fetches and assigns the project data fetched from the API.

2. **`IncidentTypeSelect.vue`:**
   - Simplified the filter options by removing redundant `project_id` assignment, ensuring that the correct project is always applied.

3. **`ServiceSelect.vue`:**
   - Corrected the assignment of `project_id` in filter options, changing it from a map operation to direct access, which is more efficient and accurate.
   - Added an additional parameter to `createParametersFromTableOptions` to specify context.

#### Previous behavior
https://github.com/user-attachments/assets/8d890e97-5413-4084-85d6-499a3b6eed52

### DIFF
--- a/src/dispatch/static/dispatch/src/case/type/NewEditSheet.vue
+++ b/src/dispatch/static/dispatch/src/case/type/NewEditSheet.vue
@@ -165,7 +165,6 @@ import ProjectSelect from "@/project/ProjectSelect.vue"
 import TemplateSelect from "@/document/template/TemplateSelect.vue"
 import ProjectApi from "@/project/api"
 
-
 export default {
   setup() {
     return {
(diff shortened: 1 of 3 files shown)


src/dispatch/static/dispatch/src/case/type/NewEditSheet.vue
```diff
@@ -221,7 +222,10 @@ export default {
   created() {
     if (this.$route.query.project) {
       this.project = { name: this.$route.query.project }
```
aniekanoffiong, line 224: Probably makes sense to take out this initial assignment then, since the subsequent assignment is not optional.

aniekanoffiong, line 224:
```suggestion
```

metroid-samus (approved):

PR #5522, opened 2024-11-22
Do not throw errors when trying to list members from a non-existent Google Group

metroid-samus:


src/dispatch/plugins/dispatch_google/groups/plugin.py
```diff
@@ -173,8 +173,13 @@ def remove(self, email: str, participants: List[str]):
     def list(self, email: str):
```
mvilanova, line 173: nit
```suggestion
    def list(self, email: str) -> list[str]:
```

mvilanova (approved):

PR #5564, opened 2024-12-03
Add tags and tickets fields to get cases request

metroid-samus:


kevgliss:
hmm, I don't think we want to do it this way. When include is supported:

https://github.com/Netflix/dispatch/blob/7bc0a674373da62fbd7514095ceb277b796c5632/src/dispatch/case/views.py#L125

We create a special result set using:

https://github.com/Netflix/dispatch/blob/7bc0a674373da62fbd7514095ceb277b796c5632/src/dispatch/common/utils/views.py#L4

We should figure out a way to support nested objects via that function such that additionally nested fields can be gathered as well.

This would also solve the problem across the incident, case and task apis.

Lmk if you want to pair on that approach, it's been a long time since I looked at that code and it's sort of magical.

whitdog47 (approved):

kevgliss (approved):

PR #5583, opened 2024-12-06
fix(ui): filter incident type by selected project

whitdog47:
Fixes a bug that doesn't properly filter the incident type for the selected project.


jneelamekam (approved):
Thank you for fixing this, lgtm.

PR #5616, opened 2024-12-13
feat(ui): allow project to have a display name separate from its key

whitdog47:
This PR allows admins to select a display name for projects and renames the current `name` field in the UI as the "key". In the database migration, the `display_name` for all projects is set to the existing `name` field value. Fixes DISDEV-242.

image

We also fix an error in the Slack command `\dispatch-update-incident` modal that threw an error when the user tried to change the project by making the project field read-only.

image


src/dispatch/project/models.py
```diff
@@ -40,6 +40,8 @@ class Project(Base):
         cascade="all, delete-orphan",
     )
 
+    display_name = Column(String, nullable=True)
```
wssheldon, line 43: will this stay nullable or will it be non-nullable post-migration?

wssheldon (approved):

src/dispatch/project/models.py
whitdog47: Good point. I don't think it should be nullable. I fixed this and the database migration.

PR #5618, opened 2024-12-13
fix(case): close escalated case when associated incident is closed

whitdog47:
This PR ensures that when an incident is closed, all associated escalated cases are automatically closed and their case resolution includes a message indicating closure as part of the incident. Fixes DISDEV-243.


src/dispatch/incident/flows.py
```diff
@@ -544,6 +546,15 @@ def incident_closed_status_flow(incident: Incident, db_session=None):
...
                     )
 
+    # if there are any associated cases, we close them as well
+    for case in incident.cases:
+        case.resolution = (
+            f"Closed as part of incident {incident.name}. See incident for more details."
+        )
+        case.resolution_reason = CaseResolutionReason.escalated
+        case.status = CaseStatus.closed
+        case_flows.case_closed_status_flow(case=case, db_session=db_session)
```
wssheldon, line 556: Handles any error to let the rest of the process complete.

```suggestion
for case in incident.cases:
    try:
        case.resolution = (
            f"Closed as part of incident {incident.name}. See incident for more details."
        )
        case.resolution_reason = CaseResolutionReason.escalated
        case.status = CaseStatus.closed
        case_flows.case_closed_status_flow(case=case, db_session=db_session)
    except Exception as e:
        log.exception(
            f"Failed to close case {case.name} while closing incident {incident.name}. Error: {str(e)}"
        )
```

wssheldon (approved):
Good looks

This feels more like an enhancement to me and less a bug -- along the same lines I wouldn't consider this tech debt (us being on SQLAlchemy 1.2 is tech debt, as an example, from my perspective).

src/dispatch/incident/flows.py
```diff
@@ -544,6 +546,19 @@ def incident_closed_status_flow(incident: Incident, db_session=None):
                         document=document, db_session=db_session
                     )
 
+for case in incident.cases:
```
wssheldon, line 549: spacing got messed up post suggestion

whitdog47: Fixed. thanks for catching

PR #5646, opened 2024-12-21
fix(utils): adding new labels

whitdog47:
Adding new labels to github_utils.


mvilanova (approved):

mvilanova:
We should add a section for each new label [here](https://github.com/Netflix/dispatch/pull/5646/files#diff-576c7ab013c9e1c76334bbcf8471a646704f2ba14083c86c8e5bff9292a01034R130) as well.

PR #5732, opened 2025-01-27
Adds configurable delayed message notification

aaronherman:
Adds the option to enable/disable sending the below Slack message on an incident

image


src/dispatch/static/dispatch/src/incident/priority/NewEditSheet.vue
```diff
@@ -108,6 +108,13 @@
                   hint="Check if this incident priority should be the default."
                 />
               </v-col>
+              <v-col cols="12">
+                <v-checkbox
+                  v-model="disable_delayed_message_warning"
+                  label="Disable delayed message warning"
+                  hint="Would you like to disable Dispatch from notifying users when they send a message in Slack when it's outside of the Incident Commander's 9a-5p working schedule?"
```
metroid-samus, line 115: Nit
```suggestion
                  hint="Would you like to disable Dispatch from notifying users when they send a message in Slack outside of the Incident Commander's 9a-5p working schedule?"
```

metroid-samus (approved):

PR #5740, opened 2025-01-28
fix(slack): individual may not be defined

whitdog47:
Fixes a bug when `individual` is not defined.


metroid-samus (approved):

src/dispatch/plugins/dispatch_slack/incident/interactive.py
```diff
@@ -806,7 +806,7 @@ def handle_timeline_added_event(
 
         source = "Slack message"
         # if the individual is not found, see if it is a bot
-        if individual is None:
+        if not individual:
```
wssheldon, line 809: It may still be undefined here, we need to set `individual` to `None` outside the `except`

wssheldon (approved):

PR #5742, opened 2025-01-29
fix: case type query params get ignored on Report Case form

aaronherman:
This pattern matches the way Incident ReportSubmissionCard is set up since there isn't a "Load More" and fixes the issue where this.project isn't set in time for the evaluation


whitdog47 (approved):
Sometimes simpler is better. Thanks!

metroid-samus (approved):

PR #5759, opened 2025-02-03
feat(signal): adding historical events to signal definitions

whitdog47:
This pull request introduces the capability to log and view create and update events associated with signal definitions. This enhancement allows users to track changes and updates to signals over time, providing greater transparency and traceability.

#### Key Changes
- **Database Schema Update**: 
  - Added a new column `signal_id` to the `event` table to establish a relationship between events and signals.
  - Created a foreign key constraint to ensure referential integrity.

- **Model Updates**:
  - Updated the `Event` and `Signal` models to include relationships for tracking events.

- **Service Enhancements**:
  - Introduced `log_signal_event` function to log events related to signal creation and updates.
  - Modified `create` and `update` functions in the signal service to log events whenever a signal is created or updated.

- **API Changes**:
  - Updated the `create_signal` and `update_signal` endpoints to include the current user, enabling event logging with user context.

- **Frontend Additions**:
  - Added a new `HistoryViewer.vue` component to display the timeline of events for a signal.
  - Integrated the history viewer into the signal table, allowing users to view the history of a signal directly from the table interface.

#### Sample screenshot
Screenshot 2025-02-02 at 9 57 33 PM


src/dispatch/event/service.py
```diff
@@ -184,6 +185,59 @@ def log_case_event(
...
+def log_signal_event(
+    db_session,
+    source: str,
+    description: str,
+    signal_id: int,
+    individual_id: int = None,
+    started_at: datetime = None,
+    ended_at: datetime = None,
+    details: dict = None,
+    dispatch_user_id: int = None,
```
wssheldon, line 197:
```suggestion
    individual_id: int | None = None,
    started_at: datetime | None = None,
    ended_at: datetime | None = None,
    details: dict | None = None,
    dispatch_user_id: int | None = None,
```

wssheldon (approved):

src/dispatch/event/service.py
mvilanova: We should do the same in the log incident and case functions.

PR #5768, opened 2025-02-06
fix(signal): only record change event if changes detected

whitdog47:
This PR modifies the signal update logic to ensure that change events are only recorded when actual changes are detected. This prevents unnecessary logging and improves the efficiency of the event logging system.


wssheldon (approved):

mvilanova:
@whitdog47 do we still need these changes if we're going to avoid updating the signal definition on the origin when this has not changed?

PR #5786, opened 2025-02-24
Propose additional security check to auth/views.py

ViktorMares:
Hi,

There is an IDOR vulnerability that allows any user to edit another user (via the Experimental Features button), by simply changing the user ID in the path and in the HTTP request. This happens because the application is missing a security check (when "role" is not provided in the PUT HTTP Request's body)

This additional security check will cause the application to return a 403 Forbidden if a user that is not an "Owner" is trying to edit another user, essentially remediating an IDOR vulnerability.

### DIFF
--- a/src/dispatch/auth/views.py
+++ b/src/dispatch/auth/views.py
@@ -146,6 +146,14 @@ def update_user(
     user_in: UserUpdate,
     current_user: CurrentUser,
 ):
+    """Check if Current_user is Owner and is trying to edit another user"""
+    current_user_organization_role = current_user.get_organization_role(organization)
+    if current_user_organization_role != "Owner" and current_user.id != user_id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=[{"msg": "A user that is not an Owner is trying to update another user."}],
+        )
+    
     """Update a user."""
     user = get(db_session=db_session, user_id=user_id)
     if not user:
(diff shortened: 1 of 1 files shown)


src/dispatch/auth/views.py
```diff
@@ -146,6 +146,14 @@ def update_user(
     user_in: UserUpdate,
     current_user: CurrentUser,
 ):
+    """Check if Current_user is Owner and is trying to edit another user"""
+    current_user_organization_role = current_user.get_organization_role(organization)
+    if current_user_organization_role != "Owner" and current_user.id != user_id:
```
whitdog47, line 151:
```suggestion
    if current_user_organization_role != UserRoles.owner and current_user.id != user_id:
```

whitdog47 (changes requested):

src/dispatch/auth/views.py
```diff
@@ -146,6 +146,14 @@ def update_user(
...
     current_user: CurrentUser,
 ):
+    """Check if Current_user is Owner and is trying to edit another user"""
+    current_user_organization_role = current_user.get_organization_role(organization)
+    if current_user_organization_role != UserRoles.owner and current_user.id != user_id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=[{"msg": "A user that is not an Owner is trying to update another user."}],
+        )
+
```
whitdog47, line 156:
```suggestion

```

whitdog47: remove the whitespace from the blank line

whitdog47 (changes requested):

whitdog47 (approved):

ViktorMares:
@whitdog47 I think the merge request has been unsuccessful, due to the lack of PR labels, when we tried to push it. Can you please try to merge again, as the merge protection doesn't allow me to do it.

whitdog47:
Thanks for the contribution!

PR #5801, opened 2025-03-05
fix(ui): make content scrollable when bottom sheet is visible

whitdog47:
Previously, after selecting an incident or case, the content was no longer scrollable due to the bottom sheet appearing. This fixes that scroll lock. Fixes DNRPLAT-94


aaronherman (approved):
📜

PR #5807, opened 2025-03-06
fix(db session): fixes db session typing

mvilanova:


whitdog47 (approved):
Thanks for the cleanup!

PR #5808, opened 2025-03-07
feat(slack plugin): moves investigate button to actions section

mvilanova:
Screenshot 2025-03-06 at 5 05 27 PM

### DIFF
--- a/src/dispatch/cli.py
+++ b/src/dispatch/cli.py
@@ -567,12 +567,12 @@ def stamp_database(revision, revision_type, tag, sql):
 )
 @click.option("--revision-type", type=click.Choice(["core", "tenant"]))
 @click.option(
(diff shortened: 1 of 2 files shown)


whitdog47 (approved):

whitdog47:
@mvilanova - what do you think about having the button conditionally appear based on whether that project has an investigation plugin enabled or not?

mvilanova:
> @mvilanova - what do you think about having the button conditionally appear based on whether that project has an investigation plugin enabled or not?

Yeah, that's a good idea.

mvilanova:
> @mvilanova - what do you think about having the button conditionally appear based on whether that project has an investigation plugin enabled or not?

Yeah, that's a good idea.

PR #5814, opened 2025-03-10
Add Classic Cost Model for Cases and Calculate case costs using both cost models in parallel.

metroid-samus:
This PR implements the classic cost model functionality for cases, mirroring the existing incident cost model while adapting for case-specific concepts.

# Changes

## Core Implementation
Added role-based engagement multipliers for case roles:
 * Owner: 1.0 (equivalent to incident commander)
 * Reporter: 0.5
 * Participant: 0.5
 * Observer: 0 (no cost calculation)

Implemented time calculation functions that handle case status transitions:
 * New/Triage: Active time tracking
 * Closed: Cost calculation stops at closed_at
 * Escalated: Cost calculation stops at escalated_at

## Cost Calculation
 * Both cost models (new and classic) now run in parallel
 * Results are returned as a dictionary with both cost calculations
 * Time calculations account for 8-hour workdays when cases extend beyond 24 hours

## Testing
The implementation includes tests for:
 * Role engagement multiplier accuracy
 * Time calculation edge cases
 * Cost calculation verification
 * Classic Cost Model integration
 * Status transition handling

## Migration
No migration required. The classic model will begin calculating costs for cases automatically while maintaining the existing new model calculations.

### DIFF
--- a/src/dispatch/case/models.py
+++ b/src/dispatch/case/models.py
@@ -19,6 +19,7 @@ from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
 from sqlalchemy_utils import TSVectorType, observes
 
+from dispatch.case.enums import CostModelType
 from dispatch.case.priority.models import CasePriorityBase, CasePriorityCreate, CasePriorityRead
 from dispatch.case.severity.models import CaseSeverityBase, CaseSeverityCreate, CaseSeverityRead
 from dispatch.case.type.models import CaseTypeBase, CaseTypeCreate, CaseTypeRead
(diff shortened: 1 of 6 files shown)


wssheldon (approved):

src/dispatch/case/models.py
```diff
@@ -199,10 +200,22 @@ def participant_emails(self) -> list:
...
+
+    @hybrid_property
+    def total_cost_new(self):
+        total_cost = 0
+        if self.case_costs:
+            for cost in self.case_costs:
+                if cost.case_cost_type.model_type == CostModelType.classic:
+                    continue
                 total_cost += cost.amount
         return total_cost
```
wssheldon, line 220: nit: could fail fast here to reduce nesting and improve readability

PR #5832, opened 2025-03-14
feat(incident): allow exclusion of incident types from review documents

aaronherman:
This PR allows an Incident Type to be configurable to have its tasks excluded from review documents.

Adds a new column `exclude_from_review` on incident_type
Updates` `incident_stable_status_flow()` to return early if exclude_from_review is True
Adds a checkbox to incident/type/NewEditSheet.vue


src/dispatch/incident/flows.py
```diff
@@ -513,6 +514,13 @@ def incident_stable_status_flow(incident: Incident, db_session=None):
         log.info("The post-incident review document has already been created. Skipping creation...")
         return
 
+    incident_type = get_incident_type(
+        db_session=db_session, incident_type_id=incident.incident_type_id
+    )
+    if incident_type.exclude_from_review:
+        log.info("Incident type is excluded from review. Skipping creation...")
```
metroid-samus, line 521: Nit: 
```suggestion
        log.info(f"Incident type {incident_type.name} is excluded from review. Skipping creation...")
```

metroid-samus (approved):

PR #5834, opened 2025-03-17
Remove newline character(s) from Jira ticket title.

metroid-samus:


src/dispatch/plugins/dispatch_jira/plugin.py
```diff
@@ -465,7 +465,7 @@ def create_case_ticket(
                 "issuetype": issuetype,
                 "assignee": assignee,
                 "reporter": reporter,
-                "summary": title,
+                "summary": title.strip(),
```
aaronherman, line 468: Since a new line could be in the middle of the summary, I think we'll need to use `.replace()`
```suggestion
                "summary": title.replace("\n", ""),
```

aaronherman (approved):
I added `.replace("\n", "")` on two other summary fields, just in case!

PR #5837, opened 2025-03-17
Enhancement/new endpoint for signal data

whitdog47:
This pull request introduces a new feature to the signal module, allowing users to retrieve signal data based on a given entity name and type. The changes include the addition of a new endpoint, the necessary backend logic to support this functionality, and comprehensive tests to ensure reliability.

**Key Changes:**

1. **New Model:**
   - Added `SignalData` model to encapsulate the response structure for the new endpoint, including fields for the number of signals alerted, snoozed, and the status of snoozes (active or expired).

2. **Service Layer:**
   - Implemented `get_signal_data` function in `service.py` to query the database for signal data based on the provided entity name and type.
   - Utilized JSONB operations to filter signals associated with the specified entity.
   - Added logic to calculate active and expired snoozes, as well as signals with and without snooze actions.
   - Introduced a parameter `num_days` to allow filtering signals based on a date threshold.

3. **API Endpoint:**
   - Added a new GET endpoint `/data` in `views.py` to expose the signal data retrieval functionality.
   - The endpoint accepts query parameters for `entity_value`, `entity_type_id`, and `num_days`.

4. **Testing:**
   - Added comprehensive tests in `test_signal_data_service.py` to verify the functionality of `get_signal_data`.
   - Tests cover basic functionality, date-based filtering, snooze filters, expired snooze filters, and scenarios where no signals are found.
   - Ensured that the tests validate the correct calculation of signal counts and snooze statuses.

5. **Enhancements:**
   - Improved query performance by leveraging SQLAlchemy's JSONB capabilities and efficient filtering.
   - Enhanced the API's flexibility by allowing optional date-based filtering.


src/dispatch/signal/models.py
```diff
@@ -379,6 +379,13 @@ class AdditionalMetadata(DispatchBase):
     important: Optional[bool]
 
 
+class SignalData(DispatchBase):
+    num_signals_alerted: Optional[int]
```
aliciamatsumoto, line 383: explicitly referencing that these are counts of signal instances (rather than signal definitions) would be more pedantic but perhaps less intuitive - throwing out the idea but no strong opinion either way

src/dispatch/signal/models.py
```diff
@@ -379,6 +379,13 @@ class AdditionalMetadata(DispatchBase):
     important: Optional[bool]
 
 
+class SignalData(DispatchBase):
```
aliciamatsumoto, line 382: brainstorming other name ideas ... would "statistics" or "counts" perhaps more clearly distinguish the endpoint from one providing qualitative details or metadata?

aliciamatsumoto (approved):

src/dispatch/signal/models.py
whitdog47: Good idea. Updated.

PR #5858, opened 2025-03-27
fix(signal): guard against object not found when recording changes

whitdog47:
Fixes an issue where the signal contains a service, priority, or type that doesn't exist in the system.


jschroth (approved):
👍

PR #5898, opened 2025-04-15
feat(int): adding internal scheduler placeholder

whitdog47:


src/dispatch/cli.py
```diff
@@ -641,6 +641,7 @@ def dispatch_scheduler():
     """Container for all dispatch scheduler commands."""
     # we need scheduled tasks to be imported
     from .case.scheduled import case_close_reminder, case_triage_reminder  # noqa
+    from .case.scheduled_internal import schedule_placeholder  # noqa
```
kevgliss, line 644: What about just doing `*` ? That seems a tad more flexible so you can have multiple and they could be named whatever they want.

kevgliss (approved):

src/dispatch/cli.py
whitdog47: @kevgliss , wildcard imports aren't allowed inside a function. When the file is imported, it will read the entire file anyway and schedule all of the annotated functions.

PR #5908, opened 2025-04-18
feat(entity): when a new entity_type is created, find new entities

whitdog47:
This PR updates the entity_type creation to back propagate the creation of entities based on these types to previous signal instances. This allows a user to create a new entity_type and use it immediately in a snooze.


aliciamatsumoto (approved):

src/dispatch/entity_type/service.py
```diff
@@ -75,6 +75,25 @@ def create(*, db_session: Session, entity_type_in: EntityTypeCreate) -> EntityTy
...
+            db_session.query(SignalInstance).filter(SignalInstance.signal_id == signal.id).all()
+        )
+        for signal_instance in signal_instances:
+            # Extract and create entities for this instance using only the new entity_type
+            new_entities = find_entities(db_session, signal_instance, [entity_type])
+            # Associate new entities with the signal_instance
+            for entity in new_entities:
+                if entity not in signal_instance.entities:
+                    signal_instance.entities.append(entity)
+    db_session.commit()
```
wssheldon, line 95: what will happen if there's tens of thousands of signal instances?

wssheldon: I point it because I've attempted to do the same and it was quite slow. There's ways to optimize this but it hasn't been added.

whitdog47: What if we just do this for signal_instances associated with cases in either `New` or `Triage` state?

whitdog47: Created a new post endpoint that allows the passing of a case_id, so we only create entities for that case.

PR #5920, opened 2025-04-21
feat(user engagement): only add participants to case threads that originate from signals in production

mvilanova:


whitdog47 (approved):
Great idea - thank you!

mvilanova:
Found another way of not notifying.

PR #5921, opened 2025-04-21
Nit: equal spacing on both button sides.

forestmonster:
My local instance is borked, but will this fix the button spacing @aaronherman ?

form


wssheldon (approved):

aaronherman:
Events, Cases and Incidents should all be looking better now. Thanks @forestmonster!

PR #5926, opened 2025-04-22
ui/ux(case): fix hotkey layout, add tag selector, disable graph tab

ghost:
Adds this back - also unlikely to cause any table load issues

#### Screencast

https://github.com/user-attachments/assets/8dce8a26-134f-4d5a-a72c-fa276b37d5ce


src/dispatch/static/dispatch/src/case/CaseAttributesDrawer.vue
```diff
@@ -150,6 +151,15 @@ const handleResolutionUpdate = (newResolution) => {
             <ProjectSearchPopover :project="modelValue.project?.name" class="pl-6" />
           </v-col>
         </v-row>
+
+        <v-row no-gutters class="pt-6">
+          <v-col cols="12">
+            <div class="dispatch-font mb-2">Labels</div>
```
wssheldon, line 157:
```suggestion
            <div class="dispatch-font mb-2">Tags</div>
```

whitdog47 (approved):

PR #5931, opened 2025-04-24
feat(slack): adds add entity button in snooze modal

whitdog47:
Adds a button to the Snooze Signal modal that goes directly to the signal viewer in the Dispatch UI, allowing the user to add a new entity type quickly.

image

### DIFF
--- a/src/dispatch/plugins/dispatch_slack/case/interactive.py
+++ b/src/dispatch/plugins/dispatch_slack/case/interactive.py
@@ -30,6 +30,7 @@ from dispatch.case import service as case_service
 from dispatch.case.enums import CaseResolutionReason, CaseStatus
 from dispatch.case.models import Case, CaseCreate, CaseRead, CaseUpdate
 from dispatch.case.type import service as case_type_service
+from dispatch.config import DISPATCH_UI_URL
 from dispatch.conversation import flows as conversation_flows
 from dispatch.entity import service as entity_service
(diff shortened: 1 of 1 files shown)


wssheldon (approved):

src/dispatch/plugins/dispatch_slack/case/interactive.py
```diff
@@ -733,6 +735,10 @@ def snooze_button_click(
         case_id = case.id
         subject.type = SignalSubjects.signal_instance
         subject.id = case.signal_instances[0].signal.id
+        case_url = (
+            f"{DISPATCH_UI_URL}/{case.project.organization.slug}/cases/{case.name}/"
+            f"signal/{case.signal_instances[0].id}"
```
wssheldon, line 740: I feel like this comes up often -- maybe it could be a property on the `case` model?

wssheldon: just a 💡

aliciamatsumoto (approved):

src/dispatch/plugins/dispatch_slack/case/interactive.py
wssheldon: I think you'd get the most recent version of what's in the database since it's re-fetched above:

https://github.com/Netflix/dispatch/pull/5931/files#diff-a257e599f8cc347ecae069c092685a9c93f5b03435abfca301214ceea755d06cR734

The property would effectively be shorthand for the same thing here

PR #5935, opened 2025-04-28
fix: allow GET on incidents/cases to include participants & other fields

aaronherman:
A user reported that `include[]=participants` hasn't been working for a few months. We currently are using the [`CasePagination`](https://github.com/Netflix/dispatch/blob/main/src/dispatch/case/models.py#L386-L391) for [`get_cases`](https://github.com/Netflix/dispatch/blob/main/src/dispatch/case/views.py#L135) but I think we should use `CaseExpandedPagination` to support all `include[]` options.

### DIFF
--- a/src/dispatch/case/models.py
+++ b/src/dispatch/case/models.py
@@ -305,7 +305,6 @@ class CaseReadMinimal(CaseBase):
     project: ProjectRead
     assignee: ParticipantReadMinimal | None
     case_costs: list[CaseCostReadMinimal] = []
-    participants: Optional[List[ParticipantRead]] = []
 
 
 CaseReadMinimal.update_forward_refs()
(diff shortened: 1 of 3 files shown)


wssheldon:
Does that resolve the issue? If so, would `Incidents` have a similar issue?

```python
@router.get("", summary="Retrieve a list of incidents.")
def get_incidents(
    common: CommonParameters,
    include: List[str] = Query([], alias="include[]"),
    expand: bool = Query(default=False),
):
    """Retrieves a list of incidents."""
    pagination = search_filter_sort_paginate(model="Incident", **common)

    if expand:
        return json.loads(IncidentExpandedPagination(**pagination).json())

    if include:
        # only allow two levels for now
        include_sets = create_pydantic_include(include)

        include_fields = {
            "items": {"__all__": include_sets},
            "itemsPerPage": ...,
            "page": ...,
            "total": ...,
        }
        return json.loads(IncidentPagination(**pagination).json(include=include_fields))
    return json.loads(IncidentPagination(**pagination).json())
```

wssheldon (approved):
Regardless, I do think we want this change.

aaronherman:
Good catch - I validated that incidents also don't support all fields and have updated it to use IncidentExpandedPagination

PR #5945, opened 2025-04-30
debt: upgrade sqlalchemy to 1.4.36

ghost:
The issue in the comment has been resolved and being on `1.3` significantly limits what dependencies we can use related to SQLAlchemy.


mvilanova (approved):

src/dispatch/database/core.py
```diff
@@ -247,7 +264,13 @@ def refetch_db_session(organization_slug: str) -> Session:
 @contextmanager
 def get_session() -> Session:
     """Context manager to ensure the session is closed after use."""
-    session = SessionLocal()
+    schema_engine = engine.execution_options(
```
mvilanova, line 267: We should encapsulate this logic in a function that takes an org slug as a param, so we don't need to duplicate the code in `get_organization_session()`

src/dispatch/database/service.py
```diff
@@ -370,7 +396,7 @@ def apply_filter_specific_joins(model: Base, filter_spec: dict, query: orm.query
         (Incident, "IndividualContact"): (Incident.participants, True),
         (Incident, "Term"): (Incident.terms, True),
         (Signal, "Tag"): (Signal.tags, True),
-        (Signal, "TagType"): {Signal.tags, True},
+        (Signal, "TagType"): (Signal.tags, True),
```
mvilanova, line 399: nice catch

PR #5964, opened 2025-05-07
feat(db): adds user confirmation to database cli commands

whitdog47:
Adds an additional layer of checks for database `init` and `restore` CLI commands to verify database configuration.


mvilanova (approved):

mvilanova:
May want to investigate why playwright is complaining before merging...

PR #5975, opened 2025-05-16
fix(signal): set to None to allow missing parameters

whitdog47:
This PR adds default None values to several optional fields in signal and project models to allow missing parameters without validation errors.

### DIFF
--- a/src/dispatch/project/models.py
+++ b/src/dispatch/project/models.py
@@ -132,7 +132,7 @@ class ProjectUpdate(ProjectBase):
 
 
 class ProjectRead(ProjectBase):
-    id: PrimaryKey | None
+    id: PrimaryKey | None = None
     stable_priority: IncidentPriorityRead | None = None
 
     model_config = ConfigDict(from_attributes=True)
--- a/src/dispatch/signal/models.py
+++ b/src/dispatch/signal/models.py
@@ -387,21 +387,21 @@ class SignalStats(DispatchBase):
 
 class SignalInstanceBase(DispatchBase):
     project: ProjectRead | None
-    case: CaseReadMinimal | None
+    case: CaseReadMinimal | None = None
     canary: bool | None = False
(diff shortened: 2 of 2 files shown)


src/dispatch/project/models.py
```diff
@@ -132,7 +132,7 @@ class ProjectUpdate(ProjectBase):
 
 
 class ProjectRead(ProjectBase):
-    id: PrimaryKey | None
+    id: PrimaryKey | None = None
```
mvilanova, line 135: `id` should not be optional when returning a project object

src/dispatch/signal/models.py
```diff
@@ -387,21 +387,21 @@ class SignalStats(DispatchBase):
 
 class SignalInstanceBase(DispatchBase):
     project: ProjectRead | None
-    case: CaseReadMinimal | None
+    case: CaseReadMinimal | None = None
     canary: bool | None = False
     entities: list[EntityRead] | None = []
     raw: dict[str, Any]
-    external_id: str | None
+    external_id: str | None = None
```
mvilanova, line 394: Shouldn't be optional

src/dispatch/signal/models.py
```diff
@@ -387,21 +387,21 @@ class SignalStats(DispatchBase):
...
     created_at: datetime | None = None
 
 
 class SignalInstanceCreate(SignalInstanceBase):
-    signal: SignalRead | None
-    case_priority: CasePriorityRead | None
-    case_type: CaseTypeRead | None
-    conversation_target: str | None
-    oncall_service: ServiceRead | None
+    signal: SignalRead | None = None
```
mvilanova, line 400: I think all of these shouldn't be optional either. We can make the mandatory and put out fixes where the code complains.

mvilanova (approved):
Let's take a harder look at the models in a future PR, as they're not in an ideal state.

PR #5982, opened 2025-05-17
fix(case): more explicit None defaults

whitdog47:


wssheldon (approved):

wssheldon:
I'm assuming anything that was `Optional` without a default before (or with it I suppose) that was updated to `* | None` must now set a default of `None`?

Does this get all of them?

whitdog47:
@wssheldon turns out there were 31 more files that needed it!

PR #6000, opened 2025-05-20
fix: slack plugin activity fetch

ghost:


src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -447,16 +436,10 @@ def fetch_events(
...
-                client=client,
-                subject=subject,
-                oldest=oldest,
-            )
-        except Exception:
-            logger.exception("Error fetching events from Slack")
+            activity = event.fetch_activity(client, subject, oldest)
+            return activity
+        except Exception as e:
+            logger.exception("An error occurred while fetching incident events from the Slack plugin.", exc_info=e)
```
mvilanova, line 442:
```suggestion
            logger.exception("An error occurred while fetching incident or case events from the Slack plugin.", exc_info=e)
```

src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -447,16 +436,10 @@ def fetch_events(
...
-                # It's a class, instantiate it
-                event = event(plugin_event)
-            return event.fetch_activity(
-                client=client,
-                subject=subject,
-                oldest=oldest,
-            )
-        except Exception:
-            logger.exception("Error fetching events from Slack")
+            activity = event.fetch_activity(client, subject, oldest)
```
mvilanova, line 439: If `event` is a class we will get the `ThreadActivityEvent.fetch_activity() missing 1 required positional`

src/dispatch/plugins/dispatch_slack/plugin.py
```diff
@@ -447,16 +436,10 @@ def fetch_events(
             event = self.get_event(plugin_event)
             if event is None:
                 raise ValueError(f"No event found for Slack plugin event: {plugin_event}")
-            if isinstance(event, type):
-                # It's a class, instantiate it
-                event = event(plugin_event)
```
mvilanova, line 452:
```
if isinstance(event, type):
    try:
        event = event(plugin_event)
    except TypeError:
        event = event()
```

src/dispatch/plugins/dispatch_slack/plugin.py
mvilanova, line 439:
```suggestion
            activity = event().fetch_activity(client, subject, oldest)
```

mvilanova (approved):

PR #6152, opened 2025-07-24
feat(ui): Adds icons for case and incident in table status column

mvilanova:
This PR adds an incident icon to the status column in the case table and a case icon to status column in the incident able when a case is escalated to an incident.

Screenshot 2025-07-24 at 12 35 15 PM
Screenshot 2025-07-24 at 12 35 35 PM

### DIFF
--- a/src/dispatch/static/dispatch/src/case/CaseStatus.vue
+++ b/src/dispatch/static/dispatch/src/case/CaseStatus.vue
@@ -26,7 +26,6 @@
             variant="text"
             density="comfortable"
             class="ml-1"
-            size="small"
             color="orange"
             @click.stop="navigateToIncident(incidents[0])"
           />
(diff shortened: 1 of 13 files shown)


src/dispatch/static/dispatch/src/case/CaseStatus.vue
```diff
@@ -17,6 +17,22 @@
...
+    <template v-if="incidents && incidents.length > 0">
+      <v-tooltip location="bottom" :text="`View incident`">
+        <template #activator="{ props }">
+          <v-btn
+            v-bind="props"
+            icon="mdi-fire"
+            variant="text"
+            density="comfortable"
+            class="ml-1"
+            size="small"
```
aaronherman, line 29: nit: the other buttons aren't small
```suggestion
```

aaronherman (approved):

PR #6169, opened 2025-08-05
fix(search_filter): adds refresh to search filter when not showing

whitdog47:
Adds a refresh button to the search filter expression edit dialog to allow users to re-fetch the search filter expression when edits need to be reverted or when the expression doesn't load initially.

- Adds a new `get` action to the search store for fetching individual search filters
- Implements a refresh button in the EditDialog component that calls the new get action
- Creates a new API endpoint to retrieve a specific search filter by ID

image


nathanmyee (approved):
:shipit:

PR #6173, opened 2025-08-06
Emojify and simplify change notifications for incidents and cases

nathanmyee:
This PR adds emojis and simplifies the text formatting to make change notifications easier to read. Updated notifications for other event types, like summaries or creation, will follow.

Sample Slack post for incident change notification:
image

Sample Slack post for case change notification:
image

## Full emoji changes

  Notification Headers:
  - :rotating_light: Added to incident notification titles
  - :briefcase: Added to case notification titles

  Role Indicators:
  - :firefighter: Added to incident commanders
  - :detective: Added to case assignees

  Field Labels:
  - :memo: Added to title fields
  - :label: Added to type change notifications

  🔄 Dynamic Status & Change Notifications

  Status Changes (with contextual emojis):
  - :fire: Active incidents
  - :shield: Stable incidents
  - :white_check_mark: Closed incidents/cases
  - :new: New cases
  - :mag: Triage cases
  - :arrow_up: Escalated cases

  Priority/Severity Changes (with directional indicators):
  - :arrow_up: Increased priority/severity
  - :arrow_down: Decreased priority/severity
  - :left_right_arrow: Lateral changes

  Visibility Changes:
  - :unlock: Open visibility
  - :lock: Restricted visibility
  - :eye: Other visibility states


src/dispatch/plugins/dispatch_slack/messaging.py
```diff
@@ -190,12 +190,18 @@ def build_unexpected_error_message(guid: str) -> str:
 def format_default_text(item: dict):
     """Creates the correct Slack text string based on the item context."""
     if item.get("title_link"):
-        return f"*<{item['title_link']}|{item['title']}>*\n{item['text']}"
+        text_part = f"\n{item['text']}" if item.get('text') else ""
+        return f"*<{item['title_link']}|{item['title']}>*{text_part}"
```
aaronherman, line 194: Would this work?
```suggestion
        return f"*<{item['title_link']}|{item['title']}>*{item.get('text', "")}"
```

nathanmyee: I took a closer look at this suggestion and it's a bit different than my original intention. a newline is appended regardless if `text` is empty or not. by separating it out, the newline is only appended if `text` is valid.

PR #6180, opened 2025-08-12
fix(slack): resolved by block error

whitdog47:
This PR fixes an error in the Slack block since `resolved_by` is an `IndividualContact`, we should remove the erroneous resolution to `individual`.


nathanmyee (approved):
🚀

